PR #7, opened 2016-12-06
Fix lower bound on http-client

bos:


Gabriel439:
Thank you for catching that! :)

PR #14, opened 2016-12-17
Boost http-client upperbound and fix compilation fallout

fosskers:
### TODO
- [x] Boost stack resolver to Dec 1's nightly
- [x] Boost `http-client` and `http-client-tls` upper bounds
- [x] Fix `HttpException` usage

### Motivation
A project I'm working on uses `servant-0.9`, which is only available in Stackage nightlies so far. These same nightlies use the latest major version of `http-client`, which Dhall does not, meaning that even adding `dhall-1.0.1` to the `extra-deps` field of my `stack.yml` won't let deps resolve properly. I'd like to try Dhall for my applications configuration, hence this PR.

**Note:** The `.cabal` here is lying a bit, since I didn't touch the lower bounds for `http-client` and `http-client-tls`. Applications written with less than the most recent major version of both will not compile. However, you said on Twitter that you'd add the correct CPP magic afterwards to handle this.

**Note:** You use `microlens`, which I'm a fan of, but your `trifecta` dep pulls `lens`:

dhall


Gabriel439:
It will take me some time to get to this because I'm busy preparing for a talk this weekend but I agree that there's no point using `microlens` now that this uses `trifecta`

PR #37, opened 2017-04-16
Increase connection timeout for imports

markus1189:
### TL;DR

Set the response timeout from 1s to 30s, because 1s is much too optimistic.

### Why

From time to time I have the _problem_ luck of being with a rather modest internet connection (~376Kb/s) and I noticed that even running something simple like:

```
let not = https://ipfs.io/ipfs/QmcTbCdS21pCxXysTzEiucDuwwLWbLUWNSKwkJVfwpy2zK/Prelude/Bool/not
in  not True
```

via `dhall` fails with `ConnectionTimeout`.  I saw that the connection response timeout is set to `1 second`, which is much too short for a bad internet connection. 

### How
To address this, I raised the time limit from `1 second` to `30 seconds`, because that is also the time that is used in the default `managerResponseTimeout` (https://hackage.haskell.org/package/http-client-0.5.6.1/docs/src/Network-HTTP-Client-Types.html#ManagerSettings).

With that setting I can finally use `dhall` even when I don't have a fast internet connection available. :D

### DIFF
--- a/src/Dhall/Import.hs
+++ b/src/Dhall/Import.hs
@@ -325,9 +325,9 @@ needManager = do
         Nothing -> do
             let settings = HTTP.tlsManagerSettings
 #if MIN_VERSION_http_client(0,5,0)
-                    { HTTP.managerResponseTimeout = HTTP.responseTimeoutMicro 1000000 }  -- 1 second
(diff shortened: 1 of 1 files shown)


Gabriel439:
Yeah, great idea!  Sorry about picking such a strict default initially

PR #39, opened 2017-04-17
Add new argument: --version

markus1189:
Just noticed that there is no easy way to see which version of `dhall` you are using.  To remedy this, I added a new flag to the `dhall` executable, `--version` which prints it.


Gabriel439:
Thank you for contributing this! Very helpful :)

PR #40, opened 2017-04-19
New builtin function: Text/length

markus1189:
I think it would be handy to have a function that calculates the `length` of a `Text` literal.

This pr adds a new builtin function `Text/length` that does just that and is also normalized:

```
$ dhall <<< '\(x: Bool) -> Text/length "foo" + Text/length "bar" + Text/length ""'
∀(x : Bool) → Natural

λ(x : Bool) → +6
```

Take this with a grain of salt, it was more a "I would like to hack on dhall" inspired feature and not particularly driven by practical need ;)  I'm okay if you want to keep the language minimal and don't merge this.

### DIFF
--- a/src/Dhall/Parser.hs
+++ b/src/Dhall/Parser.hs
@@ -137,6 +137,7 @@ identifierStyle = IdentifierStyle
         , "Integer"
         , "Double"
         , "Text"
+        , "Text/length"
         , "List"
         , "List/build"
         , "List/fold"
@@ -446,6 +447,7 @@ exprF embedded = choice
             ,   noted      exprF05
             ,   noted      exprF06
             ,   noted      exprF07
+            ,   noted      exprF35
             ,   noted      exprF12
             ,   noted      exprF13
             ,   noted      exprF14
@@ -605,6 +607,10 @@ exprF embedded = choice
         a <- exprA embedded
         symbol ")"
         return a
+
+    exprF35 = do
+        reserve "Text/length"
+        return TextLength
 
 const :: Parser Const
 const = const0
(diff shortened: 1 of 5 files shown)


src/Dhall/Parser.hs
```diff
@@ -446,6 +447,7 @@ exprF embedded = choice
             ,   noted      exprF05
             ,   noted      exprF06
             ,   noted      exprF07
+            ,   noted      exprF35
```
markus1189: this position was chosen rather arbitrarily, can you give some guidance what makes sense here?

markus1189:
Oh and related to this: I think a test suite would be great ;)

Gabriel439:
This is one function I've been debating whether to add for a while

The main reason I haven't added it so far is to avoid users testing for the empty string by using `Natural/isZero (Text/length someText)`.  Actually, more generally I try to avoid users inspecting strings at all since usually it's a design smell for a configuration file to do any sort of parsing or care what the contents of a string are.

On the other hand, it has nice algebraic properties like:

```haskell
Text/length (x ++ y) = Text/length x + Text/length y

Text/length "" = +0
```

But I'm still slightly reluctant to add it for the former set of reasons

Gabriel439:
Also, I agree that Dhall is overdue for a test suite.  I will try to add a test suite over this weekend

markus1189:
I'm okay with either result, whether it's merged or not ;).  Are there other language features you wanted to test that I might take a stab at?

Gabriel439:
You could implement an `Optional/build` with a fusion rule to ensure that it cancels out with `Optional/fold`

markus1189:
I'll close this for now, if you want to add it anyway at some point feel free to do so.

PR #41, opened 2017-04-28
Optional/build with fusion

markus1189:
As suggested in #40, I started implementing `Optional/build` with fusion.  I am not yet done completely, because there is one normalization that I don't quite get.  I'll add a note inline, once that is done I can update the pr and this is hopefully mergeable ;)

### DIFF
--- a/src/Dhall/Core.hs
+++ b/src/Dhall/Core.hs
@@ -267,13 +267,15 @@ data Expr s a
     | OptionalLit (Expr s a) (Vector (Expr s a))
     -- | > OptionalFold                             ~  Optional/fold
     | OptionalFold
+    -- | > OptionalBuild                            ~  Optional/build
+    | OptionalBuild
     -- | > Record            [(k1, t1), (k2, t2)]   ~  { k1 : t1, k2 : t1 }
     | Record    (Map Text (Expr s a))
     -- | > RecordLit         [(k1, v1), (k2, v2)]   ~  { k1 = v1, k2 = v2 }
     | RecordLit (Map Text (Expr s a))
     -- | > Union             [(k1, t1), (k2, t2)]   ~  < k1 : t1 | k2 : t2 >
     | Union     (Map Text (Expr s a))
-    -- | > UnionLit (k1, v1) [(k2, t2), (k3, t3)]   ~  < k1 = t1 | k2 : t2 | k3 : t3 > 
+    -- | > UnionLit (k1, v1) [(k2, t2), (k3, t3)]   ~  < k1 = t1 | k2 : t2 | k3 : t3 >
     | UnionLit Text (Expr s a) (Map Text (Expr s a))
     -- | > Combine x y                              ~  x ∧ y
     | Combine (Expr s a) (Expr s a)
@@ -339,6 +341,7 @@ instance Monad (Expr s) where
     Optional         >>= _ = Optional
     OptionalLit a b  >>= k = OptionalLit (a >>= k) (fmap (>>= k) b)
     OptionalFold     >>= _ = OptionalFold
+    OptionalBuild    >>= _ = OptionalBuild
     Record    a      >>= k = Record (fmap (>>= k) a)
     RecordLit a      >>= k = RecordLit (fmap (>>= k) a)
     Union     a      >>= k = Union (fmap (>>= k) a)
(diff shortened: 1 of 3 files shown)


src/Dhall/Core.hs
```diff
@@ -1130,6 +1142,7 @@ normalize e = case e of
             App NaturalIsZero (NaturalLit n) -> BoolLit (n == 0)
             App NaturalEven (NaturalLit n) -> BoolLit (even n)
             App NaturalOdd (NaturalLit n) -> BoolLit (odd n)
+            -- What does this do? (TODO)
```
markus1189: I am not quite sure what the rule for `ListBuild` does here in the normalization case, could you maybe explain a little bit?  For example, what is `Var "Cons"` referring to?

src/Dhall/Core.hs
```diff
@@ -1396,6 +1415,7 @@ isNormalized e = case shift 0 "_" e of  -- `shift` is a hack to delete `Note`
         App NaturalIsZero (NaturalLit _) -> False
         App NaturalEven (NaturalLit _) -> False
         App NaturalOdd (NaturalLit _) -> False
+        -- see question above (TODO)
```
markus1189: Belongs to the comment above, because we also have to add it here

src/Dhall/Core.hs
```diff
@@ -1105,6 +1113,10 @@ normalize e = case e of
             App NaturalBuild (App NaturalFold e') -> normalize e'
             App NaturalFold (App NaturalBuild e') -> normalize e'
 
+            -- fold/build fusion for `Optional`
+            App OptionalBuild (App OptionalFold e') -> normalize e'
```
Gabriel439: It should probably be like the fusion rule for lists:

```haskell
App (App OptionalBuild _) (App (App OptionalFold _) e') = e'
```

markus1189: Oh my! You are right of course. Fixed

src/Dhall/Core.hs
Gabriel439: So what is going on here is that the user supplies a function to build the list which might look like this:

```haskell
List/build
Integer
(   λ(list : Type)
→   λ(cons : Integer → list → list)
→   λ(nil  : list)
→   cons 1 (cons 2 (cons 3 nil))
)
```

... and the abstract syntax tree for that would be:

```haskell
App
(App ListBuild Integer)
(Lam "list" Type
    (Lam "cons" (Pi "_" Integer (Pi "_" "list" "list"))
        (Lam "nil" "list"
            (App (App "cons" 1) (App (App "cons" 2) (App (App "cons" 3) "nil"))) ) ) )
```

... and we basically need to turn that syntax tree into a vector.

One simple approach to do this would be to pattern match on the syntax tree, extract the labels that the user chose for `cons` and `nil` and use those labels to detect if the built list is correct.  The code for that might look something like this:

```haskell
k' cons nil = go0 k

go0 (App _ (Lam l _ (Lam c _ (Lam n _ e)))) = go1 e
  where
    go (App (App (Var (V c' 0)) x) e')
        | c == c' = cons x (go e')
    go (Var (V n' 0))
        | n == n' = nil
    go _ = error "Urk!"
```

... however that's not exactly right because it doesn't deal with variable shadowing.  For example, it gives the wrong answer if the user supplies the same variable name for the `cons` and `nil` arguments, like this:

```haskell
List/build
Integer
(   λ(list : Type)
→   λ(x : Integer → list → list)
→   λ(x  : list)
→   x@1 1 (x@1 2 (x@1 3 x))
)
```

The way I worked around this was to use `dhall`'s normalize functionality to replace the user's chosen label names with labels of my choice.  In other words, I take the user's function and apply it to two unique labels like this:

```haskell
(   λ(list : Type)
→   λ(x : Integer → list → list)
→   λ(x  : list)
→   x@1 1 (x@1 2 (x@1 3 x))
)
(List t)
Cons
Nil
```

... which reduces to:

```haskell
Cons 1 (Cons 2 (Cons 3 Nil))
```

That ensures that the variable names are unique so that I don't have to worry about shadowing.

This is not the most efficient implementation, but it was the easiest to implement at the time.

markus1189: Ahh that's clever ;) Thanks!

markus1189:
Oops I didn't mean to close this, I only temporarily deleted the wrong remote branch :D

PR #42, opened 2017-04-29
Add test suite

markus1189:
Okay this Optional/build stuff is taking me a little longer that I wanted :)  In the meantime, I started to add tests for my normalization rules, so I thought that it would be good to get at least the test suite into master ASAP.

I use tasty and tasty-hunit, but I don't know if this is okay for you, so any feedback appreciated!


Gabriel439:
Awesome!  Thanks for setting this up :)

PR #43, opened 2017-04-30
Escaped field names.  Fixes #31

### LINKED ISSUE #31: Use "in" as a field name (despite its keyword status)
Is there any way to produce a record with a field name "in"? (I fear not, since it's a keyword for `let` expressions.) I was playing with the idea of using Dhall to produce a YAML file  (with `dhall-to-yaml`) for use with Swagger (http://swagger.io), which as part of defining parameters that an API endpoint can accept, uses an object key "in" to specify where  a parameter can occur.

For example,

```
{ swagger = "2.0"
, parameters = {
    foo = {
      name = "foo"
    , in = "path"
    , type = "string"
    }
  }
  ...
}
```

to produce

```
    swagger: "2.0"
    parameters:
        foo:
            name: "foo"
            in: "path"
            type: "string" 
    ...
```

Gabriel439:
You can now escape an field or variable name using backticks, like this:

```haskell
    let `let` = 2
in  let `in`  = True
in  { `True` = `let`
    , `Type` = λ(`Kind` : Bool) → `Kind` && `in`
    }
```

The main purpose of this is to support arbitrary record field names when
marshalling Dhall into other file formats (such as JSON).  For example,
this Dhall configuration file:

```haskell
{ swagger = "2.0"
, parameters = {
    foo = {
      name = "foo"
    , `in` = "path"
    , type = "string"
    }
  }
  ...
}
```

... can be used to generate this JSON file:

```json
    swagger: "2.0"
    parameters:
        foo:
            name: "foo"
            in: "path"
            type: "string"
```

### DIFF
--- a/src/Dhall/Core.hs
+++ b/src/Dhall/Core.hs
@@ -36,6 +36,7 @@ module Dhall.Core (
 
     -- * Miscellaneous
     , internalError
+    , reservedIdentifiers
     ) where
 
 #if MIN_VERSION_base(4,8,0)
@@ -44,6 +45,7 @@ import Control.Applicative (Applicative(..), (<$>))
 #endif
 import Data.Bifunctor (Bifunctor(..))
 import Data.Foldable
+import Data.HashSet (HashSet)
 import Data.Map (Map)
 import Data.Monoid ((<>))
 import Data.String (IsString(..))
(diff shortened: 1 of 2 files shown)


markus1189:
Does this forbid spaces in keys?  E.g.:

```
{ `foo bar` = "test" }
```
? Or does `aeson` escape those automatically?

markus1189:
Ah nevermind, that should not be a problem.  :+1:

Gabriel439:
Yeah, this currently only permits the same set of characters as unescaped names.  The only thing this does is avoid conflicts with reserved identifiers

PR #44, opened 2017-05-01
Implement Optional/build with fusion

markus1189:
Finally got around to do this ;)

- Defines a new primitive `OptionalFold`
- Implements normalization:
  - inlining of Optional/build if constants are used as args
  - fusion of fold/build and build/fold

### DIFF
--- a/src/Dhall/Core.hs
+++ b/src/Dhall/Core.hs
@@ -42,6 +42,7 @@ module Dhall.Core (
 #else
 import Control.Applicative (Applicative(..), (<$>))
 #endif
+import Control.Applicative (empty)
 import Data.Bifunctor (Bifunctor(..))
 import Data.Foldable
 import Data.Map (Map)
@@ -182,7 +183,7 @@ data Expr s a
     = Const Const
     -- | > Var (V x 0)                              ~  x
     --   > Var (V x n)                              ~  x@n
-    | Var Var             
+    | Var Var
     -- | > Lam x     A b                            ~  λ(x : A) -> b
     | Lam Text (Expr s a) (Expr s a)
     -- | > Pi "_" A B                               ~        A  -> B
@@ -253,7 +254,7 @@ data Expr s a
     -- | > ListLength                               ~  List/length
     | ListLength
     -- | > ListHead                                 ~  List/head
-    | ListHead 
+    | ListHead
     -- | > ListLast                                 ~  List/last
     | ListLast
     -- | > ListIndexed                              ~  List/indexed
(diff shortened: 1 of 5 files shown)


src/Dhall/Core.hs
```diff
@@ -1130,6 +1141,22 @@ normalize e = case e of
             App NaturalIsZero (NaturalLit n) -> BoolLit (n == 0)
             App NaturalEven (NaturalLit n) -> BoolLit (even n)
             App NaturalOdd (NaturalLit n) -> BoolLit (odd n)
+            App (App OptionalBuild t) k
+                | check     -> OptionalLit t (Data.Vector.fromList k')
```
Gabriel439: Minor suggestion: instead of using `Data.Vector.fromList` here, you can have `go` build a `Vector` directly:

```haskell
go (App (Var "Just") e') = pure e'
go (Var "Nothing")       = empty
```

src/Dhall/Core.hs
```diff
@@ -1396,6 +1429,15 @@ isNormalized e = case shift 0 "_" e of  -- `shift` is a hack to delete `Note`
         App NaturalIsZero (NaturalLit _) -> False
         App NaturalEven (NaturalLit _) -> False
         App NaturalOdd (NaturalLit _) -> False
+        App (App OptionalBuild t) k0 -> isNormalized t && isNormalized k0 && not (check0 k0)
+          where
+            check0 (Lam _ _ (Lam just _ (Lam nothing _ k))) = check1 just nothing k
+            check0 _ = False
+
+            check1 just nothing (App (Var (V just' n)) _) =
+                just == just' && n == (if just == nothing then 1 else 0)
```
Gabriel439: Related to your previous question: this is very similar to how the more efficient version of normalization would work.  It's just that I wrote `isNormalized` long after I wrote `normalize` so I wrote `isNormalized` with more careful attention to performance

src/Dhall/TypeCheck.hs
```diff
@@ -444,6 +444,13 @@ typeWith _      OptionalFold      = do
                 (Pi "optional" (Const Type)
                     (Pi "just" (Pi "_" "a" "optional")
                         (Pi "nothing" "optional" "optional") ) ) ) )
+typeWith _      OptionalBuild     = do
+    return
+        (Pi "a" (Const Type)
+            (Pi "f" f (App Optional "a") ) )
```
Gabriel439: Minor suggestion: replace `"f"` with `"_"` for consistency with `List/build` so that it doesn't show up in the type

tests/Normalization.hs
```diff
@@ -1,18 +1,73 @@
...
+  normalize' e @?= NaturalLit 3
+  where e = NaturalPlus (NaturalLit 1) (NaturalLit 2)
 
-normalize' :: Expr () () -> Expr () ()
+optionalFold :: TestTree
+optionalFold = testGroup "Optional/fold" [ just, nothing ]
+  where test label inp out = testCase label $ do
+             isNormalized (e inp) @?= False
+             normalize' (e inp) @?= out
+        e inp = (OptionalFold `App` Text `App` OptionalLit Text inp `App`
```
Gabriel439: Clever use of infix `App`!  That would make the `normalization` code a lot easier to read with fewer parentheses since you I believe can also pattern match on an infix constructor

tests/Normalization.hs
```diff
@@ -1,18 +1,73 @@
...
+                                             Natural `App` (Lam "j" Text (NaturalLit 1)) `App`
+                                                 (NaturalLit 2))
+        just = test "just" [TextLit "foo"] (NaturalLit 1)
+        nothing = test "nothing" [] (NaturalLit 2)
+
+optionalBuild :: TestTree
+optionalBuild = testCase "Optional/build" $ do
```
Gabriel439: There are two test cases worth adding to deal with common ways to implement build incorrectly.  First, you want to test that shadowing is handled correctly:

```haskell
Optional/build
Integer
(λ(optional : Type) → λ(x : Integer → optional) → λ(x : optional) → x@1 3)
```

... and you also want to test that it correctly handles an irreducible `build` like this:

```haskell
  λ(id : ∀(a : Type) → a → a)
→ Optional/build
  Integer
  ( λ(optional : type)
  → λ(just : Integer → optional)
  → λ(nothing : optional)
  → id optional (just 3)
  )
```

... where "correctly handles" means that it doesn't fail with an `error` and doesn't accidentally interpret it as a reducible `build`

markus1189:
Thanks for the review.  I addressed all of your points (tests all pass, yay!)

Gabriel439:
Awesome!  Thank you for contributing this :)

Issue #49, opened 2017-05-07
Add new conversion primitives


Gabriel439:
As mentioned in #47, we need the following new builtin functions in order to render numeric values:

* [x] `Natural/show : Natural → Text`
* [x] `Integer/show : Integer → Text`
* [x] `Double/show : Double → Text`
* [x] `Natural/toInteger : Natural → Integer`

The `show` functions should render in the exact same format that Dhall parses to source code.  The purpose of the `Natural/toInteger` function is to help render `Natural` numbers without a leading `+` by converting them to `Integer`s first

Gabriel439:
All four conversion primitives are done now, so I'll mark this resolved

PR #50, opened 2017-05-08
Implement Natural/toInteger

markus1189:
As in #49 this implements the `Natural/toInteger: Natural -> Integer` conversion.

I also slightly refactored the tests to use custom assertions.

Something that is still missing: A Tutorial section on this function.  I guess it makes sense to have a whole section about `Conversions`, that lists all primitive/library conversions as `Bool -> Text`, `Natural -> Integer`, `Integer -> Text` etc.  What are your thoughts?


Gabriel439:
I think the tutorial section can wait until we have string interpolation complete

PR #51, opened 2017-05-08
Fix `build . fold` fusion tests

Gabriel439:
The previous tests for `build . fold` fusion for `Optional` values would
succeed even without fusion.  This change updates the tests to only
succeed if fusion is occurring.

### DIFF
--- a/tests/Normalization.hs
+++ b/tests/Normalization.hs
@@ -92,43 +92,39 @@ fusion = testGroup "Optional build/fold fusion" [ fuseOptionalBF
 
 fuseOptionalBF :: TestTree
 fuseOptionalBF = testCase "fold . build" $ do
-  j <- test "just \"foo\""
-  j `assertNormalizesTo` "+42"
-  n <- test "nothing"
-  n `assertNormalizesTo` "+2"
-  where
-    test e = code [NeatInterpolation.text|
-Optional/fold
-Text
-(   Optional/build
-    Text
-    (   λ(optional : Type)
-    →   λ(just : Text → optional)
-    →   λ(nothing : optional)
-    →   $e
+  e0 <- code [NeatInterpolation.text|
+    λ(  f
+    :   ∀(optional : Type)
+    →   ∀(just : Text → optional)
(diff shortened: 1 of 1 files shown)


markus1189:
Why would it also succeed without fusion? 😕

Gabriel439:
Using this as an example:

```haskell
Optional/build
Natural
(   λ(optional : Type)
→   λ(just : Natural → optional)
→   λ(nothing : optional)
→   just
    (   Optional/fold
        Text
        (["foo"] : Optional Text)
        Natural
        (λ(just : Text) → +42)
        +2
    )
)
```

It will first try to normalize all the arguments to `build` before normalizing `build` itself.  That means it will normalize the lambda and since Dhall can normalize under lambda that will normalize to:

```haskell
Optional/build
Natural
(   λ(optional : Type)
→   λ(just : Natural → optional)
→   λ(nothing : optional)
→   just +42
)
```

Then the normalization rule for `Optional/build` will kick in and build the `Optional` value, which normalizes to:

```haskell
[+42] : Optional Integer
```

Similar concept for the other example:

```haskell
Optional/fold
Text
(   Optional/build
    Text
    (   λ(optional : Type)
    →   λ(just : Text → optional)
    →   λ(nothing : optional)
    →   just "foo"
    )
)
Natural
(λ(j : Text) → +42)
+2
```

Dhall will normalize the lambda before normalizing the `fold`, and the lambda can be normalized by itself, giving:

```haskell
Optional/fold
Text
(["foo"] : Optional Text)
Natural
(λ(j : Text) → +42)
+2
```

... and then `Optional/fold` is normalized in the usual way:

```haskell
+42
```

You can verify this yourself by just disabling the fusion rules for `Optional/fold` and `Optional/build` and checking that the tests still pass before this change

The reason the updated tests work is that they block normalization of the lambda by replacing it with a bound variable that can't be normalized further, so the only way to simplify the code is via the fusion rule

markus1189:
Okay that makes sense, thanks for the explanation!

Gabriel439:
You're welcome!

PR #52, opened 2017-05-08
Implement Natural/show

markus1189:
As the title says + I also renamed the `exprF01,exprF02,...` to something (hopefully) more meaningful, because it became a little unwieldy ;)

Also relates to #49


Gabriel439:
Looks great!  Also, good call on properly renaming those parsers :)

PR #54, opened 2017-05-09
Implement Double/show and Integer/show

markus1189:
As in #49, this implements the two remaining conversions.


Gabriel439:
Excellent work!  Thanks for doing this :)

PR #56, opened 2017-05-13
Add tests for Prelude examples

Gabriel439:
This adds tests that match the examples in the Prelude documentation and
also fixes a couple of mistakes caught along the way

### DIFF
--- a/tests/Examples.hs
+++ b/tests/Examples.hs
@@ -220,154 +220,154 @@ _Bool_and_0 = Test.Tasty.HUnit.testCase "Example #0" (do
     e <- Util.code [NeatInterpolation.text|
 ./Prelude/Bool/and ([True, False, True] : List Bool)
 |]
-    Test.Tasty.HUnit.assertEqual "" (Util.normalize' e) "False" )
+    Util.assertNormalizesTo e "False" )
 
 _Bool_and_1 :: TestTree
 _Bool_and_1 = Test.Tasty.HUnit.testCase "Example #1" (do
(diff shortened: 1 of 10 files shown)


tests/Examples.hs
```diff
@@ -0,0 +1,1053 @@
...
+                ]
+            ]
+        ]
+
+_Bool_and_0 :: TestTree
+_Bool_and_0 = Test.Tasty.HUnit.testCase "Example #0" (do
+    e <- Util.code [NeatInterpolation.text|
+./Prelude/Bool/and ([True, False, True] : List Bool)
+|]
+    Test.Tasty.HUnit.assertEqual "" (Util.normalize' e) "False" )
```
markus1189: What do you think about using `Util.assertNormalizesTo e "False"` instead?

tests/Examples.hs
```diff
@@ -0,0 +1,1053 @@
...
+|]
+    return () )
+
+_Monoid_01 :: TestTree
+_Monoid_01 = Test.Tasty.HUnit.testCase "Example #1" (do
+    _ <- Util.code [NeatInterpolation.text|
+./Prelude/Bool/or
+    : ./Prelude/Monoid Bool
+|]
+    return () )
```
markus1189: We could avoid the `do` completely, using `(void $ Util.code [NeatInterpolation.text| ... |])` or we could have a new assertion, e.g. `assertTypeChecks [|NeatInterpolation.text| ... |]`.  That would make it more obvious what's the intention

PR #57, opened 2017-05-14
Update Prelude

Gabriel439:
This adds new builtins to the Prelude and also adds corresponding tests

### DIFF
--- a/Prelude/Bool/show
+++ b/Prelude/Bool/show
@@ -1,0 +1,16 @@
+{-
+Render a `Bool` as `Text` using the same representation as Dhall source code
+(i.e. beginning with a capital letter)
+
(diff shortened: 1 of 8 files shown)


markus1189:
What is the advantage of having this as both a builtin and a file in `Prelude`?

Gabriel439:
@markus1189: The main reasons are:

* Discoverability - the user can discover the whole standard library (both built-in and derived) by browsing the Prelude
* Documentation - the user gets documentation/examples/type-signatures for builtins, just like the rest of the Prelude
* Consistency - the user doesn't have to remember which utilities are builtins and which ones are derived

markus1189:
Okay I see ;)

PR #60, opened 2017-05-22
Add support for string interpolation.  Fixes #47

### LINKED ISSUE #47: Multiline strings and string interpolation
I am currently writing a blog system with Haskell and I store the posts as Haskell values:
```haskell
module Entries.E170426 where

import Imports
import Types

entry = Entry
  { entryTitle = "Hello World!"
  , entryCreated = fromGregorian 2017 04 26
  , entryUpdated = fromGregorian 2017 04 26
  , entryKeywords = [Blogging]
  , entryCategory = []
  , entryLanguage = English
  , entryComments = Github
  , entryAbstract = [md|

This is my first blog post.

|]
  , entryContent = [md|
... 
|]}
```

The benefit of this is, that I can add extra information like keywords and rely on the type checker to check that they are valid (for example using the keyword "Bloging" would be a compile-time error).
The reason I don't use pandoc filters for this is that I want use arbitrary Haskell data types in the quasiquote via an escaping mechanism allowing me to implement microformats and schema.org markup with ease. The problem is that I will have to use haskell-src-exts as an additional dependency for that.

So I am currently looking at dhall as an alternative for that use case. I would write a blog entry in a dhall file as an Entry data type and import the files into Haskell. However, I need multi-line strings (like in the md quasi-quote above) and string interpolation in the form of
```
let entry = """
some text..
@{ show $ Rating
  { name = "Haskell"
  , rating = 5
  }
}
some text..
""" in entry
```
As this is probably not an uncommon requirement for a configuration language, I wanted to ask whether it would make sense to add these features.

Gabriel439:
You can now interpolate any expression of type `Text` into a string literal,
like this:

```haskell
let renderRating
    =   λ(rating : { name : Text, rating : Natural })
    →   let score = Integer/show (Natural/toInteger rating.rating)
    in  ''
        # My thoughts about ${rating.name} (${score} out of 5)
        ''
in  renderRating { name = "Dhall", rating = +6 }
```

### DIFF
--- a/src/Dhall/Parser.hs
+++ b/src/Dhall/Parser.hs
@@ -175,7 +175,7 @@ doubleQuoteLiteral embedded = do
     _  <- Text.Parser.Char.char '"'
     go
   where
-    go = go0 <|> go1 <|> go2
+    go = go0 <|> go1 <|> go2 <|> go3
 
     go0 = do
         _ <- Text.Parser.Char.char '"'
@@ -190,6 +190,18 @@ doubleQuoteLiteral embedded = do
         return (TextAppend a b)
 
     go2 = do
+        _ <- Text.Parser.Char.text "'${"
+        b <- go
+        let e = case b of
+                TextLit cs ->
+                    TextLit ("${" <> cs)
+                TextAppend (TextLit cs) d ->
+                    TextAppend (TextLit ("${" <> cs)) d
+                _ ->
+                    TextAppend (TextLit "${") b
+        return e
+
+    go3 = do
         a <- Text.Parser.Token.characterChar
         b <- go
         let e = case b of
@@ -262,9 +274,10 @@ doubleSingleQuoteString embedded = do
 
     p1 =    p2
         <|> p3
-        <|> p4 (Text.Parser.Char.char '\'')
-        <|> p5
-        <|> p4 Text.Parser.Char.anyChar
+        <|> p4
+        <|> p5 (Text.Parser.Char.char '\'')
+        <|> p6
+        <|> p5 Text.Parser.Char.anyChar
 
     p2 = do
         _  <- Text.Parser.Char.text "'''"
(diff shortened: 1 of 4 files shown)


src/Dhall/Parser.hs
```diff
@@ -164,59 +163,156 @@ sepBy1 p sep = do
...
+
+    let process0 (TextAppend (TextLit t) e) =
+            TextAppend (TextLit (trim0 t)) (process1 e)
+        process0 (TextAppend e0 e1) =
+            TextAppend e0 (process1 e1)
+        process0 (TextLit t) =
+            TextLit (trim0 t)
+        process0 e = e
+    
+    return (process0 expr0)
```
anfelor: What about passing the `trim` function?
```haskell
let processWith trim (TextAppend (TextLit t) e) =
        TextAppend (TextLit (trim t)) (processWith trim1 e)
    processWith _ (TextAppend e0 e1) =
        TextAppend e0 (processWith trim1 e1)
    processWith trim (TextLit t) =
        TextLit (trim t)
    processWith _ e = e

return (processWith trim0 expr0)
```

src/Dhall/Parser.hs
```diff
@@ -164,59 +163,156 @@ sepBy1 p sep = do
...
 
-        nonEmptyLines = filter (not . isEmpty) lines_
+    go0 = do
+        _ <- Text.Parser.Char.char '"'
+        return (TextLit mempty) 
 
-        indentLength line =
+    go1 = do
+        _ <- Text.Parser.Char.text "${"
+        a <- exprA embedded
```
Gabriel439: Yes, that's correct.  You can interpolate any arbitrary expression of type `Text`

The example you gave triggered a bug with the parser, though, which is that it wouldn't handle whitespace after `${`, but I just fixed that and now your example works:

```bash
$ dhall
let renderRating
    =   λ(rating : { name : Text, rating : Natural })
    →  ''
        # My thoughts about ${rating.name} (${
                Integer/show (Natural/toInteger rating.rating)
              } out of 5)
        ''
in  renderRating { name = "Dhall", rating = +6 }
<Ctrl-D>
Text

"\n# My thoughts about Dhall (6 out of 5)\n"
```

anfelor:
Looks good to me. :+1:

ruuda:
Would it be possible prepend indentation at every line when interpolating, and strip a trailing newline? There is an open pull request to do this for Nix: https://github.com/NixOS/nix/pull/580. For example, the following
```dhall
let
  inner = 
    ''
    BEGIN
      A
      B
    END
    ''
in
  ''
  ${inner}
  BEGIN
    ${inner}
  END
  ''
```
would turn into
```
BEGIN
  A
  B
END
BEGIN
  BEGIN
    A
    B
  END
END
```
rather than
```
BEGIN
  A
  B
END

BEGIN
  BEGIN
  A
  B
END

END
```

Gabriel439:
@ruuda: Yeah, I think that is reasonable.  The hardest part is just standardizing how it should behave.  Note that the standard hasn't yet standardized the current semantics of multi-line strings so that has to come first, but once that is done the standard.

I'd suggest opening an issue here to request this:

*  https://github.com/dhall-lang/dhall-lang/issues 

That repository is the home for the language standard

PR #67, opened 2017-06-15
Relax bound on trifecta for GHC 8.2

niteria:
trifecta-1.7 is the first version of trifecta
that builds with GHC 8.2 and Cabal 2.0

FWIW the testsuite still passes.


Gabriel439:
Thank you!

PR #79, opened 2017-06-28
Added `normalizeWith` function.

aleator:
Added `normalizeWith` function to complement `typeWith`.
Dhall is so very attractive base for simple DSLs that I felt it
needed to be pushed over the edge.

Naturally, using `normalizeWith` loses all the nice features of Dhall.
If your context isn't total or strongly normalizing then embedding it
to Dhall will not improve things.

I didn't spend too much time testing the change since I have some doubts whether this is an acceptable regarding the original mission of Dhall.

### DIFF
--- a/src/Dhall/Core.hs
+++ b/src/Dhall/Core.hs
@@ -32,6 +32,7 @@ module Dhall.Core (
     , subst
     , shift
     , isNormalized
+    , isNormalizedWith
 
     -- * Pretty-printing
     , pretty
@@ -72,6 +73,8 @@ import qualified Data.Vector
 import qualified Data.Vector.Mutable
 import qualified Filesystem.Path.CurrentOS as Filesystem
 import qualified NeatInterpolation
+
+import Debug.Trace
 
 {-| Constants for a pure type system
 
@@ -1132,7 +1135,7 @@ subst _ _ (Embed p) = Embed p
     leave ill-typed sub-expressions unevaluated.
 -}
 normalize ::  Expr s a -> Expr t a
-normalize = normalizeWith id
+normalize = normalizeWith (const Nothing)
 
 
 {-| Reduce an expression to its normal form, performing beta reduction and applying
@@ -1158,7 +1161,7 @@ normalizeWith ctx e0 = loop (shift 0 "_" e0)
  e'' = bimap (\_ -> ()) (\_ -> ()) e0
 
  text = "NormalizeWith.loop (" <> Data.Text.pack (show e'') <> ")"
- loop e =  case ctx e of
+ loop e =  case e of
     Const k -> Const k
     Var v -> Var v
     Lam x _A b -> Lam x _A' b'
@@ -1283,7 +1286,9 @@ normalizeWith ctx e0 = loop (shift 0 "_" e0)
               where
                 just' y = App just y
                 toMaybe = Data.Maybe.listToMaybe . Data.Vector.toList
-            _ -> App f' a'
+            _ ->  case ctx (App f' a') of
+                    Nothing -> App f' a'
+                    Just app' -> loop app'
           where
             a' = loop a
     Let f _ r b -> loop b''
@@ -1466,7 +1471,15 @@ normalizeWith ctx e0 = loop (shift 0 "_" e0)
 
 -- | Use this to wrap you embedded functions (see `normalizeWith`) to make them
 --   polymorphic enough to be used.
-type CtxFun a = forall s. Expr s a -> Expr s a
+type CtxFun a = forall s. Expr s a -> Maybe (Expr s a)
+
+-- | Check if an expression is in a normal form given a context of evaluation.
+--   Unlike `isNormalized`, this will fully normalize and traverse through the expression. 
+--   
+--   It is much more efficient to use `isNormalized`.
+isNormalizedWith :: (Eq s, Eq a) => CtxFun a -> Expr s a -> Bool
+isNormalizedWith ctx e = e == (normalizeWith ctx e)
+
 
 -- | Quickly check if an expression is in normal form
 isNormalized :: Expr s a -> Bool
--- a/src/Dhall/Core.hs
+++ b/src/Dhall/Core.hs
@@ -27,6 +27,8 @@ module Dhall.Core (
 
     -- * Normalization
     , normalize
+    , normalizeWith
+    , CtxFun(..)
     , subst
     , shift
     , isNormalized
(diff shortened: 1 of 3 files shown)


Gabriel439:
I think this is a great idea, but I would suggest a different approach, which is to change the type of `normalizeWith` to `CtxFun a -> Expr s a -> Expr t a`, and you apply the `CtxFun` at the top-level instead of in the `App` case, like this:

```haskell
normalizeWith :: CtxFun a -> Expr s a -> Expr t a
normalizeWith (CtxFun f) e = case f e of
    ...

normalize = normalizeWith (CtxFun id)
```

The reason why is that you might want to implement functions that cannot be normalized until they are fully saturated.  An example of such a function in Dhall is `List/head`: the first argument is the type to specialize to, and that's not enough to begin the normalization process

The other reason I suggest this change is that I expect this to be more efficient.  Doing a `Map` lookup on every `Var` will decrease the performance of the standard `normalize` function, whereas applying `id` at the top-level is cheap and the compiler will just inline that away.

aleator:
I've experimented a little bit with applying `CtxFun` on the top level. If I understood correctly, there is a slight difficulty (or a decision point) here. On the top level, the argument part of the app is still wrapped in bunch of `Note`s, making it more tricky to actually write the `CtxFun`s. 

Any opinions how this should be handled? Just `(shift 0 .)` the `CtxFun`s before use?

Gabriel439:
Yeah, that would work.  It might also be simpler and more efficient to `shift 0` the entire expression just once before beginning recursion, like this:

```haskell
normalizeWith (CtxFun f) e0 = loop (shift 0 e0)
  where
    loop e = case f e of
        ...
```

src/Dhall/Core.hs
```diff
@@ -1129,38 +1131,64 @@ subst _ _ (Embed p) = Embed p
...
+normalizeWith ctx e = loop (shift 0 "_" e)
+ where
+    -- This is to avoid a `Show` constraint on the @a@ and @s@ in the type of
+    -- `loop`.  In theory, this might change a failing repro case into
+    -- a successful one, but the risk of that is low enough to not warrant
+    -- the `Show` constraint.  I care more about proving at the type level
+    -- that the @a@ and @s@ type parameters are never used
+ e'' = bimap (\_ -> ()) (\_ -> ()) e
+
+ text = "(loop) (" <> Data.Text.pack (show e'') <> ")"
```
Gabriel439: Perhaps expand `loop` here to `normalizeWith.loop`

aleator:
How about now? Seems to work in my use cases quite nicely now.

src/Dhall/Core.hs
```diff
@@ -1375,75 +1403,70 @@ normalize e = case e of
...
-    -- `normalize`.  In theory, this might change a failing repro case into
-    -- a successful one, but the risk of that is low enough to not warrant
-    -- the `Show` constraint.  I care more about proving at the type level
-    -- that the @a@ and @s@ type parameters are never used
-    e'' = bimap (\_ -> ()) (\_ -> ()) e
 
-    text = "normalize (" <> Data.Text.pack (show e'') <> ")"
+-- | Use this to wrap you embedded functions (see `normalizeWith`) to make them
+--   polymorphic enough to be used.
+type CtxFun a = forall s. Expr s a -> Expr s a
```
Gabriel439: Perhaps inline the `CtxFun` type synonym into the type of `normalizeWith` since it's only used once

src/Dhall/Core.hs
```diff
@@ -1129,38 +1131,64 @@ subst _ _ (Embed p) = Embed p
     However, `normalize` will not fail if the expression is ill-typed and will
     leave ill-typed sub-expressions unevaluated.
 -}
-normalize :: Expr s a -> Expr t a
-normalize e = case e of
+normalize ::  Expr s a -> Expr t a
+normalize = normalizeWith (id)
```
Gabriel439: Parentheses are not necessary around the `id`

aleator: Fixed. I also applied this to other places where search/replace has ended up with extra parentheses.

src/Dhall/Core.hs
aleator: Ah. Of course.

src/Dhall/Core.hs
aleator: I left this alone, since it saved some typing when writing tests.

aleator:
Here is a new version of this. However, I hit a snag when trying to write tests. Should there be `isNormalized` that corresponds with `normalizeWith`, since it now also may depend on the context of normalization?

src/Dhall/Core.hs
Gabriel439: In that case, maybe rename it to something like `Normalizer` since it no longer is part of a context any longer

Gabriel439:
Yeah, you would need an `isNormalizedWith` function

aleator:
What would happen with a context that can diverge? 

I think it might be best to leave users of `normalizeWith` to come up with their own solutions to this and just write the tests differently when `CtxFun /= id`

aleator:
I've now moved the application of CtxFun back to App-case from the top level. The previous attempt at this didn't work out, since the normalization was cut short. Also, having all of the ast available to CtxFun seems like an overkill ("Hey, lets go and replace all True's with 42's"). When it is applied as the last case of App you know that it won't be likely to mess up pre-existing stuff that is outside of the capture App. 

There are also two tests, which demonstrate how this works.

Gabriel439:
This seems fine.  My only remaining suggestion is to change the type of the function back to `forall a . Expr s a -> Expr s a`, and change the logic to:

```haskell
_ ->  ctx (App f' a')
```

... and then `normalizeWith` becomes:

```haskell
normalize = normalizeWith id
```

... for the same reason as before: efficiency

tests/Normalization.hs
```diff
@@ -29,6 +31,36 @@ conversions = testGroup "conversions" [ naturalShow
...
+  let tyCtx  = insert "min" (Pi "_" Natural (Pi "_" Natural Natural)) empty 
+      valCtx e = case e of
+                    (App (App (Var (V "min" 0)) (NaturalLit x)) (NaturalLit y)) -> Just (NaturalLit (min x y))
+                    _ -> Nothing
+  e <- codeWith tyCtx "min (min +11 +12) +8 + +1" 
+  assertNormalizesToWith valCtx e "+9"
+
+doubleReduction :: TestTree
+doubleReduction = testCase "doubleReduction" $ do
+  let tyCtx  =  insert "min" (Pi "_" Natural (Pi "_" Natural Natural))
```
Gabriel439: Minor suggestion: you can use `code` to input these types:

```haskell
minType        <- code "Natural → Natural → Natural"
fiveorlessType <- code "Natural → Natural"
wurbleType     <- code "Natural → Integer
let tyCtx = insert "min"        minType
          . insert "fiveorless" fiveorlessType
          . insert "wurble"     wurbleType
          $ empty
```

I don't feel very strongly about this suggestion, so feel free to ignore me :)

src/Dhall/Core.hs
```diff
@@ -1375,75 +1408,78 @@ normalize e = case e of
...
-    -- `normalize`.  In theory, this might change a failing repro case into
-    -- a successful one, but the risk of that is low enough to not warrant
-    -- the `Show` constraint.  I care more about proving at the type level
-    -- that the @a@ and @s@ type parameters are never used
-    e'' = bimap (\_ -> ()) (\_ -> ()) e
 
-    text = "normalize (" <> Data.Text.pack (show e'') <> ")"
+-- | Use this to wrap you embedded functions (see `normalizeWith`) to make them
+--   polymorphic enough to be used.
+type CtxFun a = forall s. Expr s a -> Maybe (Expr s a)
```
Gabriel439: I still suggest naming this something like `Normalizer` instead of `CtxFun` since it no longer uses a context

aleator: Sorry. I missed your original comment. Will do!

tests/Normalization.hs
aleator: Looks better that way. Will do.

aleator:
I don't think `_ ->  ctx (App f' a')` would work. Suppose that we have a custom App (matching, say, `App (Var "foo") (NaturalLit x)`) sitting at the top level, and we hit this case. Then `ctx` turns this into `(App (Var "bar") (NaturalLit x)` and the normalizer terminates.

What if the `ctx` contains a definition for  `(App (Var "bar") (NaturalLit x)` as well? 

We also can't loop here without looping forever. We need to know if `ctx` actually did something before continuing. 

I can probably whip up a criterion for testing how much this slows things down if you are really worried. Can you think of a good test case? (Normalize the prelude? The tutorial?).

Gabriel439:
Oh yeah, I hadn't though about the need to continue looping.  Then I think it's fine as is

I think the performance impact will probably be minimal since it only affects the case where normalization is blocked, which is not that common and not on the critical path

PR #82, opened 2017-07-02
add Type and Interpret instance for strict Text

chris-martin:
This instance is nice to have if you want `Generic`-derived `Interpret` instances for types where your `Text` doesn't happen to be of the lazy variety.

### DIFF
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -29,7 +29,8 @@ module Dhall
     , natural
     , integer
     , double
-    , text
+    , lazyText
+    , strictText
     , maybe
     , vector
     , GenericInterpret(..)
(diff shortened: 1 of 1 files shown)


src/Dhall.hs
```diff
@@ -344,6 +344,14 @@ text = Type {..}
 
     expected = Text
 
+{-| Decode strict `Text`
+
+>>> input text' "\"Test\""
+"Test"
+-}
+text' :: Type Data.Text.Text
```
Gabriel439: Minor suggestion: perhaps call this `strictText` and rename the old one to `lazyText`.  It's okay if it's a breaking change

chris-martin: Sounds good.

chris-martin:
Oh, I also forgot to export `text'`, of course.

chris-martin:
Updated, now has `strictText` and `lazyText`.

Gabriel439:
Thanks for adding this! :)

Issue #83, opened 2017-07-03
Getting best mileage with `typeWith`?


aleator:
I'm playing with Dhall (probably in ways it wasn't supposed to be used) and I'm having troubles with the `typeWith` function. What I would want is to have few opaque type name that can be referred to in my config. For most part this works fine, except that `load` does typechecking. This means that I cannot refer to my opaque types in anything that has been imported.

The above is probably by design, but it certainly makes `typeWith` less useful for the end user.  Would a patch that enables passing type context into `load` be acceptable?

Gabriel439:
Yes, I agree that this is worth supporting

Gabriel439:
Closing, as I think this is fixed by #84

PR #84, opened 2017-07-04
Added a `loadWith` function.

aleator:
See #83.

### DIFF
--- a/src/Dhall/Import.hs
+++ b/src/Dhall/Import.hs
@@ -158,6 +158,7 @@ import qualified Data.Text.Lazy.Encoding
 import qualified Data.Vector
 import qualified Dhall.Core
 import qualified Dhall.Parser
+import qualified Dhall.Context
 import qualified Dhall.TypeCheck
 import qualified Filesystem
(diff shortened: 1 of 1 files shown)


src/Dhall/Import.hs
```diff
@@ -674,7 +676,10 @@ loadDynamic p = do
 
 -- | Load a `Path` as a \"static\" expression (with all imports resolved)
 loadStatic :: Path -> StateT Status IO (Expr Src X)
-loadStatic path = do
+loadStatic = loadStaticWith Dhall.Context.empty
+
+loadStaticWith :: Dhall.Context.Context (Expr Src X) -> Path -> StateT Status IO (Expr Src X)
+loadStaticWith ctx path = do
```
Gabriel439: I think you also need to change the recursive call to `loadStatic` inside this function to use `loadStaticWith`, so that the context is propagated to transitive imports

aleator: Oops.

Gabriel439:
Thank you for contributing this, too! :)

PR #85, opened 2017-07-04
Raw input

aleator:
Working with `normalizeWith` (see #79) seems to necessitate converting Dhall values to Haskell quite often. However, the existing machinery for this (ie. `input`) is not adequate to the task, since it works with `Text` instead of `Expr`'s.

I would made a new module like `Dhall.Customization` for this function, but that would have required restructuring the `Dhall` module or exposing the `Type` type. Having this in `Dhall` may just confuse new users...

### DIFF
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -16,6 +16,7 @@ module Dhall
     (
     -- * Input
       input
+    , rawInput
     , detailed
 
     -- * Types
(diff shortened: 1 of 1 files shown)


aleator:
Urgh. I messed up the branch. Will fix in a minute.

aleator:
Fix'd

src/Dhall.hs
```diff
@@ -16,6 +16,7 @@ module Dhall
     (
     -- * Input
       input
+    , rawInput
```
Gabriel439: My only suggestion is to move this closer to the bottom of the export list under a `-- * Miscellaneous` header before the `-- * Re-exports` section since most people won't be using this function

aleator: Done.

Gabriel439:
Thanks! :)

PR #86, opened 2017-07-10
Expose Haskell type expression.

bosu:
When working in Dhall AST level it allows to expose Haskell values in
the Context as implicit variables.

### DIFF
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -19,7 +19,7 @@ module Dhall
     , detailed
 
     -- * Types
-    , Type
+    , Type(..)
     , Interpret(..)
     , InvalidType(..)
     , auto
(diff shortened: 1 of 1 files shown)


Gabriel439:
How about just exposing the internals of the `Type` constructor?  I think at this point it's stable enough that I feel fine exporting the entire type

bosu:
I am certainly ok with it! Do you want me to send a trivial pull request?

What about the names: are `expected` and `extract` good enough?

Gabriel439:
Yeah, I think it's fine to keep the existing accessor names

bosu:
Done. Please see the amended pull request.

Gabriel439:
Awesome!  Thank you :)

PR #91, opened 2017-07-27
Export InputType

bosu:
Apparently this export was forgotten. Hopefully the intention was to fully export `InputType`.

IMHO the reasons for full export are:

* Symmetry with the already exported `Type` :)
* It allows to pass Haskell values through the `normalizeWith`


Gabriel439:
Oops!  Yes, this was an oversight of mine.  Thanks for catching this and fixing it :)

PR #93, opened 2017-07-31
Ability to load expressions purely.

bosu:
* Refactor `loadStaticWith` to work in `MonadCatch` monad
using `exprFromPath` as a callback.

* Add `load'`, `loadWith'` functions to allow resolving
Path expressions using callback in any `MonadCatch` monad.

* Consolidate `StateT Status` evaluation into internal `evalStatus`
function.

As proposed in https://github.com/Gabriel439/Haskell-Dhall-Library/issues/92.

### DIFF
--- a/src/Dhall/Import.hs
+++ b/src/Dhall/Import.hs
@@ -1,7 +1,8 @@
-{-# LANGUAGE CPP                #-}
-{-# LANGUAGE DeriveDataTypeable #-}
-{-# LANGUAGE OverloadedStrings  #-}
-{-# LANGUAGE RecordWildCards    #-}
+{-# LANGUAGE CPP                 #-}
+{-# LANGUAGE DeriveDataTypeable  #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE RecordWildCards     #-}
 {-# OPTIONS_GHC -Wall #-}
 
 {-| Dhall lets you import external expressions located either in local files or
@@ -112,9 +113,11 @@ module Dhall.Import (
 
 import Control.Applicative (empty)
 import Control.Exception
-    (Exception, IOException, SomeException, catch, onException, throwIO)
+    (Exception, IOException, SomeException, onException, throwIO)
 import Control.Lens (Lens', zoom)
 import Control.Monad (join)
+import Control.Monad.Catch (throwM, MonadCatch(catch))
+import Control.Monad.Trans.Class (lift)
 import Control.Monad.IO.Class (MonadIO(..))
 import Control.Monad.Trans.State.Strict (StateT)
 import Data.ByteString.Lazy (ByteString)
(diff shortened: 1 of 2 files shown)


src/Dhall/Import.hs
```diff
@@ -103,6 +104,8 @@ module Dhall.Import (
       exprFromPath
     , load
     , loadWith
+    , load'
+    , loadWith'
```
Gabriel439: What do you think about just exporting two load functions:

* `load` (with no custom context and runs in `IO`)
* `loadWith` (with custom context and runs in `MonadCatch`, which you currently call `loadWith'`)

My reasoning here is to reduce API proliferation.  Also, people who want to customize the context will probably also want to customize the path import logic, too

This would be a breaking change (since it changes the type of `loadWith` which is already part of the public API), but I'd be fine with that since the next release has a minor breaking change anyway

src/Dhall/Import.hs
```diff
@@ -662,21 +667,35 @@ exprFromPath m (Path {..}) = case pathType of
...
+    loadStaticWith (exprFromPath m) ctx path
+
+-- | Resolve all imports within an expression using `Path` resolving callback
+-- in arbitrary `MonadCatch` monad.
+load' :: MonadCatch m => (Path -> m (Expr Src Path)) -> Expr Src Path -> m (Expr Src X)
+load' from_path = loadWith' from_path (Dhall.Context.empty)
+
+-- | Resolve all imports within an expression using a custom typing context and Path
+-- resolving callback in arbitrary `MonadCatch` monad.
+loadWith' :: MonadCatch m => (Path -> m (Expr Src Path))
```
Gabriel439: For these generalized functions that run in `MonadCatch`, I think you can generalize their types to replace the `Path` with a type variable.  For example, I think you could change the type of this function to:

```haskell
loadWith
    :: MonadCatch m
    => (a -> m (Expr Src a))
    -> Dhall.Context.Context (Expr Src X) -> Expr Src a -> m (Expr Src X)
```

The main motivation for this generalization is to provide built-in imports that are not `Path`s.  For example, going back to the original motivation behind this pull request, you wanted to be able to support an import like this:

```haskell
\(x : Big) -> ...
```

... but with the current pull-request that wouldn't work because `Big` will not parse as a `Path`, so the user would have to provide a path like `./Big` instead, which might not be what you want.  However, if you were to generalize `loadWith'` to work with things other than `Path` (with matching changes to the parser) then you can support your original example

Note that the `Dhall.Parser` module already supports generalizing the parser to parse things other than `Path`, too, thanks to the `exprA` function:

```haskell
exprA :: Show a => Parser a -> Parser (Expr Src a) 
```

src/Dhall/Import.hs
bosu: Implemented in the updated pull request.

src/Dhall/Import.hs
bosu: > For these generalized functions that run in MonadCatch, I think you can generalize their types to replace the Path with a type variable.

It looks like significant parts of `Dhall.Import` deal with `Path` data type itself. E.g. see `canonicalizePath`, local path handling in `loadStaticWith` etc. Doing `loadWith` with generic
type would require having this functionality in some sort of callback...

> ... but with the current pull-request that wouldn't work because Big will not parse as a Path, so the user would have to provide a path like ./Big instead, which might not be what you want. However, if you were to generalize loadWith' to work with things other than Path (with matching changes to the parser) then you can support your original example

I am indeed doing the dirty `./Big` hack currently.

To fix it properly, I would suggest changing `Path` to be `Path a` and adding something like `Custom a` constructor. Then we can leverage `exprA` parsing to parse some kind of `custom:Big` special `Path`. Also the `custom` prefix can be something which the user defines.

IMHO, this kind of change is better done independently from the current pull request.

Gabriel439:
Yeah, I agree that fixing `./Big` should be done in a separate pull request

This looks great!  I'll go ahead and merge

Gabriel439:
Also, for future reference you don't need to rebase or force push when submitting pull requests.  GitHub has a new feature to squash multiple commits into a single clean commit when merging

PR #95, opened 2017-08-04
Add Eq instance for Src

bosu:
It is convenient to check equality of `Expr Src X` sometimes.


Gabriel439:
Oops!  Thanks for catching that.  Yeah, I agree that `Expr Src X` should be comparable

PR #100, opened 2017-08-20
Add Inject instance for tuples

bosu:
`(a, b)` is mapped to `{ _1 = a, _2 = b }`

### DIFF
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE DefaultSignatures   #-}
+{-# LANGUAGE DeriveAnyClass      #-}
 {-# LANGUAGE DeriveFunctor       #-}
+{-# LANGUAGE DeriveGeneric       #-}
 {-# LANGUAGE FlexibleInstances   #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE OverloadedStrings   #-}
@@ -22,6 +24,7 @@ module Dhall
     , InputType(..)
     , Interpret(..)
     , InvalidType(..)
+    , R2(..)
     , auto
     , InterpretOptions(..)
     , defaultInterpretOptions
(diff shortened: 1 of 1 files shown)


Gabriel439:
What I'd suggest doing here is to first define an `R2` type, like this:

```haskell
data R2 a b = R2 { _1 :: a, _2 :: b }
    deriving (Generic, Inject, Interpret)
```

... and then defining the `Inject` instance for the tuple by taking advantage of the `Contravariant` instance for `InputType`:

```haskell
instance (Inject a, Inject b) => Inject (a, b) where
    injectWith = fmap (contraMap adapt) injectWith
      where
        adapt (_1, _2) = R2 {..})
```

The advantage of this approach is that it's easy to scale out to larger tuple sizes

The other advantage is that you can also export the `R{N}` family of constructors so that users can also use those directly instead of tuples if they want the correspondence to the Dhall type to be more clear

Gabriel439:
Also, while you're doing this, you can also define a matching `Interpret` instance for tuples, too

bosu:
> What I'd suggest doing here is to first define an R2 type, like this:

You are right, this looks much better. I'll update the pull request.

> Also, while you're doing this, you can also define a matching Interpret instance for tuples, too

Will do.

bosu:
Rebased and updated the pull request.

src/Dhall.hs
```diff
@@ -103,6 +105,12 @@ instance Show InvalidType where
 
 instance Exception InvalidType
 
+-- | Dhall uses this type to represent (a, b) tuple.
+data R2 a b = R2 { _1 :: a, _2 :: b } deriving (Show, Generic)
+
+instance (Inject a, Inject b) => Inject (R2 a b)
```
Gabriel439: If you add the `DeriveAnyClass` extension, then you can simplify this further as just:

```haskell
data R2 a b = R2 { _1 :: a, _2 :: b }
    deriving (Generic, Inject, Interpret, Show)
```

bosu:
Update pull request uses `DeriveAnyClass`

Gabriel439:
I'll go ahead and merge this and then fix the build failure on `master` separately since this pull request is not the cause of the problem

PR #103, opened 2017-08-21
Interpret instance for unnamed data fields

bosu:
Maps unnamed data fields to the positional record keys. For example:

`data Foo = Foo Bool Integer Natural deriving (Generic, Interpret)`

is mapped to:

`Record { _1 : Bool, _2 : Integer, _3 : Natural }`

Dhall data type.

### DIFF
--- a/dhall.cabal
+++ b/dhall.cabal
@@ -144,6 +144,7 @@ Test-Suite test
         Util
     Build-Depends:
         base               >= 4        && < 5   ,
+        containers         >= 0.5.0.0  && < 0.6 ,
         dhall                                   ,
         tasty              >= 0.11.2   && < 0.12,
         tasty-hunit        >= 0.9.2    && < 0.10,
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -53,6 +53,7 @@ module Dhall
 
 import Control.Applicative (empty, liftA2, (<|>), Alternative)
 import Control.Exception (Exception)
+import Control.Monad.Trans.State.Strict
 import Data.Functor.Contravariant (Contravariant(..))
 import Data.Monoid ((<>))
 import Data.Text.Buildable (Buildable(..))
@@ -434,7 +435,7 @@ class Interpret a where
     autoWith:: InterpretOptions -> Type a
     default autoWith
         :: (Generic a, GenericInterpret (Rep a)) => InterpretOptions -> Type a
-    autoWith options = fmap GHC.Generics.to (genericAutoWith options)
+    autoWith options = fmap GHC.Generics.to (evalState (genericAutoWith options) 1)
 
 instance Interpret Bool where
     autoWith _ = bool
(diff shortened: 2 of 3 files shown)


src/Dhall.hs
```diff
@@ -529,104 +544,88 @@ instance (Constructor c1, Constructor c2, GenericInterpret f1, GenericInterpret
...
 
 instance (Selector s, Interpret a) => GenericInterpret (M1 S s (K1 i a)) where
-    genericAutoWith opts@(InterpretOptions {..}) = Type {..}
+    genericAutoWith opts@(InterpretOptions {..}) = do
+        name <- case selName n of
+                    "" -> do i <- get
+                             put (i + 1)
+                             pure ("_" ++ show i)
+                    nn -> pure nn
+        pure (Type { expected = expected_name name, extract = extract_name name })
```
Gabriel439: Minor suggestion: you can simplify this a little bit by relocating `expected` and `extract` inside the `do` block, like this:

```haskell
name <- case ...
let extract (RecordLit m) = ...
let expected = ...
pure (Type {..})
```

tests/Regression.hs
```diff
@@ -12,8 +19,25 @@ regressionTests :: TestTree
...
+          , unnamedFields
         ]
 
+data Foo = Foo Integer Bool | Bar Bool Bool Bool | Baz Integer Integer
+    deriving (Show, Dhall.Generic, Dhall.Interpret)
+
+unnamedFields :: TestTree
+unnamedFields = Test.Tasty.HUnit.testCase "Unnamed Fields" (do
+    let ty = Dhall.auto @Foo
+    Test.HUnit.assertEqual "Good type" (Dhall.expected ty) (Dhall.Core.Union (
```
Gabriel439: I believe `Test.Tasty.Hunit` also provides `assertEqual`, so you don't need to pull in the additional `HUnit` dependency

(To be honest, I'm not even sure what's the difference between the two libraries.  I didn't originally author the test suite)

dhall.cabal
```diff
@@ -144,6 +144,8 @@ Test-Suite test
         Util
     Build-Depends:
         base               >= 4        && < 5   ,
+        containers         >= 0.5.7.1  && < 0.6 ,
```
Gabriel439: You can lower the lower bound on `containers` to `>= 0.5.0.0` (consistent with the `containers` dependency for the main library) since that version still has `Data.Map.fromList`

src/Dhall.hs
```diff
@@ -505,20 +506,34 @@ defaultInterpretOptions = InterpretOptions
     for automatically deriving a generic implementation
 -}
 class GenericInterpret f where
-    genericAutoWith :: InterpretOptions -> Type (f a)
+    genericAutoWith :: InterpretOptions -> State Int (Type (f a))
```
bosu: The amended commit much simplified the diff. IMHO Reader may be unnecessary anymore...

tests/Regression.hs
```diff
@@ -1,7 +1,14 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
```
bosu: +1 for dropping GHC 7.8

bosu:
Sorry for losing the remarks by using amended push request. All of them were addressed. Now I see the advantages of additional commits :)

Regarding R2 removal: it is still needed by Inject instance... Probably the same kind of refactoring can happen there.

Gabriel439:
Thanks! :)

PR #107, opened 2017-08-23
Implement Inject instance for unnamed tuples.

bosu:
Implementation is similar to https://github.com/Gabriel439/Haskell-Dhall-Library/pull/103.

Last patch in the series shows that `R2` is necessary: generic instance is not being picked up for tuples.


Gabriel439:
Actually, you don't need `R2` if you enable the `StandaloneDeriving` extension.  Then you can simplify the two instances for tuples to:

```haskell
deriving instance (Interpret a, Interpret b) => Interpret (a, b)

deriving instance (Inject a, Inject b) => Inject (a, b)
```

If you do that then you delete everything `R2`-related

bosu:
Done.

Gabriel439:
Awesome!  Thanks for doing this :)

PR #113, opened 2017-08-28
Inject instances for Data.Set, Data.Sequence

bosu:


Gabriel439:
Thanks!

PierreR:
Hope you won't mind a clueless question. Does all these new instances provide any benefits for a lambda user ?

Gabriel439:
@PierreR: Yes.  The `Inject` instance affects what you can provide as an argument to a lambda marshalled from Dhall to Haskell

That means that this would now be legal after this change:

```haskell
example = do
    function <- input auto "List/head Integer" :: IO (Set Integer -> Maybe Integer)
    print (function Data.Set.empty) -- "Nothing"
    print (function (Data.Set.fromList [1, 2])) -- "Just 1"
```

PR #116, opened 2017-08-29
Add bad variable name to UnboundVariable

bosu:
Currently UnboundVariable type message is static, even
in detailed mode.

Add bad variable name to the UnboundVariable and show it
in both short and detailed modes.

### DIFF
--- a/src/Dhall/TypeCheck.hs
+++ b/src/Dhall/TypeCheck.hs
@@ -125,7 +125,7 @@ typeWith _     (Const c         ) = do
     fmap Const (axiom c)
 typeWith ctx e@(Var (V x n)     ) = do
     case Dhall.Context.lookup x n ctx of
-        Nothing -> Left (TypeError ctx e UnboundVariable)
+        Nothing -> Left (TypeError ctx e (UnboundVariable x))
         Just a  -> return a
 typeWith ctx   (Lam x _A  b     ) = do
     let ctx' = fmap (Dhall.Core.shift 1 (V x 0)) (Dhall.Context.insert x _A ctx)
@@ -653,7 +653,7 @@ instance Buildable X where
 
 -- | The specific type error
 data TypeMessage s
-    = UnboundVariable
+    = UnboundVariable Text
     | InvalidInputType (Expr s X)
     | InvalidOutputType (Expr s X)
     | NotAFunction (Expr s X) (Expr s X)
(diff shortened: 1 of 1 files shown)


Gabriel439:
Actually, there is a reason why it doesn't include the variable name, which is that the variable name is already included immediately below in the source extract:

```
$ dhall <<< 'x'

Error: Unbound variable

x

(stdin):1:1
```

bosu:
> Actually, there is a reason why it doesn't include the variable name, which is that the variable name is already included immediately below in the source extract:

Yes, I was generating bad AST in Haskell, had no source to work with, therefore have not seen the variable name.

I admit that this use case is a bit esoteric :). Do you think that printing the variable twice is too bad?
Here is how it looks now:

```
$ echo 'x' | dhall

Error: Unbound variable: x

x

(stdin):1:1
```


FWIW, I am ok with discarding the pull request also.

Gabriel439:
So I would probably not change the error message, and I'll try to explain my reasoning process

The short error messages are designed for the user to read them from the bottom up.  In most cases (if the user's terminal is saturated) the output of the `dhall` command will be aligned to the bottom of the terminal window right above their command line, so anything closer to the bottom of the output is easier for the user's eye to predictably jump to.  The short message organizes information to put the most important information at the bottom and then relevance decreases as you go up.

For example, the line and column number is at the very bottom since I consider that information the most relevant to the user.  Specifically, if you were to omit all other error output and just produced the line and column number the user would probably still be able to figure out what went wrong by just opening up their editor and jumping to that source location.

Similarly, the source highlight comes immediately above that because a quick glance at the offending source code can often illuminate the problem quickly

Then after that comes the error message and it's designed to not have any variable component so that the user's eye can visually "pattern match" on the error string

So the reason that I would prefer to leave the error message unmodified is:

* it trains the user to scan the error from the bottom up (by putting the unbound variable name closer to the bottom)

* it decreases variation in the error message so that the user can visually pattern match more effectively on the error message

bosu:
Thanks for a long and detailed explanation! I understand and agree with your reasoning.

Would you accept propagating the name in UnboundVariable constructor without changing the error messages? I.e keeping the upper two hunks in the patch and leaving `prettyTypeMessage` as is?
This way interested parties can still dig the variable name if needed...

There may be comment on `prettyTypeMessage` explaining the lack of name printing there.

Gabriel439:
Yeah, I think including the variable in the constructor is still a great idea so that people using the API don't have to extract the bound variable name

bosu:
Done.

Gabriel439:
Thanks! :)

PR #121, opened 2017-09-03
Fix Tutorial mandatory type description for List

Profpatsch:
Only empty lists need annotation in newer versions of dhall.


Gabriel439:
Thanks! :)

PR #128, opened 2017-09-03
Strip leading newline from multi-line strings

### LINKED ISSUE #126: multiline strings always start with \n or do not strip
```
''
  foo
  bar
''
```

gives `"\nfoo\nbar\n"`.

That does not correspond to the nix interpretation, which results in `"foo\nbar\n"`.

If one wants to avoid the starting newline, one loses stripping:

```
  ''foo
    bar
''
```

results in `"foo\n    bar\n"` (same with nix).

Since one can always insert a prefixing `\n`, I suppose we align our multiline strings with nix.

Gabriel439:
Fixes #126

This means that:

```haskell
''
  foo
  bar
''
```

... will now be treated as:

```haskell
"foo\nbar\n"
```

... and not:

```haskell
"\nfoo\nbar\n"
```

The goal of this is to match the behavior of Nix (and also for user
convenience)

### DIFF
--- a/src/Dhall/Parser.hs
+++ b/src/Dhall/Parser.hs
@@ -264,6 +264,7 @@ doubleSingleQuoteString embedded = do
 
     p0 = do
         _ <- Text.Parser.Char.string "''"
+        _ <- optional (Text.Parser.Char.char '\n')
         p1
 
     p1 =    p2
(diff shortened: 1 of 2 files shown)


Profpatsch:
That looks like the behaviour matches nix now, yes.

PR #131, opened 2017-09-09
Replace Integer with Natural

chepner:
`Natural/fold` takes a `Natural` value, not an `Integer` value, as its first argument.


Gabriel439:
Nice catch!  Thank you for fixing that :)

Issue #10, opened 2016-12-10
Type annotation cannot refer bindings in scope


esmolanka:
In "fixed" example from #9 different kind of error occurs to the user. The type checker cannot determine that `{x : Integer, y : Integer} == Point`.

```
$ ./dhall <<< "let Point = {x : Integer, y : Integer} in ({x = 42, y = -1} : Point)"

Use "dhall --explain" for detailed errors

Point : Type

Error: Expression doesn't match annotation

({x = 42, y = -1} : Point)

(stdin):1:43
```

Being able to declare type aliases and then use them in type annotations is especially useful when it comes to List or Optional type, because literals of such types always require type annotations. Duplicating type of large record does not seem practical.

Gabriel439:
So this is actually intended behavior, surprising as that may be.

# Short explanation

Type-checking precedes normalization and normalization includes substitution of `let`-bound types

# Recommended solution

You can use the import system to solve this problem:

```
$ echo "{ x : Integer, y : Integer }" > ./Point
$ dhall <<< "{x = 42, y = -1} : ./Point"
{ x : Integer, y : Integer }

{ x = 42, y = -1 }
```

# Long explanation

Dhall has a rule that:

```
let x : t = y in e
```

... must be equivalent to this

```
(λ(x : t) → e) y
```

... both for normalization purposes and type-checking purposes.  That means that:

* Both expressions should normalize to the same result
* If one expression type-checks then the other expression must type-check
* If one expression fails to type-check then the other expression must fail to type-check

That means that this `let` expression

```
let Point : Type = {x : Integer, y : Integer} in ({x = 42, y = -1} : Point)
```

... must behave the same as this equivalent `λ` expression:

```
(λ(Point : Type) → ({ x = 42, y = -1 } : Point)) { x : Integer, y : Integer }
```

... which does not type-check in System Fω (the type system that Dhall is based on), because this sub-expression does not type-check in isolation:

```
λ(Point : Type) → ({ x = 42, y = -1 } : Point)
```

You can try to type-check either of those expressions in Dhall and you will get the same result:

```
$ dhall
(λ(Point : Type) → ({ x = 42, y = -1 } : Point)) { x : Integer, y : Integer }
^D
Use "dhall --explain" for detailed errors

Point : Type

Error: Expression doesn't match annotation

({ x = 42, y = -1 } : Point)

(stdin):1:20
```

[...]

PR #134, opened 2017-09-13
Fix unbound variable bug

Gabriel439:
Related to #10

When type-checking the following expression:

```haskell
let Day : Type = Natural in λ(x : Day) → x
```

... Dhall had a bug where it would add `Day : Type` to the context when
inferring the type of `λ(a : Day) → x` (which is `∀(x : Day) → Day`), but `Day`
was missing from the context when inferring the kind of `∀(x : Day) → Day`,
leading to an unbound variable error

This change fixes the bug and adds a regression test to prevent the bug from
recurring

### DIFF
--- a/tests/Regression.hs
+++ b/tests/Regression.hs
@@ -85,6 +85,18 @@ parsing0 = Test.Tasty.HUnit.testCase "Parsing regression #0" (do
         Left  e -> Control.Exception.throwIO e
         Right _ -> return () )
 
+typeChecking0 :: TestTree
+typeChecking0 = Test.Tasty.HUnit.testCase "Type-checking regression #0" (do
+    -- There used to be a bug in the type-checking logic where `T` would be
+    -- added to the context when inferring the type of `λ(x : T) → x`, but was
+    -- missing from the context when inferring the kind of the inferred type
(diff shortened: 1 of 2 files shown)


tests/Regression.hs
```diff
@@ -85,6 +85,18 @@ parsing0 = Test.Tasty.HUnit.testCase "Parsing regression #0" (do
...
+typeChecking0 = Test.Tasty.HUnit.testCase "Type-checking regression #0" (do
+    -- There used to be a bug in the type-checking logic where `T` would be
+    -- added to the context when inferring the type of `λ(x : T) → x`, but was
+    -- missing from the context when inferring the kind of the inferred type
+    -- (i.e. the kind of `∀(x : T) → T`).  This led to an unbound variable
+    -- error when inferring the kind
+    --
+    -- This bug was originally reported in issue #10
+    _ <- Util.code "let Day : Type = Natural in λ(x : Day) → x"
+    return ()
```
bosu: There is a parenthesis missing here (from line 89). This is the cause for Travis build check.

+1 on the fix itself, of course.

Gabriel439:
Whoops!  You can tell I was in a hurry and couldn't run tests locally this morning :)

PR #137, opened 2017-09-17
Add `dhall-format` executable

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/8

This takes a Dhall expression on standard input and outputs the same
expression to standard output except pretty-printed in a standard format


dhall-format/Main.hs
```diff
@@ -0,0 +1,80 @@
...
+import qualified Data.Text.IO
+import qualified Data.Text.Lazy
+import qualified Data.Text.Lazy.IO
+import qualified Data.Text.Prettyprint.Doc             as Pretty
+import qualified Data.Text.Prettyprint.Doc.Render.Text as Pretty
+import qualified Filesystem.Path.CurrentOS
+import qualified Options.Generic
+import qualified System.IO
+
+data Mode = Default | Resolve | TypeCheck | Normalize
```
jml: Looks to be unused?

dhall-format/Main.hs
```diff
@@ -0,0 +1,80 @@
+{-# LANGUAGE DataKinds          #-}
+{-# LANGUAGE DeriveGeneric      #-}
+{-# LANGUAGE ExplicitNamespaces #-}
+{-# LANGUAGE OverloadedStrings  #-}
+{-# LANGUAGE TypeOperators      #-}
```
jml: In non-Haskell stuff, I like scripts to have doc comments at the top saying:
- what it is
- why you might want to use it
- an example of how to use it

I like this in Haskell stuff too, but I always forget how to write the correct Haddock markup. ;)

src/Dhall/Core.hs
```diff
@@ -451,6 +457,532 @@ instance IsString (Expr s a) where
...
+angles docs = enclose "< " "< " " | " "| " " >" ">" docs
+
+braces :: [(Doc ann, Doc ann)] -> Doc ann
+braces   [] = "{}"
+braces docs = enclose "{ " "{ " ", " ", " " }" "}" docs
+
+arrows :: [(Doc ann, Doc ann)] -> Doc ann
+arrows = enclose' "" "  " " → " "→ " 
+
+enclose
```
jml: Personally would put a doc comment here.

src/Dhall/Core.hs
```diff
@@ -451,6 +457,532 @@ instance IsString (Expr s a) where
...
+  where
+    docsShort = fmap fst docs
+
+    docsLong = fmap snd docs
+
+    combineLong x y = x <> y <> Pretty.hardline
+
+    combineShort x y = x <> y
+
+enclose'
```
jml: And a doc comment here too.

Issue #145, opened 2017-09-29
dhall-format removes comments


psibi:
Example file:

```
--- This is dhall file
{ foo = "jax" }
```

The Output it gives:

```
{ foo = "jax" }
```

psibi:
Going through the code, it seems a known issue. :) Anyway keeping this open.

Gabriel439:
One thing I can do as an immediate workaround is to have the formatter preserve leading whitespace and comments at the beginning of the file.  Then if you need comments for a subexpression you can at least have the workaround of splitting it out into another file

f-f:
@Gabriel439 What's the main challenge to tackle in order to fix this? As far as I remember the main problem is that it'd be hard to associate comment fragments to other AST nodes, right? (as in: a comment line that comes before some code goes with the next node, but what to do in case of a comment fragment on the same line? Also how about the alignments? And does this mean that comments would slow down all operations, since it's a whole lot of packing-unpacking dummy AST comment nodes?)

ocharles:
The GHC approach is not to out comments in the AST, but have a separate map
Map (NodeType, SrcLoc) Annotation. So as you traverse the AST you have
SrcLocs and can find annotations for the node you're at. That's one
approach. Or we could have a trees that grow approach and erase comments
from the AST for anything but formatting.

Gabriel439:
There is also a third solution, which is a "low-tech" formatter that doesn't actually parse the AST but rather just scans the text and formats as it goes.  I believe this is how [`go fmt`](https://golang.org/src/fmt/print.go) works.

However, I think the simplest and most robust solution is the second one that @ocharles  mentioned which is just to preserve parsed whitespace in the syntax tree using the `Noted` constructors.

That still doesn't get us all the way, though, because intelligently preserving comments when formatting code is still difficult even when your AST preserves whitespace and comments.  Here are some pathological cases to consider when designing a comment-preserving formatting algorithm:

*   The formatter wants to format an expression as one line since it's less than 80 characters, but there is a line comment in the way:

    ```haskell
    { x = 1  -- !
    , y = 2
    }
    ```

*   Same as the previous example, but now there is a multi-line comment in the way:

    ```haskell
    { x = 1  {- !
                !
             -}
    , y = 2
    }
    ```

*   The formatter wants to reindent a value but doing so might break the indentation of a multi-line comment:

    ```haskell
    [ 1
    , 2
    ...
    ,     99 {- !
                !
             -}
    ]
[...]

PR #146, opened 2017-09-29
Preserve leading comments/whitespace when formatting code

Gabriel439:
Partial fix for #145

Before this change `dhall-format` would get rid of all comments when formatting
code.  After this change `dhall-format` will preserve any leading comments and
whitespace so that users can at least add top-level comment headers to their
files

cc: @psibi

### DIFF
--- a/src/Dhall/Parser.hs
+++ b/src/Dhall/Parser.hs
@@ -8,6 +8,7 @@
 module Dhall.Parser (
     -- * Utilities
       exprFromText
+    , exprAndHeaderFromText
 
     -- * Parsers
     , expr, exprA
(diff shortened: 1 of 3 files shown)


bosu:
For the record, this looks good to me. You are welcome to merge if @psibi is ok with it.

src/Dhall/Parser.hs
```diff
@@ -1010,16 +1014,39 @@ instance Exception ParseError
...
+
+    In other words, if you have a Dhall file of the form:
+
+> -- Comment 1
+> {- Comment -} 2
+
+    Then this will preserve @Comment 1@, but not @Comment 2@
+
+    This is used by @dhall-format@ to preserve leading comments and whitespace
+-}
```
psibi: 👍 Nice docs.

psibi:
Works as advertised. And thanks for the PR, this is much better than what we had before.

Gabriel439:
You're welcome!

PR #154, opened 2017-10-07
Implement standardized grammar

Gabriel439:
This updates the parser to match the newly standardized Dhall grammar located
at:

https://github.com/dhall-lang/dhall-lang/blob/master/standard/dhall.abnf

This is mostly the same as the old grammar, with a few important differences:

* Trailing commas and bars are *NO LONGER* supported for unions/records/lists
* Path imports no longer permit commas
* Quoted strings use JSON-style escape sequences
* The grammar for http/https is now standards-compliant
* Environment variables can now be quoted in order to support the full range
  of POSIX-compliant variable names

This change also includes a newly added test suite to exercise the parsing
code

This new implementation is also over twice as fast as the original implementation
on the example from https://github.com/dhall-lang/dhall-haskell/issues/108


bosu:
It is OK to merge by me. Obvious gotcha: dhall.cabal conflict :)

joelburget:
Were the parser changes done by hand?

Gabriel439:
@joelburget: Yes

PR #158, opened 2017-10-09
Remove `''${` escape sequence for double-quoted strings

Gabriel439:
The purpose of this change is to match this change to the grammar:

https://github.com/dhall-lang/dhall-lang/pull/46

I'm waiting a day before merging that change to the grammar and then after that it's
safe to merge this change


bosu:
+1 from me.

Issue #162, opened 2017-10-13
Discuss safety options for Dhall


Gabriel439:
The purpose of this issue is to discuss what the command-line API for the `dhall/dhall-*` executables should be when users want to disable certain Dhall features for additional safety

The most common requests I've gotten for how to improve safety are:

* Import-related safety:
    * **Option 1**: Disable all imports except `ipfs` imports
        * Disable all URLs except `https://ipfs.io/*`
        * Disable all paths except `/ipfs/*` (for use in conjunction with `ipfs mount`)
        * Disable all environment variable imports
        * This permits ease of distribution of Dhall libraries (like the Prelude)
        * This makes Dhall expressions mostly pure
            * `https://ipfs.io` still vulnerable to network-related interference (i.e. DNS changes)
            * Nothing enforces that `/ipfs` is used by `ipfs mount`
    * **Option 2**: Disable all remote imports
        * this prevents Dhall as a denial-of-service vector
        * this prevents Dhall from leaking information through `GET` requests
        * this prevents Dhall from leaking information through support for custom headers
    * **Option 3**: Disable all imports
        * this makes Dhall 100% pure
        * this prevents Dhall from being tricked into an infinite import loop by malicious hosts
            * i.e. `http://example.com/1 → http://example.com/2 → http://example.com/3 → ...`
* Normalization-related safety:
    * **Option 4**: Enforce that the input expression must already be in normal form
        * This gives a linear upper bound on how long it takes to interpret a configuration file
        * People who prefer this option either:
            * Don't want to use Dhall as a programming language at all, or:
            * Want to resolve/normalize Dhall expressions ahead-of-time

Mitigating factors:

* Dhall has one built-in safety feature, which is that remote imports cannot reference local imports (i.e paths and environment variables)
* The only IO that Dhall permits is imports

The Dhall API already supports all of the above features and they can all be implemented in linear time.  The main question is what *should* the command-line API support and what should the granularity of the API be

For example, one extreme is that the compiler just adds one new `--safe` flag which enforces the maximum possible safety by banning all imports and enforcing that the expression is in normal form.  This is simple but some users might want a solution with an intermediate amount of safety.

Another extreme is that we add four new flags (for each of the above options), which bloats the command-line API but satisfies everybody's use cases (that I'm aware of)

gfixler:
What about option arguments for imports?

    --imports=[ipfs|noremotes|none] --safe

Gabriel439:
[...]

PR #167, opened 2017-10-26
Change Dhall's integrity check to use an "intensional" hash

Gabriel439:
This updates the recently added support for integrity checking to hash the
final normalized result instead of hashing the text of the import.  This gives
much stronger guarantees about the safety of the imported value, since the
hash now reflects changes to the expression's semantic value

This also lets you safely refactor imported values without changing the hash so
long as the refactor is behavior-preserving.  The hash provides a quick and
dirty protection against human error when used in this way

### DIFF
--- a/Prelude/Bool/and
+++ b/Prelude/Bool/and
@@ -5,7 +5,7 @@ The `and` function returns `False` if there are any `False` elements in the
 Examples:
 
 ```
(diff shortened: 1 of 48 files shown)


Gabriel439:
This relates to #162

Gabriel439:
Merging as @bosu is currently unavailable

PR #169, opened 2017-10-27
Fix Travis badge in README

Widdershin:
Seems to be pointing at an old location of this repository, I have updated it to point at the current location.


Gabriel439:
Thank you for fixing that!

PR #186, opened 2017-12-04
Simple custom typechecking for Embed terms

ramirez7:
[This PR is best viewed with `?w=1`](https://github.com/dhall-lang/dhall-haskell/pull/186/files?w=1)

I've been exploring extending dhall via the `Embed` type constructor. Parsing and Normalization are easy to extend as-is, but type checking is not. This is because the current type checking module only works on `Expr s X`, but I need it to work with `Expr s a`.

This PR adds `typeWithA` and the `Typer` alias:
```haskell
type Typer a = forall s. a -> Expr s a
typeWithA :: Eq a => Typer a -> Context (Expr s a) -> Expr s a -> Either (TypeError s a) (Expr s a)
```
The caller can specify how to type their embedded terms pretty easily with this. The `Eq` constraint is needed in case the typeOf an embedded term is another embedded term.

`typeWith` still exists and is now written in terms of `typeWithA`:
```haskell
typeWith :: Context (Expr s X) -> Expr s X -> Either (TypeError s X) (Expr s X)
typeWith = typeWithA absurd
```

------

I've successfully used this branch to extend dhall with comparison operators for Doubles, Integers, and Naturals. There are end-to-end (i.e. parsing+typechecking+normalization) tests that illustrate the extension [here](https://github.com/ramirez7/dhall-extensions/blob/8e6d8a48272d081d603f920ad2aa04789520aea5/test/Dhall/Ord/Double/TheSpec.hs#L29). The corresponding `typer` can be found [here](https://github.com/ramirez7/dhall-extensions/blob/8e6d8a48272d081d603f920ad2aa04789520aea5/src/Dhall/Ord/Double/TypeCheck.hs#L11-L13). This PR is so that I can hopefully not use that branch and just use `dhall-haskell` proper :)

I'm currently working on exposing some `time` functionality in dhall as well. [Its `typer`](https://github.com/ramirez7/dhall-extensions/blob/8e6d8a48272d081d603f920ad2aa04789520aea5/src/Dhall/Time/TypeCheck.hs#L13-L23) is a little more interesting and makes use of that `Eq` constraint because it yields `Embed` for some terms.

Overall, I'm pretty excited to use dhall this way! Piggybacking on dhall's core gives me a lot for free as opposed to writing a DSL from scratch :)

### DIFF
--- a/src/Dhall/TypeCheck.hs
+++ b/src/Dhall/TypeCheck.hs
@@ -10,8 +10,10 @@ module Dhall.TypeCheck (
     -- * Type-checking
       typeWith
     , typeOf
+    , typeWithA
 
     -- * Types
+    , Typer
     , X(..)
     , TypeError(..)
     , DetailedTypeError(..)
@@ -40,7 +42,7 @@ import qualified Data.Vector
 import qualified Dhall.Context
 import qualified Dhall.Core
 
-axiom :: Const -> Either (TypeError s) Const
+axiom :: Const -> Either (TypeError s a) Const
 axiom Type = return Kind
 axiom Kind = Left (TypeError Dhall.Context.empty (Const Kind) Untyped)
 
@@ -61,7 +63,7 @@ match (V xL nL) (V xR nR) ((xL', xR'):xs) =
     nL' = if xL == xL' then nL - 1 else nL
     nR' = if xR == xR' then nR - 1 else nR
 
-propEqual :: Expr s X -> Expr t X -> Bool
+propEqual :: Eq a => Expr s a -> Expr t a -> Bool
 propEqual eL0 eR0 =
     State.evalState
         (go (Dhall.Core.normalize eL0) (Dhall.Core.normalize eR0))
@@ -112,6 +114,7 @@ propEqual eL0 eR0 =
             loop [] [] = return True
             loop _  _  = return False
         loop (Data.Map.toList ktsL0) (Data.Map.toList ktsR0)
+    go (Embed eL) (Embed eR) = return (eL == eR)
     go _ _ = return False
 
 {-| Type-check an expression and return the expression's type if type-checking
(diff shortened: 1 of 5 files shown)


src/Dhall/TypeCheck.hs
```diff
@@ -121,529 +124,535 @@ propEqual eL0 eR0 =
...
-            case s of
-                Const Type -> return ()
-                _ -> Left (TypeError ctx e (InvalidListType t))
-            flip Data.Vector.imapM_ xs (\i x -> do
-                t' <- typeWith ctx x
-                if propEqual t t'
+typeWith :: Context (Expr s X) -> Expr s X -> Either (TypeError s X) (Expr s X)
+typeWith = typeWithA absurd
+
+type Typer a = forall s. a -> Expr s a
```
Gabriel439: May I suggest:

```haskell
type Typer s a = a -> Either (TypeError s a) (Expr s a)
```

... that way type-checking an embedded expression can potentially fail

ramirez7: I'm a bit conflicted on this. On the one hand, this type signature is a bit more general and powerful. But on the other hand, I kind of like that the current type signature ensures that every possible embedded term is well-typed. Applications of embedded terms can still be poorly-typed though.

In practice, values of `a` come from..
1. Parsing via a custom `Parser a`. In this case, we should probably fail to parse instead of generate a poorly-typed term. I think any type error of a freshly-parsed `a` should probably just be a parse error.
2. Normalization of applications via a custom `Normalizer a`. In this case, I think normalization should always result in a well-typed value.

I'm having trouble thinking of a use-case that necessitates this more general type. Any ideas?

Gabriel439: Thinking about this more, I agree with you.  If we really need support for it we can add it later once we have a motivating use case

PR #188, opened 2017-12-09
set hPrint encodings to use utf8, independent of locale

Profpatsch:
On systems with the locale set to non-UTF8 (e.g. `LC_ALL=C`, the default POSIX
locale), handles are opened in that locale, so printing dhall’s error messages
fails on the first non-ASCII character:

```
Error: Not a function

Explanation: Expressions separated by whitespace denote function application,
like this:

    dhall: <stderr>: hPutChar: invalid argument (invalid character)
```

We set the stderr handle accordingly.

For example, in nix builds, dhall outputs:

```
these derivations will be built:
  /nix/store/im88hab49pp2rajjzx50fyvig7i7cbpr-dhall.drv
building path(s) ‘/nix/store/0khp05jvcgim2q48k3jm016acx0545qa-dhall’


Error: Not a function

Explanation: Expressions separated by whitespace denote function application,   
like this:                                                                      
                                                                                
                                                                                
    dhall: <stderr>: hPutChar: invalid argument (invalid character)
builder for ‘/nix/store/im88hab49pp2rajjzx50fyvig7i7cbpr-dhall.drv’ failed with exit code 1
error: build of ‘/nix/store/im88hab49pp2rajjzx50fyvig7i7cbpr-dhall.drv’ failed
```

With this change, the build runs as expected and the correct glyphs are printed.

### DIFF
--- a/dhall-format/Main.hs
+++ b/dhall-format/Main.hs
@@ -71,6 +71,7 @@ main = do
 
     let handler e = do
             let _ = e :: SomeException
+            System.IO.hSetEncoding System.IO.stderr System.IO.utf8
             System.IO.hPrint stderr e
             System.Exit.exitFailure
 
--- a/dhall-format/Main.hs
+++ b/dhall-format/Main.hs
@@ -90,6 +90,7 @@ main = do
                     Pretty.renderIO handle (Pretty.layoutSmart opts doc)
                     Data.Text.IO.hPutStrLn handle "" )
             Nothing -> do
+                System.IO.hSetEncoding System.IO.stdin System.IO.utf8
                 inText <- Data.Text.Lazy.IO.getContents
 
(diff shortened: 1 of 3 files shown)


Profpatsch:
Just noticed the input stream has to be explicitely set to UTF-8 as well. Preparing commit …

Gabriel439:
Thanks for fixing that!  I'll merge as soon as CI finishes

Profpatsch:
Maybe it’s a nice idea to move the duplicated code in the various CLI utils into a shared intermediate module? `Dhall.CLI` or similar.

Gabriel439:
@Profpatsch: Yeah, I think it's good to push convenient shared utilities into the main library so that people can easily write their own forks of the various `dhall-*` executables

Issue #189, opened 2017-12-11
Annotate MissingFile exception with absolute path


Profpatsch:
```
↳ ../imports/Prelude; 

Error: Missing file
```

I suggest adding the absolute filepath it tried to stat, this should help debugging complex setups.
Not sure if that were a security risk in some cases, though.

Gabriel439:
There's no security risk if you return the canonicalized path which Dhall already stores internally.  In other words if you had `../foo/bar → ./baz` then the canonicalized path would be `../foo/baz`.  Would that be alright?

Gabriel439:
I'll mark this closed since the corresponding pull request is merged

PR #190, opened 2017-12-11
annotate MissingFile errors

Profpatsch:
Should make it easier to debug complex dhall configurations.

Implements https://github.com/dhall-lang/dhall-haskell/issues/189#issuecomment-350789514

### DIFF
--- a/src/Dhall/Import.hs
+++ b/src/Dhall/Import.hs
@@ -164,6 +164,7 @@ import qualified Data.ByteString.Lazy
 import qualified Data.CaseInsensitive
 import qualified Data.List                        as List
 import qualified Data.Map.Strict                  as Map
+import qualified Data.Text
 import qualified Data.Text.Encoding
 import qualified Data.Text.IO
 import qualified Data.Text.Lazy                   as Text
@@ -301,15 +302,22 @@ instance Show PrettyHttpException where
 #endif
 
 -- | Exception thrown when an imported file is missing
-data MissingFile = MissingFile
+data MissingFile = MissingFile FilePath
     deriving (Typeable)
 
 instance Exception MissingFile
 
 instance Show MissingFile where
-    show MissingFile =
+    show (MissingFile path) =
             "\n"
-        <>  "\ESC[1;31mError\ESC[0m: Missing file\n"
+        <>  "\ESC[1;31mError\ESC[0m: Missing file "
+        <>  Data.Text.unpack formattedPath
+        <>  "\n"
+      where
+        formattedPath = case Filesystem.Path.CurrentOS.toText path of
+            (Right t) -> t
+            (Left  t) -> t
+                <> "\n\ESC[1;31mWarning\ESC[0m: Filename contains non-displayable characters"
 
 -- | Exception thrown when an environment variable is missing
 newtype MissingEnvironmentVariable = MissingEnvironmentVariable { name :: Text }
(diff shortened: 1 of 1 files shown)


src/Dhall/Import.hs
```diff
@@ -589,7 +598,7 @@ exprFromPath m (Path {..}) = case pathType of
                 exists <- Filesystem.isFile path
                 if exists
                     then return ()
-                    else Control.Exception.throwIO MissingFile
+                    else Control.Exception.throwIO (MissingFile path)
```
Profpatsch: Is `path` already canonicalized here?

src/Dhall/Import.hs
```diff
@@ -134,6 +134,7 @@ import Data.Text.Lazy.Builder (Builder)
 import Data.Traversable (traverse)
 #endif
 import Data.Typeable (Typeable)
+import qualified Filesystem.Path.CurrentOS
```
Gabriel439: For alignment reasons I try to keep all the qualified imports together at the bottom

src/Dhall/Import.hs
```diff
@@ -176,11 +178,11 @@ import qualified Dhall.Parser
 import qualified Dhall.Context
 import qualified Dhall.TypeCheck
 import qualified Filesystem
-import qualified Filesystem.Path.CurrentOS
 import qualified Network.HTTP.Client              as HTTP
 import qualified Network.HTTP.Client.TLS          as HTTP
 import qualified Filesystem.Path.CurrentOS        as Filesystem
 import qualified System.Environment
+import qualified System.Directory
```
Gabriel439: What's the reason for the `System.Directory` import?  It seems like it's not used

src/Dhall/Import.hs
```diff
@@ -301,15 +303,22 @@ instance Show PrettyHttpException where
...
             "\n"
-        <>  "\ESC[1;31mError\ESC[0m: Missing file\n"
+        <>  "\ESC[1;31mError\ESC[0m: Missing file "
+        <>  Data.Text.unpack formattedPath
+        <>  "\n"
+      where
+        formattedPath = case Filesystem.Path.CurrentOS.toText path of
+            (Right t) -> t
+            (Left  t) -> t
+                <> "\n(filename contains non-displayable characters)"
```
Gabriel439: Minor nitpick: Perhaps use: `\ESC[1;31mWarning\ESC[0m: Filename contains ...`

dhall.cabal
```diff
@@ -100,6 +100,7 @@ Library
         containers           >= 0.5.0.0  && < 0.6 ,
         contravariant                       < 1.5 ,
         cryptohash                          < 0.12,
+        directory            >= 1.2.2.0  && < 1.4 ,
```
Gabriel439: This import looks like it is unused

src/Dhall/Import.hs
Profpatsch: Oh right, I played around with `makeAbsolute` and forgot to remove it once I saw that it leads nowhere.

src/Dhall/Import.hs
Gabriel439: Yes.  The part where `exprFromPath` gets called (very indirectly) is here: https://github.com/Profpatsch/dhall-haskell/blob/8b741bae1fba6c7a007949db7f5e8058b5adabc7/src/Dhall/Import.hs#L752

src/Dhall/Import.hs
Profpatsch: Fixed in patch.

Profpatsch:
Should have addressed your comments.

Gabriel439:
Awesome!  I'll merge once CI is green.  Thanks for doing this :)

PR #205, opened 2018-01-18
Adds support for `as Template`.

angerman:
This would resolve the issue I ran into in dhall-lang/dhall-lang#67.

This will basically allow to inject open templates into dhall expressions,
and decouple dhall logic from the pure templating logic for interpolation.

Example:
```bash
cat << EOF > index.html                                                                                                                                                                   ~/Projects/zw3rk/dhall-haskell
<html>
  <head><title>\${page.title}</title></head>
  <body><h1>\${page.header}</h1>
        <p>dhall is \${page.value}!</p>
  </body>
</html>
EOF
```
with
```bash
$ dhall <<< 'let page = { title = "dhall templates"
                        , header = "Templates"
                        , value = "a great templating language" }
             in (./index.html as Template)' 
```

will result in
```
<html>
  <head><title>dhall templates</title></head>
  <body><h1>Templates</h1>
        <p>dhall is a great templating language!</p>
  </body>
</html>
```

### DIFF
--- a/src/Dhall/Core.hs
+++ b/src/Dhall/Core.hs
@@ -135,7 +135,7 @@ instance Buildable PathType where
     build (Env env) = "env:" <> build env
 
 -- | How to interpret the path's contents (i.e. as Dhall code or raw text)
-data PathMode = Code | RawText deriving (Eq, Ord, Show)
+data PathMode = Code | RawText | Template deriving (Eq, Ord, Show)
 
 data PathHashed = PathHashed
     { hash     :: Maybe Data.ByteString.ByteString
@@ -161,6 +161,7 @@ instance Buildable Path where
       where
         suffix = case pathMode of
             RawText -> "as Text"
+            Template -> "as Template"
             Code    -> ""
 
 instance Pretty Path where
@@ -2472,6 +2473,7 @@ reservedIdentifiers =
         , "Double"
         , "Double/show"
         , "Text"
+        , "Template"
         , "List"
         , "List/build"
         , "List/fold"
--- a/src/Dhall/Import.hs
+++ b/src/Dhall/Import.hs
@@ -97,6 +97,15 @@
     > ocuments. You may use this\n    domain in examples without prior coordination or
     >  asking for permission.</p>\n    <p><a href=\"http://www.iana.org/domains/exampl
     > e\">More information...</a></p>\n</div>\n</body>\n</html>\n"
+
+    If you wish to just interpolate the contents of a path with the current scope as
+    @Text@ then add @as Template@ to the end of the import:
+
+    > $ echo 'value of var is \${var}'
+    > $ dhall <<< 'let var = "val" in ./template as Template'
+    > Text
+    > 
+    > "value of var is val\n"
 -}
 
 module Dhall.Import (
@@ -143,7 +152,7 @@ import Dhall.Core
     , PathType(..)
     , Path(..)
     )
-import Dhall.Parser (Parser(..), ParseError(..), Src(..))
+import Dhall.Parser (Parser(..), ParseError(..), Src(..), exprFromText)
 import Dhall.TypeCheck (X(..))
 #if MIN_VERSION_http_client(0,5,0)
 import Network.HTTP.Client
@@ -621,7 +630,23 @@ exprFromPath (Path {..}) = case pathType of
             RawText -> do
                 let pathString = Filesystem.Path.CurrentOS.encodeString path
                 text <- Data.Text.IO.readFile pathString
-                return (TextLit (build text)) )
+                return (TextLit (build text))
+            Template -> do
+                let pathString = Filesystem.Path.CurrentOS.encodeString path
+                template <- Data.Text.IO.readFile pathString
+
+                let textPath = case Filesystem.Path.CurrentOS.toText path of
+                      Left  text -> text
+                      Right text -> text
+                    bytesPath = Data.Text.Encoding.encodeUtf8 textPath
+                let delta =
+                        Directed bytesPath 0 0 0 0
+                case exprFromText delta (Text.fromStrict $ "''" <> template <> "''") of
+                    Left parserError -> do
+                        throwIO parserError
+                    Right expr -> do
+                        return expr )
+
     URL url headerPath -> do
         m       <- needManager
         request <- liftIO (HTTP.parseUrlThrow (Text.unpack url))
@@ -715,6 +740,17 @@ exprFromPath (Path {..}) = case pathType of
                     Success expr -> return expr
             RawText -> do
                 return (TextLit (build text))
+            Template -> do
+                let urlBytes = Data.Text.Lazy.Encoding.encodeUtf8 url
+                let delta =
+                        Directed (Data.ByteString.Lazy.toStrict urlBytes) 0 0 0 0
+                case exprFromText delta ("''" <> text <> "''") of
+                    Left parserError -> do
+                        liftIO (throwIO parserError)
+                    Right expr -> do
+                        return expr
+
+
     Env env -> liftIO (do
         x <- System.Environment.lookupEnv (Text.unpack env)
         case x of
@@ -730,6 +766,16 @@ exprFromPath (Path {..}) = case pathType of
                             Success expr    -> do
                                 return expr
                     RawText -> return (TextLit (build str))
+                    Template -> do
+                        let envBytes = Data.Text.Lazy.Encoding.encodeUtf8 env
+                        let delta =
+                                Directed (Data.ByteString.Lazy.toStrict envBytes) 0 0 0 0
+                        case exprFromText delta (Text.pack $ "''" <> str <> "''") of
+                            Left parserError -> do
+                                throwIO parserError
+                            Right expr -> do
+                                return expr
+
             Nothing  -> throwIO (MissingEnvironmentVariable env) )
   where
     PathHashed {..} = pathHashed
@@ -825,7 +871,9 @@ loadStaticWith from_path ctx path = do
     --
     -- There is no need to check expressions that have been cached, since they
     -- have already been checked
-    if cached
+    --
+    -- Also do not type check templates on their own; they are not closed.
+    if cached || (pathMode path == Template)
         then return ()
         else case Dhall.TypeCheck.typeWith ctx expr of
             Left  err -> throwM (Imported (path:paths) err)
--- a/src/Dhall/Parser.hs
+++ b/src/Dhall/Parser.hs
@@ -586,6 +586,9 @@ _Double = reserved "Double"
 _Text :: Parser ()
 _Text = reserved "Text"
 
+_Template :: Parser ()
+_Template = reserved "Template"
+
 _List :: Parser ()
 _List = reserved "List"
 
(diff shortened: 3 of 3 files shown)


angerman:
I'm not very happy with this rather heavy approach to be honest (implementing it in dhall-haskell, and not in dhall-text). I found it to feel very natural though and wanted to at least provide it as a basis for discussion.

Gabriel439:
I'm probably not going to add this to the language and I would prefer to still treat templates as imported functions that interpolate text.  The reasons why are:

*   I would like users to be able to type-check and infer the type of imported files in isolation
*   I would like anybody reading an imported template to be able to to infer from the file's type what inputs the file expects
*   I would like to avoid features that overlap with existing features to keep the language features as orthogonal as possible
    * We can already import a function, and this overlaps with that
*   Templating `Text` is not the only use case for Dhall so I don't want to optimize the language for a single use case
    * In fact, I view `Text` output as a last resort.  For example, it would be better for your use case if there were a `dhall-to-html` executable that expected a strongly typed data structure as the final output instead of `Text`
*   This doesn't save that much code for the imported template.  In the worst case, you just add one line for the header and footer:

    ```haskell
    λ(args : ./Input) → ''
    ...
    ''
    ```

    ... where `args` is a record containing all the fields you want to import

angerman:
As I mentioned, I didn't have high hopes of getting this into dhall. It just ended up being the most natural way to add this feature to dhall for me. I fully understand your reasons.

For me
```
\(args : ./Input) -> ''
...
''
```
is just not an acceptable substitute as it hijacks what ever the template is into a dhall file, and as such breaks tooling around the template. For the HTML example, most editors are perfectly fine with `.html` files having placeholders in the text or attribute parts; and I can even render that html file in my browser and it will show the placeholder just fine. But adding the necessary dhall boilerplate around it, that just stops working.

Regarding the type checking of `./t as Template`, it is always going to be `Text` or am I mistaken? It's just that you get interpolation from the current scope.

I guess we come from two different view points. Where you want this to be a clean as possible and everything to be pure dhall files and I'd like to allow wrapping impure files into dhall files, while retaining the original files syntax, tooling.

I could turn my foreign template into a dhall file via

```
\(args : ./Input) -> ./template.foreign as Template
```

at which point I can retain my original `template.foreign` with its syntax and tooling and use it transparently in dhall via the wrapper.

Feel free to close this PR, and dhall-lang/dhall-lang#67 as won't fix.

Gabriel439:
What I mean by type-checking is that a user should be able to do this:

```
$ dhall <<< './t'
```

... in order to validate that the template is well-formed and to discover what value need to be in scope for that template.  I don't want users to have to visually scan a template or use trial and error to discover what free variables need to be in scope for a template.

More generally, I don't want to make significant changes to the language semantics just to work around limitations of existing tooling or editor-specific concerns, especially if they are limited to just one weakly typed application domain of Dhall.

From my point of view, an editor should not highlight the body of a Dhall `Text` literal because Dhall treats `Text` as an unstructured black box and the editor should, too.  You should focus on using an algebraic data type to encode HTML instead of tricking your editor to treat `Text` as a structured value.  If you add [editor support for Dhall](https://github.com/psibi/dhall-mode) then treating HTML as an algebraic data type would give you the syntax highlighting you desire. 

What I propose is closing this in favor of a creating new issue against the https://github.com/dhall-lang/dhall-lang repository to create a strongly typed Dhall to HTML binding (i.e. a `dhall-to-html`).  Would that serve your use case?

angerman:
With `t` being `\(args : ./Input) -> ./template.foreign as Template` that would work as you expect I believe.  I never really intended the template to actually be a proper dhall expression. It's effectively just an interpolated `Text`, that I can externalize.

I fully agree that the editor should not highlight Dhall `Text` values. But it should do so for a `.html` file. However the HTML example has likely been a bad one, as there are proper representations for HTML that are not just a stream of ascii characters.

Maybe an example like a markdown or org mode file or some other programming language for which I want to interpolate a template via dhall.

I would want to retain the template in the file format it is in and to be able to keep the tooling around it. It is after all not really a dhall file, it just a file that contains some `${...}` placeholders.

So, no I don't think that that would solve my use case(s), it might solve the HTML templating one, but I would end up writing `dhall-to-x` and build some adt for the templates target language.

I see that I might try to use dhall for something it's just not designed for.

Gabriel439:
I also forgot that there are a two reasons for why Dhall requires that imports are closed:

First, imports are type-checked before embedded into the current syntax tree (in order to improve error localization).  If an imported file has free variables then you can't localize any type errors to that imported file because it cannot be type-checked in isolation.

Second, the Dhall API permits normalizing a file without resolving imports, but that breaks if imported files have free variables.  As a very simple example, suppose you had:

```
$ cat ./foo
(λ(x : Text) → ./bar as Template) "John"

$ cat ./bar
Hello ${x}
```

Then you wouldn't be able to correctly normalize `./foo`.  However, normalization is entirely safe if the imported expression is closed.

However, I still think my overarching concern is that this seems like overfitting the language standard to the use case of templating text, which is not the primary integration for Dhall since `Text` the most weakly typed Dhall integration of them all  This is why I prefer the original proposal of adding a utility to the `dhall-to-text` library than changing the language standard.

andrewthad:
Interestingly, I wrote [something vaguely similar](https://github.com/andrewthad/dhall-html/blob/d3b9a4702549e0899f48e7c2c4d9e8dedee1fd95/src/Dhall/Html/Import.hs#L540-L551) while trying marshall HTML to dhall. What I wrote is different in that the template still has to list out all of it's lambda-bound variables. While I agree that adding something like this to the language itself would be inappropriate, I feel like it would be nice if the haskell dhall library offered a way to make this more extensible (so that I wouldn't have to copy a whole module to introduce this new behavior). I'm working on a markdown integration right now (that will convert markdown to dhall), and I'm going to run into this same problem again. I don't have a particular solution in mind, but I just wanted get another use-case for this kind of thing out there.

Gabriel439:
I'll go ahead and close this since it's unlikely to be added to the language standard

Nadrieril:
I'd like to reopen discussion on this, because I think it would be an awesome feature for dhall to have:

- Programmable templates tend to be difficult to use or not flexible enough outside of language-specific bindings. I feel like this would be an awesome niche for dhall to fill;
- If we look at Nix, it seems that this would be a really useful feature. Right now quite a few files in nixpkgs use ad-hoc `@@VARIABLE@@` interpolation that needs to be done at build-time, even though Nix itself has nice string interpolation features. This shows that even though they could start the file with `x: ''` and end it with `''` like you propose, this was deemed inelegant or confusing;
- I think a viable alternative that solves the typing concerns would be a special shebang-like header like `#!dhall-template \(x: ./Type) -> _`. This is significantly different from `\(x: Type) -> ''` because it makes the intent clear that the templated file is primarly in whatever target format we want, and not primarly in dhall. In particular, it wouldn't break syntax highlighting;
- If this is too weird, at least specifying the type when importing should be required: `./file as ({ var1: X, var2: Y } -> Text)`;
- I don't feel that it breaks orthogonality; it rather feels like a natural extension of the current capabilities of importing a file and string interpolation;
- I completely agree that Text should be avoided when more structured types would make sense, but it's a huge boost in usability to be able to convert said structured type into whatever format is actually required directly in Dhall, without needing to write language bindings. It's so much duplication and added complexity when everything is already available in dhall. It also considerably eases transition towards using these more structured types, since you would be able to replace existing files bit-by-bit.

Gabriel439:
@Nadrieril: I still don't think this should be added to the language because it introduces more complexity than it eliminates:

* We'll now be committing to supporting language features that place nice with syntax highlighters for target file formats that users template

  For example, suppose that the user wants to target HTML where a `#!dhall-template` header doesn't work.  Should we standardize support for the parent import customizing the expected header line?

* If we decide to implicitly thread variables in scope, we'll be asked to do the same for non-template imports as well

* We'll need to add different `dhall` tooling for validating templates like these

  Ideally, there is a uniform way to validate everything in the Dhall ecosystem, which is the vanilla Dhall type-checker

The second issue I have is that this encourages stringly-typed templating, which can lead to the generated file being malformed.  Many people arrive at Dhall because they are fleeing that exact anti-pattern (typically templated YAML files leading to subtle bugs).  We should be promoting people to use Dhall to generate typed abstract syntax trees that a dedicated tool (such as `dhall-to-yaml` or `dhall-nethack`) then renders, rather than encouraging end users to directly template strings that are syntactically significant.  More generally, we should design the language and ecosystem so that best practices are path of least resistance for end users.

Nadrieril:
This adds little new complexity: apart from header parsing, the rest of the template would behave exactly like a multiline string. The header would be syntactic sugar for a function type, but apart from that the file would behave as any other dhall expression.
In particular, implicit scope threading is a complete no-no.

It does encourage stringly-typed templating a bit, but I think it's worth it:
- We would have a much better story for smoothly transitioning existing configuration into dhall.
- We would make it very easy to work with any format, not matter how wild, at least on a best-effort basis, without needing to wait for the implementation of a general-purpose `dhall-to-blah` tool.
- I don't think that this would cause too many problems, since using the tool would always be more convenient than writing dhall folds manually. Manual file templating would be reserved for cases where a format-specific tool doesn't make sense or isn't needed yet.

As for needing to support more things, I'd probably agree about supporting the 3-4 main comment syntaxes but that's it. Since the header would be removed in the output it doesn't really matter. Syntax highlighting would not be the aim, only a nice side-effect of the format, and existing templating languages don't seem to care too much. The main thing is not having unmatched delimiters or unfinished string literals.
We would probably get requests for various builtins to escape strings however.

Gabriel439:
I still feel like the benefit of this (syntax highlighting for templated files) is not enough benefit for the extra standard complexity and design space that this opens up.

Also, based on survey feedback, the initial use of the language does not appear to be the bottleneck in user adoption.  Most people report getting stuck on larger design patterns and project organization rather than IDE support and syntax.

I also want to note is that this doesn't benefit the YAML-replacement use case, which is the use case that we want to polish to death for the language to get a mainstream foothold before addressing other use cases.  I would like to reserve our complexity budget for features that enhance that initial mainstream use case, before targeting other formats.

f-f:
I agree that this is not worth the tradeoff in complexity. 
In particular, the use-case of "keeping syntax highlighting" can be solved with one line of bash that properly wraps the template in the original format:
```
printf "\(x : { ... }) -> ''\n$(cat mytemplate.html)\n''"
```

Nadrieril:
Yeah you're both right, I realize the benefit is actually quite tiny now. Sorry I got so intense over this.
Especially if our focus is exclusively displacing YAML. It feels good to have a project with a clear goal and leadership :). Thanks for taking the time to answer me, as always !

Gabriel439:
@Nadrieril: You're welcome! 🙂

antislava:
@Nadrieril, @f-f 
I find this tiny utility pretty useful [zimbatm/shab: the bash templating language](https://github.com/zimbatm/shab/),  at least in simple situations. For example,

```sh
cat <<EOF > test.md
## Intro
Hello \${name}!
EOF

cat <<EOF > test.dhall
let name = "world"
in ''
$(cat ./text.md)
'' : Text
EOF

shab ./text.dhall | dhall-to-text
```

Resulting in

```markdown
## Intro
Hello world
```

`shab` is in nix packages. (Otherwise it is just a very small but clever bash script.)

Profpatsch:
`test.dhall` is not really a dhall file anymore, so you lose minor things like referential transparency ;) Here be dragons.

antislava:
@Profpatsch I agree, the `.dhall` extension can be misleading here! ;)

PR #212, opened 2018-01-19
Generalize the X argument to Expr in Type and InputType

### LINKED ISSUE #210: Could Dhall.Type have X as a type parameter?
I may be using Dhall incorrectly, but here's what I'm doing.

`dhall-to-cabal` essentially works by taking a Dhall expression, type-checking it and then walking the normalized expression to generate a Cabal `PackageDescription`. I'm mostly doing this by using lots of little `Type`s and then composing them, to use `extract` and `expected`. (Let's ignore Generics for now.)

This works well, until I start introducing some global functions. One such function I'm introducing is:

```
majorVersionBound : List Natural -> VersionRange
```

My plan was to use `normalizeWith` to turn `App "majorVersionBound" components` into `Embed VersionRange`. While this is fine, I then can no longer `extract` from the result, because `extract` fixes me to `Embed X`. It seems that this could really be a type parameter, and most combinators could universally quantify over this.

What do you think? Thanks for being so responsive through my barrage of questions today.

ocharles:
Fixes #210

### DIFF
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -128,7 +128,7 @@ instance Exception InvalidType
 True
 -}
 input
-    :: Type a
+    :: Type X a
     -- ^ The type of value to decode from Dhall to Haskell
     -> Text
     -- ^ The Dhall program
@@ -164,9 +164,9 @@ input (Type {..}) txt = do
 --   a much better user experience.
 rawInput
     :: Alternative f
-    => Type a
+    => Type x a
     -- ^ The type of value to decode from Dhall to Haskell
-    -> Expr s X
+    -> Expr s x
     -- ^ a closed form Dhall program, which evaluates to the expected type
     -> f a
     -- ^ The decoded value in Haskell
@@ -292,27 +292,31 @@ detailed =
     handler1 :: TypeError Src X -> IO a
     handler1 e = Control.Exception.throwIO (DetailedTypeError e)
 
-{-| A @(Type a)@ represents a way to marshal a value of type @\'a\'@ from Dhall
-    into Haskell
+{-| A @(Type x a)@ represents a way to marshal a value of type @\'a\'@ from
+    Dhall into Haskell
 
     You can produce `Type`s either explicitly:
 
-> example :: Type (Vector Text)
+> example :: Type x (Vector Text)
 > example = vector text
 
     ... or implicitly using `auto`:
 
-> example :: Type (Vector Text)
+> example :: Type x (Vector Text)
 > example = auto
 
     You can consume `Type`s using the `input` function:
 
-> input :: Type a -> Text -> IO a
+> input :: Type X a -> Text -> IO a
+
+    The @x@ parameter allows you to `Embed` domain-specific data into your
+    marshalling. This is often just `X`, indicating that there is no extra
+    information.
 -}
-data Type a = Type
-    { extract  :: Expr Src X -> Maybe a
+data Type x a = Type
+    { extract  :: Expr Src x -> Maybe a
     -- ^ Extracts Haskell value from the Dhall expression
-    , expected :: Expr Src X
+    , expected :: Expr Src x
     -- ^ Dhall type of the Haskell value
     }
     deriving (Functor)
@@ -322,7 +326,7 @@ data Type a = Type
 >>> input bool "True"
 True
 -}
-bool :: Type Bool
+bool :: Type x Bool
 bool = Type {..}
   where
     extract (BoolLit b) = pure b
@@ -335,7 +339,7 @@ bool = Type {..}
 >>> input natural "+42"
 42
 -}
-natural :: Type Natural
+natural :: Type x Natural
 natural = Type {..}
   where
     extract (NaturalLit n) = pure n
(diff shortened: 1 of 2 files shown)


src/Dhall.hs
```diff
@@ -292,27 +292,31 @@ detailed =
...
+    marshalling. This is often just `X`, indicating that there is no extra
+    information.
 -}
-data Type a = Type
-    { extract  :: Expr Src X -> Maybe a
+data Type x a = Type
+    { extract  :: Expr Src x -> Maybe a
     -- ^ Extracts Haskell value from the Dhall expression
-    , expected :: Expr Src X
+    , expected :: Expr Src x
```
ocharles: I am not sure about this change.

Gabriel439:
There's one other approach that I haven't tried yet: what if we changed the type of `extract` to:

```
extract :: Expr Src () -> Maybe a
```

Then you could convert to that `Expr` type by just doing `() <$`

Also, I'm pretty sure that `expected` should be `Expr Src X` because then you can do `fmap absurd` to get any other type you need.

ocharles:
While possible, it doesn't give you the ability to have something like `foo :: Type MyVerySpecificLibrary T` where you actually want to use `Embed` values in `extract`. Is such a case likely?

Gabriel439:
I think we should wait until we have a concrete use for the extra type parameter but if we do then I'm happy to add it

ocharles:
I'm stuck on that change then, as `Dhall.input` calls `normalize` which expects `Expr X` not `Expr ()`.

Gabriel439:
Oh, that is a more interesting type error

`typeOf` was specifically designed to reject programs with unresolved imports because it assumes that `Embed` is supposed to represent an import

Could you go back to why you convert `App "majorVersionBound" components` to `Embed VersionRange`?

Gabriel439:
Or let me ask a slightly different question: why do you not convert it to `Var "VersionRange"`?

ocharles:
Probably because I don't really know what I'm actually doing :smile: I kind of wanted to make `VersionRange` opaque, but now I'm thinking it might be better to just expose the sum type https://hackage.haskell.org/package/Cabal-2.0.1.1/docs/Distribution-Version.html#t:VersionRange directly as a `Union` and have `majorVersion` simply be a Dhall function that someone would import.

ocharles:
I think https://github.com/ocharles/dhall-to-cabal/blob/master/lib/Distribution/Package/Dhall.hs#L788 could just go directly in an `extract` definition, rather than normalizing to `Embed`.

Gabriel439:
You can still make `VersionRange` opaque, but you don't need to use `Embed` to do so.  Just make it a `Var` (i.e. a free variable) and that will play nicely with all the other customizable functions

Gabriel439:
Additionally, the extract definition shouldn't have to perform any normalization.  Add the logic to evaluate `majorVersions` to `normalizeWith` and then after the normalization pass do the extraction

Gabriel439:
Note that I haven't had a chance to do a deeper architectural review of `dhall-to-cabal` but soon I will have some time to give it a closer look

ocharles:
That'd be terrific. I'm going to see if I can remove the need to `Embed` now.

ocharles:
Ok, I think I can provide a bit more info as to my reasoning.

I wanted `VersionRange` to be opaque, so I extend the typing context with `VersionRange` and some smart constructors to construct them from smaller parts. This mirrors what I see in https://hackage.haskell.org/package/Cabal-2.0.1.1/docs/Distribution-Version.html. At some point, I need to turn `majorVersion { .. }` into a call to `Cabal.majorVersion`. Doing this in `normalize` felt like it made the most sense, but I have to pass around a `Cabal.VersionRange`, and `Embed` seemed to be the best option.

It seems now that you're more suggesting that I should normalize `majorVersion` into some Dhall `Expr` (e.g., `UnionLit`) and then later extract that `UnionLit` into a `Cabal.VersionRange`. The other alternative seems to be not doing custom normalisation at all, and simply having `extract` expect `App (Var "majorVersion") (RecordLit args)` but that does not feel idiomatic.

Gabriel439:
Yeah, that's correct.  You should defer using any Haskell types in the pipeline until the final extract `step`.  That means that the output of `normalize` should still be the Dhall analog of the Haskell type that you eventually want to `extract` to

ocharles:
Ok, well if I want to keep it entirely opaque then I think the "correct" solution here is to `extract` the actual function application. It does seem a tad odd though. It's that, or normalize would actually break typing. For example, if we have `{ version : VersionRange }`, `{ version = majorVersion [ +1, +0] }` might be well typed, but I would normalize to `{ version = < MajorVersion = [ +1, +0 ] > }` which wouldn't type check.

Does my confusion make sense there?

Gabriel439:
Oh yeah, you're right.  Then I agree pushing the logic to detect the fully saturated `majorVersion` function call into `extract` would be the correct thing to do here if you want the `cabal` schema to reference an opaque type.

However, you should also consider not using an opaque `VersionRange` type, too

ocharles:
Alright. For now I'll go with the former, and I'll make an issue to consider just removing `VersionRange` (or at least not making it a primitive). I think we can reject this PR for now, let me just make the changes and make sure I don't need this work.

Gabriel439:
Alright, sounds good.  I'll go ahead and close this, then

PR #220, opened 2018-01-21
Preserve string interpolation when using `dhall-format`

### LINKED ISSUE #216: dhall-format removes text interpolation
Given

```dhall
let GitHubProject : Type = { owner : Text, repo : Text } in
let gitHubProject = \( github : GitHubProject ) ->
     let gitHubRoot = "https://github.com/${github.owner}/${github.repo}"
	 in ./empty-package.dhall // { bugReports = "${gitHubRoot}/issues"  }
in gitHubProject
```

`dhall-format` produces

```dhall
    let GitHubProject : Type = { owner : Text, repo : Text }

in  let gitHubProject =
            λ(github : GitHubProject)
          →     let gitHubRoot =
                          "https://github.com/"
                      ++  github.owner
                      ++  "/"
                      ++  github.repo
                      ++  ""
            
            in    ./empty-package.dhall 
                ⫽ { bugReports = gitHubRoot ++ "/issues" }

in  gitHubProject
```

I would rather it kept it as text interpolation.

Gabriel439:
Fixes #216

This changes the internal representation of `TextLit` to be able to
store an interpolated string literal.  Preserving string interpolation in
the syntax tree means that `dhall-format` can also preserve
interpolation.

This also simplifies a lot of the string-interpolation-related code,
which is now simpler and more strongly typed.


ocharles:
Looks good! Maybe worth throwing in my example in #216 as a regression test?

ocharles:
Oh btw, there is still a comment at the top of `dhall-format`'s main stating that it doesn't preserve text interpolation - but now it does! :champagne:

Gabriel439:
Oh, good catch!  I'll fix that :)

PR #222, opened 2018-01-21
Add 'inputWith'

### LINKED ISSUE #213: A more advanced "go" function
Right now we have `input` which is fairly useful, but the moment you have to supply custom typing contexts and normalizers, the difficulty significantly rises. What do you think about an "advanced" convenience function that:

* Takes a typing context, and uses it while resolving imports
* Takes a `Normalizer` to use `normalizeWith`

I feel like this when paired with `input` is a good 90% of use cases.

ocharles:
'inputWith' is like 'input', but allows for a custom typing context and
normalizer.

Fixes #213.


ocharles:
Could do with a hand documenting this one.

Gabriel439:
Thanks!  This looks great to me

PR #233, opened 2018-01-27
Allow period in quoted labels

Gabriel439:
... as standardized in https://github.com/dhall-lang/dhall-lang/pull/77

cc: @minimal


minimal:
Great thanks!

PR #236, opened 2018-01-28
Improve error messages for list elements with the wrong type

### LINKED ISSUE #230: Error reporting for list element types happens after substitution
Apologies for the brevity but let me just vomit out my current context. We have this in a Dhall file:

```
...
      , library            =
          [   ./dhall/defaults/Library.dhall 
            ⫽ { build-dependencies =
                  [ bounded "base" [ +4, +9, +0, +0 ] [ +5 ]
                  , before "ansi-wl-pprint" [ +0, +7 ]
                  , before "base16-bytestring" [ +0, +2 ]
                  , before "bytestring" [ +0, +11 ]
                  , before "case-insensitive" [ +1, +3 ]
...
```

`build-dependencies` goes on for a while

Yet when I type check this:

```
    Your first ❰List❱ element has this type:                                        
                                                                                    
    ↳ { bounds : VersionRange, package : Text }                                                                
                                                                                    
    ... but the following element at index 18:                          
                                                                                    
    ↳ (λ(package : Text) → λ(min : List Natural) → λ(max : List Natural) → { bounds = intersectVersionRanges (orLaterVersion min) (earlierVersion max), package = package }) "text-format" [+0, +4]                                                                
                                                                                    
    ... has this type instead:                                                      
                                                                                    
    ↳ ∀(max : List Natural) → { bounds : VersionRange, package : Text }                                                                
```

The 18th element is just some function application though:

```dhall
                  , bounded "text-format" [ +0, +4 ]
```

Gabriel439:
Fixes #230

The type error for the list now returns the expression for the mismatched
element instead of the entire list.  This improves the source span
displayed to the user to focus on the invalid element.

Since the source span now matches the invalid element, there is no need
to redisplay the same element in the detailed error message.

### DIFF
--- a/src/Dhall/TypeCheck.hs
+++ b/src/Dhall/TypeCheck.hs
@@ -380,7 +380,7 @@ typeWithA tpa = loop
                             let nf_t  = Dhall.Core.normalize t
                             let nf_t' = Dhall.Core.normalize t'
                             let err   = MismatchedListElements i nf_t x nf_t'
-                            Left (TypeError ctx e err) )
+                            Left (TypeError ctx x err) )
                 return (App List t)
     loop ctx e@(ListLit (Just t ) xs) = do
         s <- fmap Dhall.Core.normalize (loop ctx t)
@@ -394,7 +394,7 @@ typeWithA tpa = loop
                 else do
                     let nf_t  = Dhall.Core.normalize t
                     let nf_t' = Dhall.Core.normalize t'
-                    Left (TypeError ctx e (InvalidListElement i nf_t x nf_t')) )
+                    Left (TypeError ctx x (InvalidListElement i nf_t x nf_t')) )
         return (App List t)
     loop ctx e@(ListAppend l r  ) = do
         tl <- fmap Dhall.Core.normalize (loop ctx l)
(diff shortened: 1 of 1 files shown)


Gabriel439:
The only reason I haven't done this yet is because the error messages are not part of the API contract and they are pretty brittle to maintain

I think if I did this it would be to add functions for automating common error message patterns (i.e. wrapping code in ASCII boxes, wrapping to 80 columns) and testing those functions.

ocharles:
Golden tests don't really imply that, they are just a mega easy way to test things which have a known output. That output can still change over time, of course, it's just trivial to say "yea yea, my changes are what I wanted". I've come to really enjoy them. In a sense, golden testing the error messages against Dhall expressions also stresses the type checker, too.

Gabriel439:
Alright, but there's still one other thing that has to be implemented, which is an easy way to update the tests to accept the new golden output.  For example, if you significantly refactor error messages and 50 tests start failing, it would be time-consuming to update all 50 of them manually.

ocharles:
With golden testing, you have a runner that just accepts the new result. With `tasty-golden`, the typical flow is to review the differences, and then just re-run with `--accept`, which rewrites the golden output files.

Gabriel439:
Ah, I didn't know that `tasty-golden` supported something like that.  I'll have to try it out

PR #237, opened 2018-01-28
`dhall-format` now can output multi-line strings

### LINKED ISSUE #229: dhall-format breaks multi-line strings
```
[nix-shell:~/work/dhall-to-cabal]$ dhall-format <<EOF
> ''
> Hello
> World
> ''
> EOF
"Hello\nWorld\n"
```

Gabriel439:
Fixes #229

This updates `dhall-format` to use multi-line string literals whenever
pretty-printing a string with at least one newline character.


src/Dhall/Core.hs
```diff
@@ -609,10 +609,35 @@ prettyNatural = Pretty.pretty
 prettyDouble :: Double -> Doc ann
 prettyDouble = Pretty.pretty
 
+-- TODO: Correctly escape multi-line literals
```
ocharles: What's the todo here about? What case isn't dealt with?

Gabriel439: Oh, whoops!  The TODO is a very real issue I still need to fix

PR #238, opened 2018-01-28
Use `lens-family-core` instead of `lens`

Gabriel439:
Related to https://github.com/fpco/stackage/issues/3238

The immediate motivation of this change is to fix the upper bound issue
linked above.  However, since we don't need the full `lens` dependency
this uses the much smaller and more stable `lens-family-core` library.  Besides
reducing the footprint of the `dhall` executable this should hopefully also
reduce the number times we need to update the upper bound.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,9 +1,9 @@
 { mkDerivation, ansi-wl-pprint, base, base16-bytestring, bytestring
 , case-insensitive, charset, containers, contravariant, cryptohash
-, deepseq, exceptions, http-client, http-client-tls, lens
-, optparse-generic, parsers, prettyprinter, stdenv, system-fileio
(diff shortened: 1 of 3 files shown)


PierreR:
Is the hope to reduce footprint related to #240 given that `trifecta` has a dependency on [lens](https://github.com/ekmett/trifecta/blob/master/trifecta.cabal#L82) contrary to megaparsec ?

Gabriel439:
Yeah, you're right that this would not significantly reduce the footprint unless this library also migrates off of `trifecta`.  However, I think the better motivation is to reduce the number of times the upper bound is bumped

PierreR:
With the new caret cabal syntax this problem will hopefully fade away in a couple of years or so. Thanks for the reply.

Issue #244, opened 2018-02-01
dhall-format changes too much


ocharles:
This is gonna be a tricky one, but I think `dhall-format` does too much. The biggest annoyances for me are

1. Reordering records. In `dhall-to-cabal` the records have a "logical" flow, especially with key/value pairs. In https://github.com/ocharles/dhall-to-cabal/blob/d7ad78df7f98ba5a74bf001f3f826eb62af01a27/dhall-to-cabal.dhall#L27 I really want `name` to come first and then the `executable` to follow. Having `name` come after the `executable` unfortunate. I may introduce a function here though, to at least make that sequencing more mandatory (and remove some syntactic weight).

2. Unpredictable line breaks. The algorithm taken at the moment is one of terseness, only splitting when necessary. I think `elm-format` got things right here, though. With `elm-format`, many expressions have a single line way of formatting and a multi-line way of formatting. If you don't have a line-break, you only get the single line formatting. Introduce at least one line break and the expression will be formatted in the multi-line style, which may use many, many, more lines than just two. For Dhall, we could imagine single-line formatting for a record, and the multi-line formatting which places every field on a new line.

Gabriel439:
(2) should be straightforward to do and is actually pretty close to how Dhall already works.  Every pretty-printed Dhall expression already has a single-line form and a multi-line form and if you run over 80 colums then the multi-line form is preferred.  Based on your description, the difference between Elm and Dhall is that Dhall independently formats the key-value pairs within a record from the record skeletone, but that's pretty easy to fix to match Elm (by having them both select single-line or multi-line form in sync with each other).

(1) is slightly trickier but still doable.  The only minor issue is that the Dhall AST doesn't preserve the order of field names since it uses a `Map` internally to store records and unions.  However, the AST could be changed (a breaking change, but I'm fine with it) to preserve the original order of field names for formatting purposes.

ocharles:
> Based on your description, the difference between Elm and Dhall is that Dhall independently formats the key-value pairs within a record from the record skeleton, but that's pretty easy to fix to match Elm (by having them both select single-line or multi-line form in sync with each other).

More that `elm-format` does not consider line length at all. If you want to put everything on one line, you can do that. The moment you introduce a single line break, it does enough to split that expression up into multi-line form.

> However, the AST could be changed (a breaking change, but I'm fine with it) to preserve the original order

[...]

PR #246, opened 2018-02-01
`dhall-format` now preserves the order of fields

Gabriel439:
Related to #244

This updates Dhall's syntax tree to use an insert-ordered hashmap to store
record fields and union alternatives.  This in turn implies that they will be
formatted in the same order that they were originally parsed.

This is a breaking change to the API due to changing the type of map used in the
syntax tree.


src/Dhall.hs
```diff
@@ -480,12 +481,17 @@ pair :: Type a -> Type b -> Type (a, b)
...
+          <*> ( Data.HashMap.Strict.InsOrd.lookup "_2" fields >>= extract r )
     extractOut _ = Nothing
 
-    expectedOut = Record (Data.Map.fromList [("_1", expected l)
-                                            ,("_2", expected r)])
+    expectedOut =
+        Record
+            (Data.HashMap.Strict.InsOrd.fromList
+                [ ("_1", expected l)
+                ,("_2", expected r)
```
ocharles: Obviously minor, but you've lost a bit of formatting.

PR #248, opened 2018-02-01
Remove `system-filepath`/`system-fileio` dependencies

### LINKED ISSUE #239: Remove `system-filepath`/`system-fileio` dependencies
`dhall` doesn't really need improved type safety for filepaths, so I don't mind getting rid of these dependencies (i.e. replace them with just `FilePath` from the `Prelude` and operations from the `filepath` library)

PierreR:
Fix #239


PierreR:
This is a first WIP draft.

Please point me to obvious mistake.
As discussed one is probably that `takeDirectory` cannot be a substitute to `parent`

src/Dhall/Import.hs
```diff
@@ -411,11 +403,11 @@ canonicalize (File hasHome0 file0:paths0) =
 
         -- `clean` will resolve internal @.@/@..@'s in @currPath@, but we still
         -- need to manually handle @.@/@..@'s at the beginning of the path
-        combine url path = case Filesystem.stripPrefix ".." path of
+        combine url path = case List.stripPrefix ".." path of
```
Gabriel439: `Data.List.stripPrefix` behaves slightly differently from `Filesystem.stripPrefix`:

```haskell
>>> Filesystem.Path.CurrentOS.stripPrefix ".." "../foo"
Just (FilePath "foo")
>>> Data.List.stripPrefix ".." "../foo"
Just "/foo"
```
I believe this should be `List.stripPrefix "../" instead

src/Dhall/Import.hs
```diff
@@ -411,11 +403,11 @@ canonicalize (File hasHome0 file0:paths0) =
...
             Just path' -> combine url' path'
               where
                 url' = parentURL (removeAtFromURL url)
-            Nothing    -> case Filesystem.stripPrefix "." path of
+            Nothing    -> case List.stripPrefix "." path of
```
Gabriel439: Similarly, I believe this should be `List.stripPrefix "./"` instead

src/Dhall/Import.hs
```diff
@@ -470,17 +460,17 @@ removeAtFromURL url
...
+    then FilePath.takeDirectory fp
+    else fp
 
 -- | Remove all @.@'s and @..@'s in the path
 clean :: FilePath -> FilePath
-clean = strip . Filesystem.collapse
+clean = strip . FilePath.normalise
   where
-    strip p = case Filesystem.stripPrefix "." p of
+    strip p = case List.stripPrefix "." p of
```
PierreR: I haven't already but I guess I also need to change "." for "./" here ?

PierreR: Do we still need to call `strip` ? (tests are passing either way)

Gabriel439: You're right that we don't need `stripPrefix` here because `System.FilePath.normalise` removes any preceding `./`s (unlike `Filesystem.Path.collapse`)

PierreR:
Let me know if you are expecting something more from this PR. I can rebase to avoid the conflict if it is more convenient on your side.

Gabriel439:
Thank you for contributing this :)

PR #259, opened 2018-02-13
Reject custom contexts that are ill-formed

### LINKED ISSUE #253: Custom contexts with a "loop" loop the import resolution process
I accidentally wrote the following

```haskell
cabalContext
  :: Ctx.Context ( Expr.Expr Dhall.Parser.Src Dhall.TypeCheck.X )
cabalContext =
  let
    versionRangeType =
      Dhall.expected versionRange

  in
  Ctx.empty
    & Ctx.insert "VersionRange" versionRangeType
```

with

```haskell
versionRange :: Dhall.Type Cabal.VersionRange
versionRange =
  let
    extract =
      ..

    expected =
      Expr.Var ( Expr.V "VersionRange" 0 )

  in Dhall.Type { .. }
```

See the problem? I've essentially said that `VerisonRange` has type `Var "VersionRange"`, which seems to loop the type checker. This was super subtle and took me the best part of 15 minutes to track down, would be great if we could catch this nonsense up front.

Gabriel439:
Fixes #253

This protects `typeWith` to reject any custom-context that might trigger a
type-checking loop


src/Dhall/TypeCheck.hs
```diff
@@ -3254,3 +3256,18 @@ instance (Buildable a, Buildable s) => Buildable (DetailedTypeError s a) where
         source = case expr of
             Note s _ -> build s
             _        -> mempty
+
+{-| This function verifies that a custom context is well-formed so that
+    type-checking will not loop
+-}
+checkContext :: Context (Expr s X) -> Either (TypeError s X) ()
```
ocharles: I think you meant to export this. If not, please export it :smile:

Gabriel439: Actually, I intentionally did not export it since `typeWith` calls it once before entering the type-checking loop.  This avoids the performance issue since it's not called on every iteration of the loop

ocharles:
Isn't it called twice though in the case where you are using the import
mechanism? It would still be nice to be able to do the check yourself.

Gabriel439:
Oh, whoops!  You're right that there is one unnecessary call to `checkContext` since it already uses `typeWith` which already uses `checkContext`.  I'll fix that

I'll also go ahead and export it, too

ocharles:
Oh, I actually meant https://hackage.haskell.org/package/dhall-1.9.1/docs/src/Dhall-Import.html#loadWithContext which calls typeWith, so we end up validating the context twice if you do import resolution and then final type checking. It's hardly the end of the world though.

Gabriel439:
Imports are type-checked twice anyway even without custom contexts.  This is normal to ensure that imports are closed and also to localize error messages better.

ocharles:
Sure, it's just a pinch more overhead. But I think it's inconsequential

Gabriel439:
I'll go ahead and merge this, then, since I think I've addressed your requests

PR #260, opened 2018-02-13
First pass at adding semantic annotations to the pretty printer

ocharles:
This is a work in progress, but I wanted to get your feedback before I went any further. Here's how things look with these changes:

2018-02-13-100834_826x1086_scrot

### DIFF
--- a/src/Dhall/Core.hs
+++ b/src/Dhall/Core.hs
@@ -497,7 +497,7 @@ instance Buildable a => Buildable (Expr s a) where
     build = buildExpr
 
 instance Pretty a => Pretty (Expr s a) where
-    pretty = prettyExpr
+    pretty = Pretty.unAnnotate . prettyExpr
 
 {-| `shift` is used by both normalization and type-checking to avoid variable
     capture by shifting variable indices
--- a/src/Dhall/Pretty/Internal.hs
+++ b/src/Dhall/Pretty/Internal.hs
@@ -99,7 +99,7 @@ dot      = syntax "."
 
 -- | Pretty-print a list
 list :: [Doc Ann] -> Doc Ann
-list   [] = literal "[]"
+list   [] = lbracket <> rbracket
 list docs =
     enclose
         (lbracket <> space)
@@ -112,7 +112,7 @@ list docs =
 
 -- | Pretty-print union types and literals
 angles :: [(Doc Ann, Doc Ann)] -> Doc Ann
-angles   [] = literal "<>"
+angles   [] = langle <> rangle
 angles docs =
     enclose
         (langle <> space)
@@ -125,7 +125,7 @@ angles docs =
 
 -- | Pretty-print record types and literals
 braces :: [(Doc Ann, Doc Ann)] -> Doc Ann
-braces   [] = literal "{}"
+braces   [] = lbrace <> rbrace
 braces docs =
     enclose
         (lbrace <> space)
@@ -713,7 +713,7 @@ prettyRecord =
 prettyRecordLit :: Pretty a => InsOrdHashMap Text (Expr s a) -> Doc Ann
 prettyRecordLit a
     | Data.HashMap.Strict.InsOrd.null a =
-        literal "{=}"
+        lbrace <> equals <> rbracke
     | otherwise
         = braces (map (prettyKeyValue equals) (Data.HashMap.Strict.InsOrd.toList a))
 
--- a/src/Dhall/Pretty/Internal.hs
+++ b/src/Dhall/Pretty/Internal.hs
@@ -272,7 +272,7 @@ prettyChunks (Chunks a b) =
         docs =
             Data.List.intersperse Pretty.hardline (fmap Pretty.pretty lazyLines)
 
-    prettyChunk (c, d) = prettyText c <> "${" <> prettyExprA d <> "}"
+    prettyChunk (c, d) = prettyText c <> syntax "${" <> prettyExprA d <> syntax rbrace
 
     prettyText t = literal (Pretty.pretty (Builder.toLazyText (escapeText t)))
 
--- a/src/Dhall/Pretty/Internal.hs
+++ b/src/Dhall/Pretty/Internal.hs
@@ -713,7 +713,7 @@ prettyRecord =
 prettyRecordLit :: Pretty a => InsOrdHashMap Text (Expr s a) -> Doc Ann
 prettyRecordLit a
     | Data.HashMap.Strict.InsOrd.null a =
-        lbrace <> equals <> rbracke
+        lbrace <> equals <> rbracket
     | otherwise
         = braces (map (prettyKeyValue equals) (Data.HashMap.Strict.InsOrd.toList a))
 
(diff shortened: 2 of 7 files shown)


ocharles:
I forgot to commit `Dhall.Pretty` but it's not very interesting - it just maps `DhallAnn` to `AnsiStyle` to specify the colours. I do propose that we move all the pretty printing logic into this module though, `Dhall.Core` is pretty huge.

ocharles:
I just realised I can't move everything into `Dhall.Pretty` due to good old orphan instances (the `Pretty Expr` instance).

Gabriel439:
You still be able to split the pretty-printing logic out by using GHC's support for mutually recursive modules.  See:

https://wiki.haskell.org/Mutually_recursive_modules#GHC

I think it would be worthwhile to split out the pretty-printing logic into a separate module if we can get the mutually recursive modules working

src/Dhall/Core.hs
```diff
@@ -627,7 +638,7 @@ prettyChunks (Chunks a b) =
             (foldMap prettyMultilineChunk a <> prettyMultilineBuilder b)
         <>  "''"
         )
-    else "\"" <> foldMap prettyChunk a <> prettyText b <> "\""
+    else Pretty.annotate Literal ("\"" <> foldMap prettyChunk a <> prettyText b <> "\"")
```
Gabriel439: Do you plan to also annotate multi-line string literals?

src/Dhall/Core.hs
```diff
@@ -494,6 +496,11 @@ instance IsString (Chunks s a) where
     case the corresponding builder.
 -}
 
+data DhallAnn = Keyword | Variable | Punctuation | Label | Literal
```
Gabriel439: Perhaps add a separate `Reserved` category for built-in sorts/kinds/types/values (i.e. `Type`/`Kind`/`Bool`/`True`) instead of classifying them as `Keyword`s.  I normally associate "keyword" with things like `if`/`then`/`else` but not built-in functions or values like `True` or `Natural/even`

Also, perhaps rename `Punctuation` to `Operator` since it looks like you treat record/union/list literals as `Literal`s (which I think makes sense)

ocharles:
> I think it would be worthwhile to split out the pretty-printing logic into a separate module if we can get the mutually recursive modules working

Oh sure, I know about `hs-boot`, I just figured you wouldn't want that. If you're ok with it, I'll do that change.

src/Dhall/Core.hs
ocharles: Yep, just haven't got that far.

src/Dhall/Core.hs
ocharles: I think `Reserved` (or `Builtin`?) is good for builtin values, separating them from keywords which are syntactic. I think I will rename `Punctuation` to `Syntax` (parens, braces, commas, etc). Dhall's operators would be `Reserved`/`Builtin` (here my preference is `Builtin`).

ocharles:
Ok, I've force pushed a finished version of this PR. 

To test this, I ran

```
find . -name '*.dhall' | xargs -l ./dist/build/dhall-format/dhall-format --inplace
```

with `master` checked out.

I then switched over to this branch, recompiled `dhall-format` and ran the same command. There are no changes.

Gabriel439:
Looks like you need to resolve a merge conflict

Also, note that you don't need to force push when you add new changes to your pull request (and I discourage it since it mutates history and messes with commit comments).  This repository enforces that pull requests are merged with "Squash and merge" in order to generate a single commit, so you don't need to keep your pull request branch clean

ocharles:
Eh, it was more that I literally restarted the branch at work. Ingenerally
don't force push. Will keep that in mind.

ocharles:
Made a bit of a mess resolving that merge conflict and have opted for a force push. I know we agreed not to do that, sorry!

Gabriel439:
@ocharles: It's alright!  It's more informational advice :)

src/Dhall/Pretty/Internal.hs
```diff
@@ -624,29 +709,29 @@ prettyKeyValue separator (key, value) =
...
 prettyRecord =
-    braces . map (prettyKeyValue ":") . Data.HashMap.Strict.InsOrd.toList
+    braces . map (prettyKeyValue colon) . Data.HashMap.Strict.InsOrd.toList
 
-prettyRecordLit :: Pretty a => InsOrdHashMap Text (Expr s a) -> Doc ann
+prettyRecordLit :: Pretty a => InsOrdHashMap Text (Expr s a) -> Doc Ann
 prettyRecordLit a
     | Data.HashMap.Strict.InsOrd.null a =
-        "{=}"
+        literal "{=}"
```
Gabriel439: I would classify this as `Syntax`, or at least the outer braces should be treated syntax for consistency with non-empty records

Gabriel439:
Actually, two more things:

How does this behave when piped to a non-interactive output or when used `--inplace`?  Does it emit color escape codes?

Be careful about using the color white since some terminals (like the default terminal on OS X) use a white background.  I would change that to uncolored for now so that it defaults to the terminal's foreground color.  I'll save a more detailed color bikeshed for later but I just wanted to mention that

ocharles:
> How does this behave when piped to a non-interactive output or when used --inplace? Does it emit color escape codes?

With `--inplace` it takes a different code path that just uses `pretty` rather than `prettyExpr`. That means all annotations are dropped and we just do normal text rendering. I'll check for pipeing later tonight, but I think I tested it and it was OK.

src/Dhall/Pretty/Internal.hs
ocharles: Yea, I couldn't really decide on this. I think you could argue it either way. I might annotate the whole thing as a literal, but then the braces as syntax.

src/Dhall/Pretty/Internal.hs
```diff
@@ -41,7 +44,23 @@ import qualified Data.Text.Prettyprint.Doc             as Pretty
 import qualified Data.Text.Prettyprint.Doc.Render.Text as Pretty
 import qualified Data.Vector
 
-prettyExpr :: Pretty a => Expr s a -> Doc ann
+data Ann
+  = Keyword     -- ^ Used for syntactic keywords
+  | Variable    -- ^ Bound variables
+  | Syntax      -- ^ Syntax punctuation such as commas, parenthesis, and braces
+  | Label       -- ^ Record labels
```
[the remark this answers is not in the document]
ocharles: Ok, so you suggest removing `Label` and changing it's uses to `Variable`? I'm ok with that.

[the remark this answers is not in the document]
ocharles: Maybe just `Name` is a better choice than `Variable`?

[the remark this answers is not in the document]
Gabriel439: I would use `Label` for both because that's what the grammar and the code calls both of them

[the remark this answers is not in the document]
ocharles: Sold! :hammer:

ocharles:
Ok, I think I've addressed all comments.

Gabriel439:
Looks great!  Thanks for doing this :)

PR #263, opened 2018-02-14
Move pretty printing out into Dhall.Pretty

ocharles:
This commit splits out the pretty printing logic from Dhall.Core into its own
module. Dhall.Core actually relies on some "internal" pretty printing functions,
so pretty printing has been moved to Dhall.Pretty.Internal - Dhall.Pretty
exposes the public facing interface.


PierreR:
@ocharles out of curiosity what are the internal pretty printing functions ? If we are already using internals maybe the use of `viaShow` and `unsafeViaShow` is pertinent as these should actually be exposed and will be in a future version (https://github.com/quchen/prettyprinter/commit/1d7cb09b21d213abc16c5a7068c8c328f591dd0b)

src/Dhall/Pretty.hs
```diff
@@ -0,0 +1,3 @@
+module Dhall.Pretty (prettyExpr) where
+
+import Dhall.Pretty.Internal
```
Gabriel439: What's the reason for the separate `Dhall.Pretty.Internal` module?  Is this related to the use of mutually recursive modules?

ocharles: As mentioned in the PR description:

> Dhall.Core actually relies on some "internal" pretty printing functions,
so pretty printing has been moved to Dhall.Pretty.Internal - Dhall.Pretty
exposes the public facing interface.

Does that help? It's things like `Core` needing `buildConst` (see the `Internal.hs-boot` for exactly what `Core` requires).

Gabriel439: Oh, whoops!  I missed that.  That makes sense

Gabriel439:
Thank you! :)

PR #266, opened 2018-02-15
Add dhalli, an interactive REPL for Dhall

ocharles:
This commit introduces dhalli, a fairly minimal Dhall REPL that supports:

* Evalution of Dhall expressions along with import resolution
* Type inspection of abritrary expressions
* Syntax highlighting and pretty printing for output
* Persistent bindings with :let
* One-expression lookup with an implicit "it" binding


ocharles:
I've based this on the semantic pretty printing branch, as I want syntax highlighting for output.

Gabriel439:
@ocharles: Note that this still needs to resolve the merge conflicts from the original pretty-printing branch

Gabriel439:
Now that the other branches are merged, could you merge `master` into this branch to reduce the diff?

dhalli/Main.hs
```diff
@@ -0,0 +1,235 @@
...
+                ( Binding { bindingType = t, bindingExpr = expr } )
+                ( envBindings e )
+          }
+    )
+
+  output 
+    ( Expr.Annot ( Expr.Var ( Dhall.V varName 0 ) ) t )
+
+addBinding _ =
+  error ":let should be of the form `:let x = y`"
```
Gabriel439: You probably want to use `liftIO . fail` instead of `error`.  The issue with `error` is that it is not evaluation-safe:

```
error message `seq` () = error message
fail  message `seq` () = ()
```

dhall.cabal
```diff
@@ -147,6 +147,22 @@ Executable dhall
     Other-Modules:
         Paths_dhall
 
+Executable dhalli
```
Gabriel439: Perhaps call this `dhall-repl` instead

dhalli/Main.hs
```diff
@@ -0,0 +1,235 @@
...
+  ]
+
+
+completer :: Monad m => Repline.WordCompleter m
+completer _ =
+  return []
+
+
+greeter :: MonadIO m => m ()
+greeter =
```
Gabriel439: I would skip the greeter.  I try to keep the tooling minimal and unintrusive

dhalli/Main.hs
```diff
@@ -0,0 +1,235 @@
...
+addBinding _ =
+  error ":let should be of the form `:let x = y`"
+
+
+options
+  :: ( Haskeline.MonadException m, MonadIO m, MonadState Env m )
+  => Repline.Options m
+options =
+  [ ( "type", dontCrash . typeOf )
+  , ( "let", dontCrash . addBinding )
```
Gabriel439: Another useful option would be `save`:

```
> :save ./x = 1
> ./x
Integer

1
```

dhalli/Main.hs
```diff
@@ -0,0 +1,235 @@
...
+        ( dontCrash . eval )
+        options
+        ( Repline.Word completer )
+        greeter
+    )
+    emptyEnv
+
+
+data Env = Env
+  { envBindings :: StrictMap.Map Dhall.Text Binding
```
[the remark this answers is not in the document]
ocharles: > If you used a Map for that then the second x would override the first x and you would turn y from a well-typed to an ill-typed term.

That's not quite true, as expressions are normalized before putting them into the environment. So in this case, you'd add `(x, 1)`, then `(y, 1)`, then replace `(x, True)`.

You're otherwise right though, so I'll make that change.

ocharles:
I've addressed all your suggestions except for `:save`. There's probably a lot of things that could be added, but I think we can work those out in a separate PR.

Gabriel439:
Thanks!  This looks amazing :)

PR #267, opened 2018-02-16
Have dhall use syntax highlighting if ANSI is supported

ocharles:


Gabriel439:
Great idea!  Thanks :)

PR #268, opened 2018-02-16
Replace trifecta with megaparsec

### LINKED ISSUE #240: Switch to `megaparsec`
Related to #108, I think there may be parsing performance improvements by switching from `trifecta`/`parsers` to `megaparsec` since `megaparsec` provides high-efficiency bulk parsing primitives while still preserving good error messages.

ocharles:
For me tests go *from* 1.08s to 1.6s :disappointed: This isn't a finished PR (I only did enough to build `lib:dhall` and `test:test`), stopping here for feedback. I think it's probably still worth doing, as `megaparsec` has more eyes on it, and more scope for performance tuning. If you're happy with this I'll merge in `master` and complete the work.

Fixes #240.


src/Dhall/Parser.hs
```diff
@@ -101,26 +91,34 @@ instance Monoid a => Monoid (Parser a) where
...
+  notFollowedBy = Text.Megaparsec.notFollowedBy
+
+instance Text.Parser.Char.CharParsing Parser where
+  satisfy = Parser . Text.Megaparsec.Char.satisfy
 
 instance TokenParsing Parser where
     someSpace =
         Text.Parser.Token.Style.buildSomeSpaceParser
-            (Parser someSpace)
+            (Parser (Text.Megaparsec.skipSome (Text.Megaparsec.Char.satisfy Data.Char.isSpace)))
```
PierreR: Do you thing replacing `satisfy isSpace` with something ala `sc = L.space space1 lineComment blockComment` might make a difference performance wise ?

ocharles: No idea! You're welcome to grab my branch and see :) I won't be able to check that for a while, probably not free to hack now until Sunday.

src/Dhall/Parser.hs
```diff
@@ -101,26 +91,34 @@ instance Monoid a => Monoid (Parser a) where
...
+
+instance Text.Parser.Combinators.Parsing Parser where
+  unexpected = fail
+  eof = Parser Text.Megaparsec.eof
+  try = Text.Megaparsec.try
+  (<?>) = (Text.Megaparsec.<?>)
+  notFollowedBy = Text.Megaparsec.notFollowedBy
+
+instance Text.Parser.Char.CharParsing Parser where
+  satisfy = Parser . Text.Megaparsec.Char.satisfy
```
Gabriel439: You will probably get a speed improvement if you implement the `text` method of the `CharParsing` class with a high-performance primitive (and probably the rest of the class, too).  One of the reasons for using `megaparsec` is that it provides bulk parsing primitives

The other reason I suggest this is because that is where the parsing bottleneck is on `master`

Gabriel439:
@ocharles: Also, if you can push your branch to this repository then I can open a pull request against your branch to get the rest of the library to build

However, I think we should get at least similar performance before merging this.  A 60% slowdown is still a lot

ocharles:
Ok, this is now up to date against `master`, and all components build. On my laptop:

* `master` tests run in ~1s
* `ollie/megaparsec` tests run in ~2.2S.

We still have work to do. I have at least implemented all of the `parsers` classes, but note that `text` and `string` do a lot of packing and unpacking. That might not be helping things, I'm going to look at that next.

The branch has been pushed to this repository.

Gabriel439:
Yeah, in my own branch based on yours I found that implementing the `text` method in terms of `string` actually slowed down performance.  The next thing I'm trying is Chris Done's suggestion of using `attoparsec` to tokenize the input but using `megaparsec` to parse based on the tokens

PierreR:
> Yeah, in my own branch based on yours I found that implementing the text method in terms of string actually slowed down performance.

I have experienced the same things ... I am bit puzzled by the `IsString` instance in the Parser module and the Chunk datatype. Isn't this what `Stream` is about in `megaparsec`  ?

Gabriel439:
I have a branch with an example use of `attoparsec` to tokenize the input and a benchmark: https://github.com/dhall-lang/dhall-haskell/tree/337d363151a3be60be01cc8a5636b4c19c9da1e2

It's very fast, so if lexing is the bottleneck that should improve performance a lot.  There's still a lot of work to do, though (like properly handling string interpolation, which I think is still possible)

vmchale:
What's the status of this? I may be able to write an `alex`-based lexer, but I do not wish to tread on any toes.

Gabriel439:
So my inclination at this point is to accept the switch to `megaparsec` even if it takes a performance hit.  We just need to clean this up and complete the transition.

I do think we should do a separate lexing phase at some point but the first step should be to complete the transition to `megaparsec` since `trifecta`/`parsers` does not permit customizing the token type like `megaparsec` does.  Once that is complete we can move forward with changing the token type from characters to lexed tokens.

src/Dhall/Parser.hs
```diff
@@ -106,26 +95,52 @@ instance (Data.Semigroup.Semigroup a, Monoid a) => Monoid (Parser a) where
...
 
 instance TokenParsing Parser where
     someSpace =
         Text.Parser.Token.Style.buildSomeSpaceParser
-            (Parser someSpace)
+            (Parser (Text.Megaparsec.skipSome (Text.Megaparsec.Char.satisfy Data.Char.isSpace)))
             Text.Parser.Token.Style.haskellCommentStyle
 
-    nesting (Parser m) = Parser (nesting m)
+    highlight _ = id
```
Gabriel439: Note that this is the default implementation of `highlight` so you can omit it

src/Dhall/Parser.hs
```diff
@@ -1531,16 +1546,16 @@ import_ = (do
         return RawText
 
 -- | A parsing error
-newtype ParseError = ParseError Doc deriving (Typeable)
+newtype ParseError = ParseError{ unwrap :: Text.Megaparsec.ParseError Char Void}
 
 instance Show ParseError where
-    show (ParseError doc) =
-      "\n\ESC[1;31mError\ESC[0m: Invalid input\n\n" <> show doc
+    show (err) =
```
Gabriel439: Minor nitpick: You don't need parentheses here any longer

Gabriel439:
I also created a pull request against your branch that includes these changes and a change to use the `parseErrorPretty'` function: https://github.com/ocharles/dhall-haskell/pull/2

Gabriel439:
Awesome!  Thanks for doing this :)

chrisdone:
Just to add some encouragement, it's really easy to write a slow parser. So even translating to megaparsec is just step one. Avoiding inefficient combinators and backtracking etc. should help.

I have a similar issue for a language parser I'm writing where I have splicing like string substitution which involves parsing two languages at once. A tokenizer in attoparsec can handle this i.e. to parse into chunks of LanguageOne and LanguageTwo, I've implemented one (parses e.g. 100KB of multi-level nesting in 20ms), but the error messages for a misbalanced delimiter in attoparsec are still useless. So, I'm considering either implementing the tokenizer with a manual bytestring parser, which can show useful error messages, or implementing the tokenizer in megaparsec. It's only a page of code so I'll probably try out megaparsec and see how it shapes up speed-wise against attoparsec.

Gabriel439:
@chrisdone: One solution which we're considering as an intermediate approach is to use the `parsers` type class to take advantage of both libraries: we first try to parse using `attoparsec` entirely for speed on the happy path and then if parsing fails we fall back on reparsing using `megaparsec` for better error messages

Then after that the next step is what you suggested: using `attoparsec` for lexing and `megaparsec` for parsing.  I've already done a very rough first draft of lexing using `attoparsec` and I was able to reproduce much faster throughput like you said.  However, switching to a lexer requires a bit more effort than what I've invested so far because the language has some constructs that are not friendly to lexing, like nested block comments or string interpolation.  I don't think it's impossible, but it just requires more work.

chrisdone:
> One solution which we're considering as an intermediate approach is to use the parsers type class to take advantage of both libraries: we first try to parse using attoparsec entirely for speed on the happy path and then if parsing fails we fall back on reparsing using megaparsec for better error messages

Right, I saw your comment on using a parsers type class. That approach sounds interesting. In the case of Dhall I imagine most code you ever parse is correct, so the successful path will happen more often. And only in dev will you actually see a parse error. So that's an interesting idea! I'd be interested to see updates on that.

> I've already done a very rough first draft of lexing using attoparsec and I was able to reproduce much faster throughput like you said. However, switching to a lexer requires a bit more effort than what I've invested so far because the language has some constructs that are not friendly to lexing, like nested block comments or string interpolation. I don't think it's impossible, but it just requires more work.

Indeed, I have the same interpolation problem, trying to do a lex on syntax like this:

```shell
demo = { # quoted 
  quoted $var; 
  echo "quoted ${} \"foo\" string";
  ${let i = [1,2,3] # unquoted 
        p = {ls -al; this is; # quoted 
           $i ${i} code}
        cmd = {stack}
    in unquoted text
       "unquoted {} string"
       p
       unquoted text2};
  quoted;
  text2;
}
```

requires some balancing and language switching between the "shell mode" and the "haskell-like mode" (`foo $(bar {mu $(foo)})` etc). Lexing this is fast to make something simply producing boundaries (see below). Obviously, those `QuotedByteString` and `UnquotedByteString` are just the untokenized strings that could be further fleshed out, either immediately or in a later consumer. 

One nice thing about this is that I could use the library of my lang parser to provide trivial editor support for syntax highlighting by exposing the tokenizer as e.g. JSON output. I'm hesitant to go all in on a single char-by-char parser straight into the AST.

I'll have a go at trying to "drop-in" megaparsec in place of parsec here. Apparently it's capable of fast `takeWhile`-type consumers, so I'll give it a go and report my numbers vs attoparsec. :+1: 

```haskell
Right
  [ UnquotedByteString "demo = "
  , UnquotedSplice
      [ QuotedByteString " "
      , QuotedComment " quoted comment"
      , QuotedByteString "\n  quoted "
      , QuotedSpliceWord "var"
      , QuotedByteString "; "
      , QuotedComment " another"
      , QuotedByteString "\n  echo "
      , QuotedString "quoted ${}\n  \"foo\" string"
      , QuotedByteString ";\n  "
      , QuotedSplice
          [ UnquotedByteString "let i = [1,2,3] "
          , UnquotedComment " unquoted"
          , UnquotedByteString "\n        p = "
          , UnquotedSplice
              [ QuotedByteString "ls -al; this is; "
              , QuotedComment " quoted"
              , QuotedByteString "\n           "
              , QuotedSpliceWord "i"
              , QuotedByteString " "
              , QuotedSplice [UnquotedByteString "i"]
              , QuotedByteString " code"
              ]
          , UnquotedByteString "\n        cmd = "
          , UnquotedSplice [QuotedByteString "stack"]
          , UnquotedByteString "\n    in unquoted text\n       "
          , UnquotedString "unquoted {} string"
          , UnquotedByteString "\n       p\n       unquoted text2"
          ]
      , QuotedByteString ";\n  quoted;\n  text2;\n"
      ]
  , UnquotedByteString "\n"
  ]
```

chrisdone:
So for a word for word drop in replacement, megaparsec is about 1.5x the time of attoparsec: https://gist.github.com/chrisdone/349c07bf80f89fc5482331cce20199e4

```
benchmarking attoparsec
time                 1.686 ms   (1.679 ms .. 1.694 ms)
mean                 1.678 ms   (1.672 ms .. 1.686 ms)
std dev              25.18 μs   (19.12 μs .. 32.91 μs)

benchmarking megaparsec
time                 2.646 ms   (2.627 ms .. 2.667 ms)
mean                 2.623 ms   (2.615 ms .. 2.634 ms)
std dev              30.27 μs   (22.03 μs .. 41.35 μs)
```

it seems like a reasonable trade-off. It's not like it's 5x or 10x slower. 

I don't know whether there's a lot of room for performance improvement in megaparsec.

chrisdone:
The difference in error messages if I omit a closing `}`:
```haskell
>  lexFile "bad.hell" -- attoparsec
endOfInput
> lexFileM "bad.hell" -- megaparsec
27:1:
   |
27 | <empty line>
   | ^
unexpected end of input
expecting "${", '"', '#', '$', or '}'

```

PR #269, opened 2018-02-16
Fix pretty-printing empty record literals

Gabriel439:
They were being printed as `{=]`


ocharles:
Hmm, did no tests catch this? I wonder if we can do something about that. Otherwise, LGTM.

PR #271, opened 2018-02-16
Fix layout when not using `--pretty`

Gabriel439:
The `layoutSmart` option appears to not behave correctly when specifying an
`Unbounded` number of columns:

https://github.com/quchen/prettyprinter/issues/49

... and the symptom of this problem is that the `dhall` executable by
default (i.e. without `--pretty`) renders all expressions in the long
format (i.e. multiline with indentation), regardless of the number of available
columns.

However, the `layoutPretty` function behaves correctly with the
`Unbounded` directive, so this change switches to using that when
rendering output without `--pretty`.

### DIFF
--- a/dhall/Main.hs
+++ b/dhall/Main.hs
@@ -114,7 +114,7 @@ makeRender options header = do
                 let doc = Pretty.pretty header <> prettyExpr e
                 render h (Pretty.layoutSmart opts doc)
             else do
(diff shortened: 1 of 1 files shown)


quchen:
Fixed the issue in the newly released `prettyprinter-1.2`. Thanks for discovering this!

Gabriel439:
You're welcome!

PR #276, opened 2018-02-17
Split out normalization code into separate files

Gabriel439:
This is part of an effort to create a language-independent compliance
test suite by splitting out actual and expected results into files that
can be shared between language implementations.

This also has the nice benefit that the Dhall code doesn't have to be embedded
and escaped within Haskell.


ocharles:
Great idea! This will come in handy for `dhall.js`.

PR #277, opened 2018-02-17
Expand haddock coverage

Gabriel439:


src/Dhall/Pretty/Internal.hs
```diff
@@ -44,20 +48,27 @@ import qualified Data.Text.Prettyprint.Doc             as Pretty
...
+-}
 data Ann
   = Keyword     -- ^ Used for syntactic keywords
   | Syntax      -- ^ Syntax punctuation such as commas, parenthesis, and braces
   | Label       -- ^ Record labels
   | Literal     -- ^ Literals such as integers and strings
   | Builtin     -- ^ Builtin types and values
 
+{-| Convert annotations to their corresponding color for syntax highlighting
+    purposes
```
ocharles: Maybe a mention that this is specifically for use with `prettyprinter-ansi-terminal`?

PR #280, opened 2018-02-18
Fix output rendering

Gabriel439:
This change fixes two issues with output rendering

The first was the following panic in the prettyprinter library when
calling `Data.Text.Prettyprinter.Doc.Render.Terminal.renderIO` on an
expression without any annotations:

```
$ dhall <<< '1' > stdout.dhall
Integer

Peeked an empty style stack! Please report this as a bug.
  error, called at src/Data/Text/Prettyprint/Doc/Render/Util/Panic.hs:38:15 in prettyprinter-1.1.1-1CDqnG9d6HQ5GZzz2F5LpU:Data.Text.Prettyprint.Doc.Render.Util.Panic
```

The fix is to use `Data.Text.Prettyprinter.Doc.Render.Text.renderIO`
instead when the expression has no annotations.

The second issue is that the color detection was not correctly
working for `stderr`, meaning that this:

```
$ dhall <<< '1' 2> stderr.dhall
```

... would write escape codes to `stderr.dhall`.  The fix is to
separately check if `stderr` supports color or not when writing
to `stderr`.


ocharles:
> `Peeked an empty style stack! Please report this as a bug.`

Have we done this? It sounds super suspicious that you can't render without annotations. Also, why is `1` not getting annotated as a literal?

ocharles:
> The fix is to use Data.Text.Prettyprinter.Doc.Render.Text.renderIO
instead when the expression has no annotations.

I'm not actually seeing this in the PR (that is, I see no check for annotation-y-ness). What have I missed?

Gabriel439:
@ocharles: There is no "check" per se for the lack of annotations, but what I meant was that any time we call `Pretty.unAnnotateS` we use `Pretty.Text.renderIO` instead of `Pretty.Terminal.renderIO`

You can reproduce the problem I described by replacing `Pretty.Text.renderIO` with `Pretty.Terminal.renderIO` if you want to see for yourself.  I don't know what the root cause and what makes things weirder is that `dhall-format` does not exhibit the same problem (although I still made the same matching change as a precaution)

ocharles:
Ahhhh, I see. @quchen, this may be of interest to you.

> . I don't know what the root cause and what makes things weirder is that dhall-format does not exhibit the same problem (although I still made the same matching change as a precaution)

I actually think we should add something in `Dhall.Pretty` of type `MonadIO m => Handle -> Expr s a -> m ()` that checks if ANSI is supported and if so outputs a pretty printed `Expr` with sane defaults. I think now `dhall`, `dhall-format` and `dhall-repl` all have their own slightly different ways of doing it.

quchen:
This is interesting – there should be no way to make the built-in renderers misbehave by anything part of the public API. Is there some way you could provide a minimum broken example that exhibits this behavior?

Gabriel439:
So the immediate cause is that `renderIO` is getting this `SimpleDocStream` after `unAnnotateS`

```
(SChar '1' (SAnnPop SEmpty))
```

... I'm still trying to narrow down why

Gabriel439:
Okay, so I've narrowed it down to an issue with `unAnnotateS`.  Here is the `SimpleDocStream` before:

```haskell
SAnnPush Literal (SChar '1' (SAnnPop SEmpty))
```
... and here is the `SimpleDocStream` after `unAnnotateS`:

```haskell
SChar '1' (SAnnPop SEmpty)
```

Gabriel439:
Alright, I narrowed this down to a small reproducing test case.  See: https://github.com/quchen/prettyprinter/issues/50

quchen:
Prettyprinter 1.2.0.1 fixes this.

ocharles:
> dhall-format does not exhibit the same problem

`dhall-format` un-annotates with `unAnnotate` on the `Doc`, rather than `unAnnotateS` on the `SimpleDoc`, that's probably why.

Gabriel439:
I'll go ahead and merge this and then follow up with another pull request to use the latest `prettyprinter` library so that we don't need the workaround

PR #281, opened 2018-02-18
Changes to syntax highlighting

Gabriel439:
The main thing this does is changing the color scheme as follows:

* Use dull green for all punctuation (including operators) and reserved
  keywords
* Use dull magenta for literals (i.e. numbers and strings)
* Use foreground color for all labels and builtin types/functions/values
* Underline builtins

Related changes include:

* Reclassifying `True`/`False` as builtins rather than literals
* Reclassifying operators as a new annotation type: `Operator`
* Including the leading `+` as part of a `Natural` literal for
  highlighting purposes

Here is the rationale for these highlighting choices:

My strongest preference is that variables should use the default foreground color
because:

* The variable names need to be the most legible part of the grammar for safety
  reasons in order to protect against confusing similarly named variables
* I assume user's terminals will be configured so that the default foreground color
  is maximally legible to them

My second preference is that built-in variables render similar to user-defined variables
names but are still distinguishable.

The reasons I believe built-in variables should render similar to user-defined variables are:

* I don't want to give built-in variables a strong syntax highlighting advantage over
  user-defined variables otherwise it might subtly disincentivize users from using
  abstraction in order to preserve coloration
* Builtins obey the grammar for variable names so I think they should render in a
  similar way
* The documentation encourages users to name variables to resemble builtin names (i.e.
  `let List/map = https://... in ...`) and those should render similar to builtins

However, I still underline builtins so that (A) users can still discern which variables are
built-in for educational purposes and discoverability and (B) so that users can quickly
discern if they mistyped a built-in name.

Given that we have no control over the default foreground color of the terminal, that
implies that if we pin built-in variables to a specific color then we can't guarantee that it
will be close to the default foreground color.  So I opted to still use the same foreground
color for built-ins but use an underline to distinguish them.  I also tried italics but that
was visually jarring in code that had even proportions of user-defined and built-in variables.
Bold is another alternative that seemed okay, but underline just seemed a little more
subtle.

I classified `True` and `False` as builtin variables for the same reason: in a pure type
system are conceptually no different from `Text`, `Type`, or `Kind`: they are built-in
variables and the fact that they are on the term-level isn't a significant distinction.
They also follow the same grammar as user-defined variables, just like all the other
built-in variables

My next preference was to use the same color for all punctuation.  It didn't have to be
dull green specifically, but that was one of the few colors that was easily legible for me on both
black and white terminal backgrounds (the other ones being vivid red and dull/vivid magenta;
both dull/vivid blue were difficult on dark backgrounds and vivid green + dull/vivid yellow + dull/vivid cyan were difficult on light backgrounds).

I also ended up using the same color for keywords, too.  My reasoning was that I wanted to
convey through the use of a single consistent color that these were all built-in syntax so that
the user always mentally associates "green" with "built-in syntax".  However, I would be open
to using a different color for keywords: my preference to use the same color for keywords and
punctuation is not that strong, but I was running out of colors to use.

Literals I left almost the same as before: they are still magenta but now dull for improved
contrast on white backgrounds.  Vivid magenta was still pretty legible, but dull magenta was
just a little bit more legible.

The reason I did not use red anywhere was that I wanted to reserve that color for error
messages so that users can tell at a glance that they got an error just by looking for the
presence of red.  That's also another reason I preferred dull magenta over vivid magenta
because it is slightly more differentiated from the vivid red of error messages.

Here are examples of how this looks on both white and dark backgrounds:

screen shot 2018-02-18 at 8 07 15 am

screen shot 2018-02-18 at 8 07 44 am

### DIFF
--- a/src/Dhall/Pretty/Internal.hs
+++ b/src/Dhall/Pretty/Internal.hs
@@ -50,13 +50,15 @@ data Ann
   | Label       -- ^ Record labels
   | Literal     -- ^ Literals such as integers and strings
   | Builtin     -- ^ Builtin types and values
+  | Operator    -- ^ Operators
 
 annToAnsiStyle :: Ann -> Terminal.AnsiStyle
-annToAnsiStyle Keyword  = Terminal.bold
-annToAnsiStyle Syntax   = mempty
-annToAnsiStyle Label    = Terminal.color Terminal.Green
-annToAnsiStyle Literal  = Terminal.color Terminal.Magenta
-annToAnsiStyle Builtin  = Terminal.color Terminal.Red
+annToAnsiStyle Keyword  = Terminal.colorDull Terminal.Green
+annToAnsiStyle Syntax   = Terminal.colorDull Terminal.Green
+annToAnsiStyle Label    = mempty
+annToAnsiStyle Literal  = Terminal.colorDull Terminal.Magenta
+annToAnsiStyle Builtin  = Terminal.underlined
+annToAnsiStyle Operator = Terminal.colorDull Terminal.Green
 
 prettyExpr :: Pretty a => Expr s a -> Doc Ann
 prettyExpr = prettyExprA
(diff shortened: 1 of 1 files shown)


ocharles:
All seems very reasonable to me! Ultimately I just want this as a tool for `dhall-repl` and `dhall` at the command line. For any stronger editing I would obviously be using my editor where I can control the colour scheme, so I don't mind what colours are chosen here.

> I classified True and False as builtin variables for the same reason: in a pure type
system are conceptually no different from Text, Type, or Kind: they are built-in
variables and the fact that they are on the term-level isn't a significant distinction.

From a technical level, maybe, but from a user level I view `True` and `False` to live in the same "world" as `42`, `3.14` and `"Hello!"`. That is - term level literals. Personally I thin kthey should be classified as literal rather than builtin. Your argument is equally valid though, so I'll be lazy and let you make the final call.

> The reason I did not use red anywhere was that I wanted to reserve that color for error
messages 

I consciously thought about this when I was writing the theme, so I'm surprised I even chose red. That was never intentional, and I agree that we shouldn't use red for anything other than failure.

PR #286, opened 2018-02-19
Enforce `-Werror` in CI

Gabriel439:
This adds the `-Werror` flag to CI so that warnings don't sneak into the
build (and also fixes the current set of warnings)


ocharles:
This enables it for more than just CI - it enables it for anyone who wants to build binaries locally. Is that what we want? I suppose we have the guarantee that anything on `master` won't have warnings.

Gabriel439:
I assume that people iterating quickly locally would be using `cabal` inside of `nix-shell`, in which case the flag wouldn't apply

ocharles:
I didn't mean for people doing development, I meant for people who don't necessarily trust binary distributions. But as I said, with this change it's impossible for them to be affected by `-Werror`, so I think my comment is a nonissue.

PR #293, opened 2018-02-21
Housekeeping

quchen:
- LTS-9/10 contain only Prettyprinter 1.1*, the required 1.2 bump will not happen until LTS-11. For this reason, Prettyprinter 1.2 is added explicitly to the stack.yaml file so that `stack build` works once again.
- Remove a couple of redundant dependencies (courtesy of [Weeder](https://hackage.haskell.org/package/weeder))


Gabriel439:
Thanks!  Also nice catch with Weeder

Issue #292, opened 2018-02-21
Support s3:// protocol with credentials discovery


imalsogreg:
We would like to be able to refer to dhall expressions by S3 address (e.g. `s3://bucket/path/to/file.dhall`), and to use compile-time credentials lookup from `~/.aws/credentials` or IAM environment variables. Given S3's authentication mechanism, getting files from s3 isn't as easy as just using an `http` endpoint with basic auth headers.

The goal is for some `foo.dhall` file checked in to github to be able import private data from `bar.dhall` stored on s3, without us including the s3 credentials in `foo.dhall`.

A couple possible ways to support this:

 - Add an `S3Url` constructor to `PathType`, and add support for parsing and importing values of this constructor. Use `amazonka`, for impure credentials discovery within `exprFromPath`
 - Parse `s3://` protocols but mangle them into `URL` values at compile time, somehow computing  the http signature and adding this to the URLs headers. No change to `exprFromPath` needed

I've started implementing the first option. But does it sound like I'm on the right track given my goals?

Gabriel439:
My main concern is that the S3 authentication mechanism is not based on an internet standard.  I would like to avoid baking in a company-specific protocol into the language and language standard

However, maybe there could be a way to add built-ins to Dhall so that the correct signing headers are computed for an ordinary HTTPS request with Dhall's support for custom headers

PR #294, opened 2018-02-22
WIP support s3:// url in dhall expressions

imalsogreg:
Addresses #292 

 - [x] Basic parsing of `s3://example/object.dhall` into an `S3URL` constructor of `PathType`
 - [x] Happy-path AWS-credentials discovery, download and parse of file at s3:// URL
 - [ ] Correct `canonicalize` for s3 paths
 - [ ] Better error-handling for malformed s3 paths and download failures
 - [ ] Code clean-up
 - [ ] Documentation
 - [ ] Tests
 - [ ] Possible replacement of `amazonka` (a heavy dependency) with something special-purpose

This simple case, with a `t.dh` dhall file in non-public `dhall-test` s3 bucket, is working:

```
-- local_file.dhall
let x = s3://dhall-test/t.dh
in  if x then 5 else 3
```

```
-- s3://dhall-test/t.dh
True
```

```
[nix-shell:~/code/dhall-haskell]$ dist/build/dhall/dhall <<< ./local_file.dhall
bkt
BucketName "ha2"
obj
ObjectKey "t.dh"
response:
True

Integer

5
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,4 +1,4 @@
-{ mkDerivation, ansi-terminal, ansi-wl-pprint, base
+{ mkDerivation, amazonka, amazonka-s3, ansi-terminal, ansi-wl-pprint, base
 , base16-bytestring, bytestring, case-insensitive, charset
 , containers, contravariant, cryptohash, deepseq, directory
 , exceptions, filepath, haskeline, http-client, http-client-tls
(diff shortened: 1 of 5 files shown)


imalsogreg:
Going to try to find a solution that isn't s3-specific

PR #306, opened 2018-02-28
Loosen upper bound on exceptions library

vmchale:
This would loosen the upper bound on the `exceptions` library, simplifying things for me downstream :)


Gabriel439:
Thanks! :)

PR #308, opened 2018-02-28
Fix test import

Gabriel439:
One of the tests had a remote import, which this changes to a relative
import so that the test suite won't fail if the network is down.


ocharles:
You could also mock out the transport layer and record/replay the HTTP session, so you know you've tested that code path too. A bit like https://github.com/vcr/vcr

Gabriel439:
Maybe, although I probably won't get around to that for a while, especially since the semantics of imports are not standardized yet

Issue #305, opened 2018-02-27
Dependency 'cryptohash' is deprecated


chris-martin:
[`cryptohash`](https://hackage.haskell.org/package/cryptohash) recommends not using this library anymore, and switching to `cryptonite`. Is there a particular reason dhall should be using this old package?

https://github.com/dhall-lang/dhall-haskell/blob/6b56aa5a0f8eea4189b408566b9dfc473c1a22ff/dhall.cabal#L105

Gabriel439:
There is no good reason why Dhall is using `cryptohash` (I don't even remember why I picked it).  I'm more than happy to switch to `cryptonite`

PR #313, opened 2018-02-28
Replace cryptohash with cryptonite (#305)

PierreR:
It is quite late for me. So here is the state of my attempt at fixing #305.

Tests are green but I would surprised if it is completely correct ?


Gabriel439:
Looks great!  Thank you :)

PR #328, opened 2018-03-17
Extensibility improvements

ramirez7:
### Summary
Some small changes to help with extensibility:
* `typeWithAN` & `judgmentallyEqualWith`: Allows custom normalization of `App` during type checking
* Some `Ord` instances
* [Record field types can now be of custom, `Embed`ded `Kind`s](https://github.com/dhall-lang/dhall-haskell/pull/328/files#diff-ade7a548af6e7fc4a77f18934c81c1ebR446)

None of these changes to the dhall-the-library should affect vanilla `dhall` (i.e. `Expr s X`)

### Motivation
I've been using these additions at work to do some stuff with an extended Dhall. Both are necessary to get it to work. Since I can't show you that, I wrote [a test suite with minimal examples of the functionality I'm using](https://github.com/ramirez7/dalek/blob/3f2d3670dfc4e9dc80630677ebd38c3b8486d86f/test/Dalek/UpstreamSpec.hs). Specifically...
* [`typeWithAN` allows generation of this union type from a list type literal](https://github.com/ramirez7/dalek/blob/3f2d3670dfc4e9dc80630677ebd38c3b8486d86f/test/Dalek/UpstreamSpec.hs#L50-L53)
* The Record field change allows me to embed [a type family on a record with fields of a closed kind](https://github.com/ramirez7/dalek/blob/3f2d3670dfc4e9dc80630677ebd38c3b8486d86f/test/Dalek/UpstreamSpec.hs#L80-L82)
* I had a change against an older version of `dhall-haskell` for the TypeLit stuff, but it looks like `judgmentallyEqual` now handles that fine :)

With these small extra features, I've found that I can write extended Dhall with very rich types that allow for an ergonomic but type-safe interface for the end-user.

Let me know if `UpstreamSpec` doesn't make the motivation for this PR clear and I can help explain or update it with more clear examples.

-------
For the `Ord` instances, what I really need is an `Ord` instance for `Expr`, but the InsOrd HashMap makes that difficult since it doesn't have an `Ord` instance. I'm just using [orphans w/`StandaloneDeriving`](https://github.com/ramirez7/dalek/blob/3f2d3670dfc4e9dc80630677ebd38c3b8486d86f/src/Dalek/Orphans.hs) in my project now and it's working fine, but eventually I'd like to get those instances unorphaned. The reason I need an `Ord` instance is because [in a `Map` extension, I'm using raw `Expr`s as the keys](https://github.com/ramirez7/dalek/blob/3f2d3670dfc4e9dc80630677ebd38c3b8486d86f/src/Dalek/Exts/Map/Core.hs#L33).

I'll probably cut a separate PR in the future to tackle `Ord (Expr s a)` (it will probably involve a `newtype` around the InsOrd HashMap)

### DIFF
--- a/src/Dhall/TypeCheck.hs
+++ b/src/Dhall/TypeCheck.hs
@@ -11,6 +11,7 @@ module Dhall.TypeCheck (
       typeWith
     , typeOf
     , typeWithA
+    , typeWithAN
     , checkContext
 
     -- * Types
@@ -32,7 +33,7 @@ import Data.Text.Lazy.Builder (Builder)
 import Data.Text.Prettyprint.Doc (Pretty(..))
 import Data.Traversable (forM)
 import Data.Typeable (Typeable)
-import Dhall.Core (Const(..), Chunks(..), Expr(..), Var(..))
+import Dhall.Core (Const(..), Chunks(..), Expr(..), Normalizer, Var(..))
 import Dhall.Context (Context)
 
 import qualified Data.Foldable
@@ -84,8 +85,21 @@ typeWithA
     -> Context (Expr s a)
     -> Expr s a
     -> Either (TypeError s a) (Expr s a)
-typeWithA tpa = loop
+typeWithA = typeWithAN (const Nothing)
+
+{-| Generalization of `typeWithA` that also allows custom normalization of types
+-}
+typeWithAN
+    :: Eq a
+    => Normalizer a
+    -> Typer a
+    -> Context (Expr s a)
+    -> Expr s a
+    -> Either (TypeError s a) (Expr s a)
+typeWithAN nrm tpa = loop
   where
+    normalizeTy = Dhall.Core.normalizeWith nrm
+    equalTy = Dhall.Core.judgmentallyEqualWith nrm
     loop _     (Const c         ) = do
         fmap Const (axiom c)
     loop ctx e@(Var (V x n)     ) = do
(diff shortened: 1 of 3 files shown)


Gabriel439:
Yeah, this looks like a good idea to me.  Just give me a couple of days to review this more closely

src/Dhall/TypeCheck.hs
```diff
@@ -425,17 +439,22 @@ typeWithA tpa = loop
...
     loop ctx e@(Record    kts   ) = do
         let process (k, t) = do
-                s <- fmap Dhall.Core.normalize (loop ctx t)
+                s <- fmap normalizeTy (loop ctx t)
                 case s of
                     Const Type -> return ()
                     Const Kind -> return ()
-                    _          -> Left (TypeError ctx e (InvalidFieldType k t))
+                    other -> loop ctx other >>= \x -> case x of
+                    -- This codepath doesn't come into play for vanilla Dhall,
```
Gabriel439: This code path does affect vanilla Dhall because it would make the following record type legal:

```haskell
{ x : List }
```

... where previously it was forbidden

Also, it seems odd that this extension is only done for record types and not elsewhere

src/Dhall/TypeCheck.hs
```diff
@@ -84,8 +85,21 @@ typeWithA
     -> Context (Expr s a)
     -> Expr s a
     -> Either (TypeError s a) (Expr s a)
-typeWithA tpa = loop
+typeWithA = typeWithAN (const Nothing)
+
+{-| Generalization of `typeWithA` that also allows custom normalization of types
+-}
+typeWithAN
```
Gabriel439: I'd suggest just making a breaking change to extend `typeWithA` to support a custom normalizer

ramirez7: 👍works for me

src/Dhall/TypeCheck.hs
ramirez7: Hmm that's true..

* I could make it this instead to help ensure vanilla Dhall isn't affected:
```haskell
other@(Embed _) -> loop ctx other -- etc
```
* Alternatively, this can remain and Vanilla Dhall can be affected slightly, but this newly-valid record type has no inhabitants anyways?

> Also, it seems odd that this extension is only done for record types and not elsewhere

Yeah I suppose it would make sense for Union types to have this extension as well - is that what you meant?

Gabriel439: Even if it has no inhabitants it should be rejected so that users get better error messages and don't get led astray

What I meant was that (A) it's not clear what this change has to do with this feature of customizing judgmental equality and (B) it's not clear why this special check is only done here instead of everywhere that the type-checker checks for a `Type` or `Kind`

Gabriel439:
Closing as this pull request is out of date.  Feel free to reopen after updating this to `master`

PR #329, opened 2018-03-18
Added an abbreviation for :save command.

aleator:
Example of use:

    ⊢ \(x:Natural) -> x +  +100
    λ(x : Natural) → x + +100

    ⊢ :save addhundred

    ⊢ ./addhundred +15
    +115


ocharles:
Hmm, this should already have been possible with :save foo = it. 'it' is a
special variable bound to the last expression evaluated.

aleator:
Yeah. I was playing with dhall repl in order to get it put stuff directly to ipfs and as practise decided to 'improve' ergonomics by shaving off the `= it` part. Ended up using it myself.

Gabriel439:
I'm mildly against this mainly because it doesn't save that many characters and is not very orthogonal to the existing `it` feature

PR #330, opened 2018-03-19
Switches from text-format to formatting

jkachmar:
[text-formatting](https://github.com/bos/text-format) appears to be unmaintained (last commit was 2015, several PRs are sitting unmerged including one that unblocks GHC 8.4). Fortunately it looks like [formatting](https://github.com/chrisdone/formatting) is a drop-in replacement for it.

I've bumped `stack.yaml` to LTS 11.1 to pull in `formatting-6.3.1` and `prettyprinter-1.2.0.1`, and verified that the project builds and testa pass. I also verified a successful build and test run against `nightly-2018-03-18`

Note that I can't seem to get `cabal2nix` to work, so I hand-edited the `default.nix` file to remove `text-format` and add `formatting`, but I'm not entirely sure if that's kosher.


Gabriel439:
Thanks for doing this!  I fix the build errors, so don't worry

jkachmar:
Whoops, didn't see your comment! I didn't realize that LTS 9.6 -> LTS 11.1 would bump GHC so I reverted it and added an extra dep for `formatting-6.3.1`, and ran the tests a few times to get an approximation of what kind of performance effects this would have (negligible, something like 1.53s vs 1.49s after 3 rounds of testing each).

Gabriel439:
Just merge https://github.com/jkachmar/dhall-haskell/pull/1/files into your branch and then CI should pass

Gabriel439:
Awesome!  Thanks for fixing this :)

PR #332, opened 2018-03-19
A command to interact with local ipfs installation

aleator:
Since dhall has some affinity with ipfs I added a dhall-repl
command to directly push expressions to ipfs.

Here is an example of use:

    ⊢ :ipfs h = (\(x:Text) -> x++x)

    Added with hash: Qmee1h5H7o8Lp2FQtRJAPNGgLKRdrUSGn5jPkW6NBBxxh8
    ⊢ h
    "Qmee1h5H7o8Lp2FQtRJAPNGgLKRdrUSGn5jPkW6NBBxxh8"

Then on a completely different machine:

    ⊢ https://ipfs.io/ipfs/Qmee1h5H7o8Lp2FQtRJAPNGgLKRdrUSGn5jPkW6NBBxxh8 "hello "
    "hello hello "
    ⊢

I find this command delightful, but I won't be offended if this doesn't get
merged.

This incurs dependencies on typed-process, directory and deepseq. Also,
it needs -threaded to work properly.


Gabriel439:
I would like to hold off on this, only because I have another plan in mind for sharing Dhall expressions in the long term

Dhall's support for semantic integrity checks is similar in spirit to IPFS hashes but with the advantage that they are Dhall-aware, so one of my long term goals is to make Dhall expressions globally content-addressable by their semantic hash.  Doing so would provide a few advantages over IPFS:

* It would guarantee that an expression addressed in this way has a fixed meaning (unlike an IPFS import, which can have impure references to other imports)
* It would make the hash insensitive to behavior-preserving changes

So the idea is that at some point you could take an import like:

```haskell
./foo sha256:f0e4c2f76c58916ec258f246851bea091d14d4247a2fc3e18694461b1816e13b
```

... and just drop the `./foo` so that `sha256:f0e4c2f76c58916ec258f246851bea091d14d4247a2fc3e18694461b1816e13b` was a valid import all by itself.  Then you could compute the hash of any expression and share the hash with another person so that they can import your expression.  This implies building infrastructure (similar to IPFS) for caching and distributing this content-addressable "Dhall store".

aleator:
Sure thing. I'm interested to hear more. Are you intending on building a centralised service? Updated scale distributed might require rather large userbase.

PR #338, opened 2018-03-27
Add RecordType, record and field

ocharles:
I bike-shedded the naming a little bit, and think RecordType, record and field might be better names than the previous "RecordBuilder", "makeRecord" and "fieldValue" names.


ocharles:
How do you want to resolve the Hydra error? It's probably best to turn off `DeriveAnyClass` here.

Gabriel439:
Yeah, it seems like `DeriveAnyClass` and `DeriveGeneric` are no longer used within that module

ocharles:
Ok, I believe this is now ready to merge.

Gabriel439:
Looks great!  Thanks :)

PR #340, opened 2018-03-28
Support for ghc-7.10.3 / lts-6.27

jneira:
Hi, in the process of integrate [the use of dhall](https://github.com/typelead/eta/issues/704) with [eta](https://github.com/typelead/eta) we have needed to compile dhall with the version of ghc used to compile eta itself. In this way we can use dhall to configure eta and to configure the packages compiled with it.
Maybe a version compilable with ghc-7.10.3 could be useful for other users so we open the pr.
Fortunately all dependencies are in a suitable range and only there is a type application that needs ghc 8.
There is the option to add conditions on ghc version in the cabal file and modules but maybe it supposes too much burden. Another one would be simply create a dedicated branch. Let me know which one do you  think is better.

### DIFF
--- a/dhall.cabal
+++ b/dhall.cabal
@@ -213,20 +213,20 @@ Executable dhall
     Other-Modules:
         Paths_dhall
 
--- Executable dhall-repl
---     Hs-Source-Dirs: dhall-repl
---     Main-Is: Main.hs
---     Build-Depends:
---         base             >= 4        && < 5   ,
---         ansi-terminal                         ,
---         dhall                                 ,
---         haskeline        >= 0.7.3.0  && < 0.8 ,
---         mtl              >= 2.2.1    && < 2.3 ,
---         repline          >= 0.1.6.0  && < 0.2 ,
---         prettyprinter                         ,
---         prettyprinter-ansi-terminal           ,
---         text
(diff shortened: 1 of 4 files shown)


dhall.cabal
```diff
@@ -178,7 +178,9 @@ Library
         text                        >= 0.11.1.0 && < 1.3 ,
         transformers                >= 0.2.0.0  && < 0.6 ,
         unordered-containers        >= 0.1.3.0  && < 0.3 ,
-        vector                      >= 0.11.0.0 && < 0.13
+        vector                      >= 0.11.0.0 && < 0.13,
+        semigroups                  >= 0.8.3.1  && < 1
```
Gabriel439: Actually, one more request: could you change this line to:

```
if !impl(ghc >= 8.0)
  build-depends: semigroups == 0.18.*
```

See: https://prime.haskell.org/wiki/Libraries/Proposals/SemigroupMonoid#RecommendedVariant

jneira:
@Gabriel439 hi, besides the requested changes, i had to add `megaparsec` and `parser-combinators` in `stack.yaml` to make stack build work.

Gabriel439:
Thanks!  Looks great to me :)

PR #344, opened 2018-03-31
Change `:type` to only print type in `dhall-repl`

Gabriel439:


ocharles:
Hm, I modelled this on the behavior of GHCI (which echoes the expression), but not necessarily against this change.

Gabriel439:
Yeah, the main reason I did it was because I was having difficulty visually picking out the type annotation.  For example:

```
⊢ :type https://ipfs.io/ipfs/QmdtKd5Q7tebdo6rXfZed4kN6DXmErRQHJ4PsNCtca9GbB/Prelude/List/replicate
      let replicate
          : Natural → ∀(a : Type) → a → List a
          =   λ(n : Natural)
            → λ(a : Type)
            → λ(x : a)
            → List/build
              a
              (   λ(list : Type)
                → λ(cons : a → list → list)
                → Natural/fold n list (cons x)
              )
  
  in  replicate
: ∀(n : Natural) → ∀(a : Type) → ∀(x : a) → List a
```

Part of the reason is that Dhall's type-level expressions are not as visually distinctive from term-level expressions as in Haskell since tin Dhall both terms and types use very similar punctuation

Issue #346, opened 2018-04-01
Sum types produce empty backtick quoting


srid:
This looks like a bug to me. Dhall's `embed . inject` when applied a sum type like the following:

```
data Amount
  = Debit Scientific
  | Credit Scientific
  deriving (Generic, Show)
```

produces this:

```
amount = < Debit = { `` = 5.45 } > 
```

I don't think this right, because those backtick pairs seem to be invalid Dhall syntax. dhall-format doesn't recognize it:

```
Error: Invalid input

(stdin):1:109: error: unspecified
    error
<y = "Internet", amount = < Debit = { `` = 5.45 } > }, { transactiondate = "02>
```

I'm using dhall  at rev 7d7b8590b5ef82598bbf36e146a0c4c20e4dec04

srid:
Looking at Dhall's union type representation, it should have produced `< Debit = 5.45 >`.

Gabriel439:
Actually, what it should have produced was this:

```haskell
< Debit = 5.45 | Credit : Double >
```

There are three issues to address here:

*   The `Credit` alternative was missing from the union

    I'm pretty sure I've found out the cause of this and can fix it

*   The `5.45` is being wrapped in a record

    It should be just `5.45`, not a record with one field whose value is `5.45`

    I'm still looking into what caused this

*   The record has a field with the wrong name

    I believe this is might be because you are using a non-standard `constructorModifier` in the `InjectOptions` which is causing this because I get a field named `_1` when I try to reproduce this:

    ```haskell
    $ git checkout 7d7b8590b5ef82598bbf36e146a0c4c20e4dec04
    $ cabal repl lib:dhall
    >>> :set -XDeriveGeneric
    >>> :set -XDeriveAnyClass
    >>> data Amount = Debit Scientific | Credit Scientific deriving (Show, Generic, Inject, Interpret)
    >>> Dhall.Core.pretty (embed inject (Debit 5.45))
    "< Debit = { _1 = 5.45 } >"
    ```

    `_1` is the expected name of this field (although there shouldn't be a record at all if this were working correctly)

Gabriel439:
Actually, I forgot that the expected result is actually:

```haskell
< Debit = { _1 = 5.45 } | Credit : { _1 : Double } >
```

This is because the Dhall encoding of Haskell's sum types is supposed to wrap the contents of each constructor in an obligatory record.  So there are actually only two issues here:

* The missing `Credit` alternative

[...]

PR #348, opened 2018-04-01
Fix derived `Inject` instance for sum types

Gabriel439:
Related to #346

This fixes the `GenericInject` instance for a sum type with two
constructors to include the type of the alternative constructor.  For
example, before this change you would get the following incorrect
conversion:

```
$ cabal repl lib:dhall
>>> :set -XDeriveGeneric
>>> :set -XDeriveAnyClass
>>> data Amount = Debit Scientific | Credit Scientific deriving (Show, Generic, Inject, Interpret)
>>> Dhall.Core.pretty (embed inject (Debit 5.45))
"< Debit = { _1 = 5.45 } >"
```

... which is missing the `Credit` alternative.

After this change you get the correct result:

```
< Debit = { _1 = 5.45 } | Credit : { _1 : Double } >
```

### DIFF
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -939,10 +939,10 @@ instance (Constructor c1, Constructor c2, GenericInject f1, GenericInject f2) =>
     genericInjectWith options@(InterpretOptions {..}) = pure (InputType {..})
       where
         embed (L1 (M1 l)) =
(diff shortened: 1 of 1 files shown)


srid:
I've verified that this fixes it!

Getting ` < Debit = { _1 = 4.88 } | Credit : { _1 : Double } >`

Gabriel439:
Awesome!  Thanks for verifying that :)

PR #371, opened 2018-04-28
Introduce normalizeWithM for monadic normalization

alexbiehl:
See the discussion in https://github.com/dhall-lang/dhall-lang/issues/133.

### DIFF
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -152,7 +152,7 @@ input
     -> IO a
     -- ^ The decoded value in Haskell
 input ty txt =
-  inputWith ty Dhall.Context.empty (const Nothing) txt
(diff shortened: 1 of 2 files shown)


alexbiehl:
- [ ] This needs a big comment saying  that this might violate assumptions

Gabriel439:
Don't forget to export `NormalizerM` and `normalizeWithM` from the module

Gabriel439:
I'll close this for now since from offline discussions it sounded like you could do this using the existing API by post-processing the abstract syntax tree to monadically normalize uninterpreted free variables

Gabriel439:
@ocharles: Yeah, it seems reasonable to proceed with this change

ocharles:
Ok, I think this is good for review.

PR #377, opened 2018-05-10
Set minor version of insert-ordered-containers dependency to 0.2.1.0

jneira:
Cause dhall-haskell uses [`HashMap.foldMapWithKey`](https://github.com/dhall-lang/dhall-haskell/blob/master/src/Dhall/Diff.hs#L269) that is not included in previous versions


Gabriel439:
Thanks for catching that!

Gabriel439:
Whoops, looks like my Hydra machine ran out of disk space.  I will have to fix it later tonight before I can merge this

PR #381, opened 2018-05-12
Switch grammar of `Natural` and `Integer`

Gabriel439:
... as standardized in https://github.com/dhall-lang/dhall-lang/pull/141

Now `Natural`s no longer require a leading `+` and non-negative `Integer`s
require a leading `+`


Gabriel439:
Good catch!  I will fix those and also see if I can add doctests to check those

PR #385, opened 2018-05-13
Include path to the failed import in `HashMismatch` errors

### LINKED ISSUE #383: `HashMismatch` does not mention import
When you're dealing with integrity failures, the message only shows the mismatching hashes. https://github.com/dhall-lang/dhall-haskell/blob/2e631db47ac2979842ce95fee439cd9a8aed2e38/src/Dhall/Import.hs#L541-L552

If you have multiple integrity checks, it's pretty hard to know which one of them failed. How would you feel about also showing the import that failed in that message? We ought to have all of the information when we notice the failure, would have to thread it through to `HashMismatch`.

If you're up for it, what copy would you want the new message to have?

Gabriel439:
Fixes #383

Dhall has a standard way to report where hash mismatch failure occur but
I forgot to use this reporting system when implementing the hash
mismatch failure.

For example, using this file:

```bash
$ cat example.dhall
./Prelude/List/replicate sha256:b0e3ec1797b32c80c0bcb7e8254b08c7e9e35e75e6b410c7ac21477ab90167ae
```

... the error message you would get before this change was:

```bash
$ dhall <<< './example.dhall'

Error: Import integrity check failed

Expected hash:

↳ b0e3ec1797b32c80c0bcb7e8254b08c7e9e35e75e6b410c7ac21477ab90167ae

Actual hash:

↳ b0e3ec1797b32c80c0bcb7e8254b08c7e9e35e75e6b410c7ac21477ab90167ad
```

... and the error message you get after this change is:

```bash
$ dist/build/dhall/dhall <<< './example.dhall'

↳ ./example.dhall
  ↳ ./Prelude/List/replicate sha256:b0e3ec1797b32c80c0bcb7e8254b08c7e9e35e75e6b410c7ac21477ab90167ae

Error: Import integrity check failed

Expected hash:

↳ b0e3ec1797b32c80c0bcb7e8254b08c7e9e35e75e6b410c7ac21477ab90167ae

Actual hash:

↳ b0e3ec1797b32c80c0bcb7e8254b08c7e9e35e75e6b410c7ac21477ab90167ad
```

### DIFF
--- a/src/Dhall/Import.hs
+++ b/src/Dhall/Import.hs
@@ -806,7 +806,7 @@ loadStaticWith from_import ctx import_ = do
             let actualHash = hashExpression expr
             if expectedHash == actualHash
                 then return ()
-                else throwM (HashMismatch {..})
+                else throwM (Imported (import_:imports) (HashMismatch {..}))
 
     return expr
 


Gabriel439:
You're welcome! :)

Issue #362, opened 2018-04-19
Improve performance by caching


aleator:
I have a rather large dhall configuration and it is really taking far too long to process. I believe an order of magnitude improvement could be had just by caching the imports. 

```
> echo "./MySetup" | fsatrace r trace -- dhall
... Output clipped
> wc -l trace
550
>sort trace|uniq -c |wc -l
37
> dhall version
1.12.0
```

For the record, there are 68 files in my directory, most of them small dhall files that are eventually collected in the huge list above.

I thought that dhall already had some caching in place?

Gabriel439:
Dhall does have caching in place, so this seems like a bug.  Let me see if I can reproduce this issue locally

aleator:
Let me know if you want the blob of files I used.

Gabriel439:
@aleator: I think the main thing that would help if you have time is a minimal reproducing example of a file being imported twice

aleator:
I'll try to cook one up, but I probably won't manage to have time until next week.

Gabriel439:
Also, it might not be necessary.  I'm in the process of formalizing the import semantics and updating the Haskell implementation to match the standard semantics so the problem might disappear after that.  I'll let you know when that is complete so that you can check again if the problem still persists.

sellout:
I haven’t yet been able to run fsatrace (on a Mac), but if this is the case, it would explain some of the slow performance we’ve been seeing. We have a couple files that take a few seconds to normalize on their own, but when referenced by other files (that reference those slower files indirectly through a few different imports), it extends to minutes.

sellout:
I’m digging into this now. I’ve identified two bugs and fixed one of them so far, but I feel like there is a third one lingering. Should have a PR soon.

sellout:
Ok, so the second bug that I identified is maybe a bit tricky. `FilePath.normalize` doesn’t eliminate `..` from paths, so you can end up importing the same file multiple times based on where you imported it _from_. `System.Directory.canonicalizePath` fixes this, but it actually looks at the filesystem, so uses `IO`. We could simply remove `/[^/]+/\.\.` (i.e., the `..` and the next directory up), but that wouldn’t behave properly around symlinks (which is why `canonicalizePath` uses `IO`.

In my simple testing, I had a dhall file trigger 532 imports. Fixing the first bug (imports containing imports don’t get cached) brought that down to 240, and if I manually canonicalize `..`, I get down to 82.

[...]

PR #388, opened 2018-05-15
Cache imports that contain imports.

sellout:
Relevant to #362.

Previously, if an import contained additional imports, it wouldn’t get
cached. This fixes that, with a couple related changes:
- type checking happens in the failed-lookup branch, rather than having to
return a `cached` boolean,
- the normalized form of the import is used more consistently (although this
doesn’t fix the related normalization problem around `../`).


sellout:
BTW, my branch is a few commits behind master, because I didn’t want to modify all my configs with the Integer/Natural change just to experiment with this.

PR #389, opened 2018-05-15
Add non-Haskell dependencies to stack.yaml.

sellout:
Not sure if this is a change you’re interested in. IMO, it’s nice for stack
users even if they don’t use Nix, since it makes the dependencies explicit.


Gabriel439:
Yeah, I'm fine with making changes to support `stack`

PR #390, opened 2018-05-16
Update import logic to more closely match standard

Gabriel439:
This is one step in a series of commits to update the Haskell
implementation to match the import semantics standardized in:

https://github.com/dhall-lang/dhall-lang/pull/127

This is a BREAKING CHANGE mostly due to restricting the grammar for
valid file paths and URLs.  However, in practice most existing code
should still work after these changes.

Specifically, this change:

* Updates the grammar for file paths and URLs
* Simplifies and cleans up the semantics for canonicalizing paths (and
  chains of paths)

The main difference in the grammar is that file paths and URL paths are
unified to have the same syntax and they are both more restricted than
before.

For example, the following URLs are no longer valid:

* `http://example.com` (missing a path to a file)
* `http://example.com/()` (invalid path characters)

... and the following paths are no longer valid:

* `/?#` (invalid path characters)

Note that this actually deviates from the standard in one way: file
paths no longer permit `?` and `#` characters (whereas the standard
allows them).  However, in the course of testing these changes I
realized that if file paths permit `?` and `#` characters and you unify
URLs to also permit them then you get an ambiguous grammar where a `?`
or `#` in a URL could be interpreted as either part of a path component
or part of a query parameter or fragment.

The semantics for canonicalizing paths is a straightforward translation
of the standard to Haskell code.

### DIFF
--- a/src/Dhall/Core.hs
+++ b/src/Dhall/Core.hs
@@ -18,7 +18,9 @@
 module Dhall.Core (
     -- * Syntax
       Const(..)
-    , HasHome(..)
+    , Directory(..)
+    , File(..)
+    , FilePrefix(..)
     , Import(..)
     , ImportHashed(..)
     , ImportMode(..)
@@ -80,9 +82,8 @@ import qualified Data.HashSet
 import qualified Data.Sequence
 import qualified Data.Set
 import qualified Data.Text
-import qualified Data.Text.Lazy                        as Text
-import qualified Data.Text.Lazy.Builder                as Builder
-import qualified Data.Text.Prettyprint.Doc             as Pretty
+import qualified Data.Text.Lazy.Builder     as Builder
+import qualified Data.Text.Prettyprint.Doc  as Pretty
 
 {-| Constants for a pure type system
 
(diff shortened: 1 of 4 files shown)


Gabriel439:
cc: @sellout 

This pull request cleans up and simplifies the canonicalization logic, so if you are dealing with misbehaving canonicalization then try out this code

Gabriel439:
Some other comments:

* This is also a breaking change to the API (by changing the `ImportType` structure significantly)
* I plan to upstream the fix to file paths to the standard to disallow `?` and `#`
* I also did an ad-hoc test to successfully import https://raw.githubusercontent.com/dhall-lang/dhall-to-cabal/master/dhall/prelude.dhall, which is a fairly involved import chain

sellout:
This looks good. Using my same example from https://github.com/dhall-lang/dhall-haskell/issues/362#issuecomment-389326485, I get 181 imports loaded. using `sort | uniq`, I get down to 70, so there’s still some duplication somewhere, but this is still an improvement over the 240 imports after #388 (and from the 532 imports like 48 hours ago).

PR #392, opened 2018-05-16
Normalize each import before it is cached.

sellout:
In tests on Dhall configs we use at work, I saw speedups between 33–90% on
extremely complicated configs with this change.

It mostly just adds a line calling `normalize`, but also threads the Normalizer
through a bunch of places to make sure we normalize properly for the context.

This is based on #390, because this optimization is likely less effective if
we’re doing more imports.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -30,6 +30,12 @@ let
   pkgs =
     import nixpkgs { inherit config; };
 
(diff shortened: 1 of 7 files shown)


Gabriel439:
Thanks! :)

PR #393, opened 2018-05-17
Add a `pwd` derivation force status updates by Hydra

Gabriel439:
Hydra doesn't update a pull request if the derivation hasn't changed
since the last revision (i.e. a cache hit).  This change adds a `pwd`
derivation which depends trivially on the current directory to force a
cache miss and status update

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -30,6 +30,12 @@ let
   pkgs =
     import nixpkgs { inherit config; };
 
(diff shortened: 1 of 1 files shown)


sellout:
I assume this is to fix the lack of a build triggered by the merge on #392?

Gabriel439:
Yep, that is exactly the reason :)

PR #396, opened 2018-05-20
Depth-first import algorithm

Gabriel439:
The standard specifies a depth-first import algorithm for resolving
imports but the Haskell implementation was using a breadth-first import
algorithm, which this change fixes

### DIFF
--- a/src/Dhall/Import.hs
+++ b/src/Dhall/Import.hs
@@ -551,11 +551,14 @@ exprFromImport (Import {..}) = do
             request <- liftIO (HTTP.parseUrlThrow url)
 
             requestWithHeaders <- case maybeHeaders of
-                Nothing           -> return request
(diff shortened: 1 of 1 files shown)


Gabriel439:
@sellout: It's simpler to standardize and simpler to implement

As far as I know it should be equivalent in behavior assuming that the order in which you import things doesn't matter

ocharles:
I think @sellout's point is that a special shouldn't be mandating a
particular implementation. If anything, it can provide a model or reference
to compare against.

Gabriel439:
@ocharles: The standard does not mandate a particular implementation.  The main reason I'm changing the Haskell implementation to match is because it's simpler and can be used as a reference implementation for how to translate the standard to code

However, the standard has to specify *an* implementation so that people have some precise starting point

PR #401, opened 2018-05-22
Fix haddocks for `Directory` type

### LINKED ISSUE #399: Bad docs for Directory
There is currently a haddock mistake in the docs for `Directory`.

Gabriel439:
Fixes #399


andrewthad:
I'm not terribly familiar with haddock. Does the first forward slash not need to be escaped as well?

Gabriel439:
@andrewthad: It's technically not necessary to escape the first slash but I'll do it just for consistency

PR #404, opened 2018-05-24
dhall-format expand --inplace option for directories

FintanH:
The purpose of this PR is so that if you have your config files gathered in a directory you can apply a directory wide format, instead of going to each file individually.

**Changes:**

* Created `InplaceOption` to distinguish between formatting in place for a single `File` or a whole `Directory` with `Ext`ensions the user can target.
* Updated CLI parser
* Moved the file formatting into a function `inplaceFile`
* Created `inplaceDirectory` which is recursive and utilises `inplaceFile`

Hopefully this useful to the toolchain :)

### DIFF
--- a/dhall-format/Main.hs
+++ b/dhall-format/Main.hs
@@ -24,12 +24,12 @@ module Main where
 
 import Control.Applicative (optional)
 import Control.Exception (SomeException)
-import Control.Monad (when)
+import Control.Monad (when, forM_)
 import Data.Monoid ((<>))
 import Data.Version (showVersion)
 import Dhall.Parser (exprAndHeaderFromText)
 import Dhall.Pretty (annToAnsiStyle, prettyExpr)
-import Options.Applicative (Parser, ParserInfo)
+import Options.Applicative (Parser, ParserInfo, (<|>))
 import System.IO (stderr)
 import System.Exit (exitFailure, exitSuccess)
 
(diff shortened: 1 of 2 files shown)


FintanH:
Great points, thanks! I guess my Unix-fu is pretty weak :)

The aim of this was to be able to have directory wide commands that can format and compile all files that we work on. Then use these in our CI to make sure that:
a) all dhall files are correct 
b) all dhall files are formatted

I'll tweak the `find` example you supplied for these purposes 👍

Gabriel439:
Sounds good!  I'll leave this open for a while in case other people want to comment

FintanH:
fwiw this is what myself and @sellout have been coming up with:
`find dhall-dir -type f \( -not -iname *.md \) -exec dhall-format --inplace {} \;`
and
`find dhall-dir -type f \( -not -iname *.md \) -exec sh -c 'dhall <<< ./{} >/dev/null 2>&1; echo $?' \; | sort | uniq`

The `-not -iname *.md` is basically to exclude our README files and everything else is dhall.

The `>/dev/null 2>&1;` is to hide type outputs, `echo $?` to see if any of them failed and the rest to output the bad exit code if that happened.

Gabriel439:
@FintanH: Note that the latest version of the `dhall` executable provides a `dhall type` subcommand you can use if all you want to do is type-check the expression

FintanH:
@Gabriel439 But imports are disabled with type checking correct?

Gabriel439:
@FintanH: Oh yeah, but you could do `dhall import | dhall type` to do just the import and type-checking stages

FintanH:
@Gabriel439: Does that improve on just using straight up `dhall`?

Gabriel439:
@FintanH: Not that much.  It's mainly so that you don't have to normalize the file

ocharles:
Rather than having `dhall-format` format a directory, have you considered having it format all references that it has write access to? That could also be interesting.

FintanH:
@ocharles That was also something that came to mind. But for us, all our dhall files are under certain directories so the only requirements were to avoid `md` files and not destroy the git repo by recursively going into `.git` 😅 The `find` solution was the easiest solution really.

I am keeping a version of a script that has the `dhall` command going through a directory because the combination of seeing good error messages and getting a failing exit code was needed, which I couldn't seem to achieve with my limited bash knowledge.

Gabriel439:
@ocharles: That is yet another reason to use `find` to drive this because `find` has support for filtering files based on permissions using the `-perm` flag

ocharles:
What I suggest is less about permissions, and more about only formatting files that are reachable from an expression. But I don't feel at all strongly about it (I'm fine with `find`, though my editor just does it on save), just throwing another idea out there.

ocharles:
Sounds like we're all in agreement to use `find` for this. Time to close this one?

FintanH:
Ya sounds good

PR #407, opened 2018-05-27
Remove the Parent constructor from FilePrefix.

### LINKED ISSUE #406: Odd representation of leading .. in import paths
To represent an import path like `../foo/bar` using the API, you need to write `Local Parent (File (Directory ["foo"]) "bar")`. But importing `../../foo/bar` needs `Local Parent (File (Directory ["foo", ".."]) "bar")`; the first `..` gets special treatment. And importing `./../../foo/bar`, which is equivalent, could be represented as `Local Here (File (Directory ["foo", "..", ".."]) "bar")`.

I think it would be simpler and more uniform to just drop the `Parent` constructor from `FilePrefix`, and put `..` segments in the `Directory`. Alternatively, if it's desired to keep `..` out of the `Directory`, then the `Here` constructor could take a `Natural` saying how many directories to go upwards, and then the documentation will need to state that `..` segments shouldn't be present in a `Directory`.

quasicomputational:
The 'Parent' constructor is entirely subsumed by 'Here'; having it
only leads to redundant representations.

Closes #406.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+1.15.0
+
+* BREAKING CHANGE TO THE API: Removed the `Parent` constructor from `FilePrefix`.
+    * Instead, use `Here` with a `".."` prefix.
+
 1.14.0
 
 * BREAKING CHANGE TO THE LANGUAGE: Switch grammar of `Natural` and `Integer`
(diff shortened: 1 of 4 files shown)


quasicomputational:
This is the first option from #406, where we just delete `Parent` and have `..` components in the `Directory`.

PR #418, opened 2018-06-01
Data instance for Expr.

### LINKED ISSUE #416: `Data` instance for `Expr Src X`
At work, many of our `Interpret` instances import Dhall files for the `expected` component of the `Type`, to avoid duplicating the types and ensure that our configs stay in sync with the interpreter.

In order to avoid a runtime dependency on the location of the configs (especially remote URLs), we use TemplateHaskell to resolve all the imports at compile time. However, since there’s no `Data` instance for `Expr`, we pretty-print the resolved AST and then re-interpret it at runtime (assuming that it has type checked and has no imports).

Having a `Data` instance should help us avoid the re-serialization / reparsing steps.

If this is deemed acceptable, I’m happy to take it on myself.

As an aside, I think the TH functions we have would make a good addition to the dhall-haskell library.

sellout:
Fixes #416.

Also adds some Template Haskell to be able to load & resolve a Dhall expression
at compile-time.

Without the `Data` instance, the AST had to be pretty-printed back to a `String`
in the Template Haskell, and in 1.14 imports now auto-normalize, which often
results in much larger ASTs, and therefore (at least with the current version)
_much_ longer parses after re-serialization. Adding the `Data` instance avoids
that.

The current implementation of the `Data` instance removes
`Data.Text.Lazy.Builder` in favor of using `Data.Text.Lazy` directly, this is
simply because the latter already has a `Data` instance. It’s probably
reasonable to instead have a non-derived `Data` instance for `Chunks` that
explicitly uses `to`/`fromLazyText`, but that approach was less mechanical, so I
punted on it.


sellout:
Another thing I’d like to do with this is somehow have `staticLoadWith` be able to call `addDependentFile` on each import (at least the file imports), so that code using `staticDhallExpression` is recompiled if any of the imports are modified.

dhall.cabal
```diff
@@ -176,6 +176,7 @@ Library
         prettyprinter               >= 1.2.0.1  && < 1.3 ,
         prettyprinter-ansi-terminal >= 1.1.1    && < 1.2 ,
         scientific                  >= 0.3.0.0  && < 0.4 ,
+        template-haskell                                 ,
```
Gabriel439: Perhaps add an upper bound to `template-haskell`

src/Dhall/TH.hs
```diff
@@ -0,0 +1,28 @@
...
+import           Language.Haskell.TH.Syntax
+
+-- A workaround for a problem in TemplateHaskell (see
+-- https://stackoverflow.com/questions/38143464/cant-find-inerface-file-declaration-for-variable)
+liftText :: Text.Text -> Q Exp
+liftText = fmap (AppE (VarE 'Text.pack)) . lift . Text.unpack
+
+importDhallExpression :: Text.Text -> IO (D.Expr D.Src D.X)
+importDhallExpression =
+  either (const (fail "invalid Dhall expression")) D.load
```
Gabriel439: Perhaps include the name `Dhall.TH.importDhallExpression` in the error string so that it's easier for people to hunt down the source for where the error originates from

src/Dhall/TH.hs
```diff
@@ -0,0 +1,28 @@
+{-# LANGUAGE TemplateHaskell #-}
+
```
Gabriel439: Could you add a module header explaining an example use case for this functionality?

src/Dhall.hs
```diff
@@ -190,6 +192,32 @@ inputWith (Type {..}) ctx n txt = do
         Just x  -> return x
         Nothing -> Control.Exception.throwIO InvalidType
 
+{-| Similar to `input`, but without interpreting the Dhall `Expr` into a Haskell
+    type.
+-}
+inputExpr
```
Gabriel439: Very handy!  Thanks for adding this :)

PR #427, opened 2018-06-04
Documenting exprFromText

FintanH:
`exprFromText`'s inputs were confusing, didn't know what `String` was. Documented inputs.

P.S. Hello from LambdaConf, see you at ZuriHac

### DIFF
--- a/src/Dhall/Parser.hs
+++ b/src/Dhall/Parser.hs
@@ -1595,7 +1595,10 @@ instance Show ParseError where
 instance Exception ParseError
 
 -- | Parse an expression from `Text` containing a Dhall program
-exprFromText :: String -> Text -> Either ParseError (Expr Src Import)
+exprFromText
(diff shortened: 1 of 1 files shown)


sellout:
“Name of source file” is helpful – I’m assuming that’s how the parameter to Trifecta is described? But
1. all of its uses in Dhall (I think) don’t actually use a file, but some other description; and
2. I _think_ we want to encourage users to deal with expressions (which may be simply an import) rather than actual files.
So, maybe there’s a different way to phrase it.

Also, did you live-code this during your talk? 😂

Gabriel439:
Maybe call the first parameter "User-friendly name describing the input expression, used in parsing error messages"

FintanH:
@sellout @Gabriel439 You guys happy with this? 

And yes, I did live PR it 😛 Be sure to check out the video when it's out and I ask Matt Parsons for $5 😂

PR #434, opened 2018-06-07
Add `Integer/toDouble`

Gabriel439:
... as standardized in https://github.com/dhall-lang/dhall-lang/pull/168 assuming that proposal is merged without objection


Prelude/Natural/toDouble
```diff
@@ -0,0 +1,16 @@
+{-
+Convert a `Natural` number to the corresponding `Double`
+
+Examples:
+
+```
+./toInteger 3 = 3.0
```
joneshf: Should this be `toDouble`?

Gabriel439: Whoops!  Yes, it should be :)

PR #439, opened 2018-06-07
Remove unnecessary normalization

foollbar:
Based on the documentation of `hashExpressionToCode`, normalization in advance is not necessary.


Gabriel439:
Yep!  That is correct :)

Thank you for fixing that

Issue #443, opened 2018-06-08
Too much memory use in compiling Dhall.Parser


wavewave:
On my laptop with 8GB memory, it took almost 100% memory to compile Dhall library, especially on Dhall.Parser. If we can reduce GHC memory usage by simple module splitting or code reorganization, it's much desirable.

Gabriel439:
@wavewave: Yeah, I can do that

Just out of curiosity, what version of GHC are you using to build the `dhall` package?

wavewave:
I use GHC 8.2.2.

FintanH:
Fwiw: This occurred once we clamped to `2f3934ddd3bbffa5da778ef106680d9f122a16a2` updating from `4a085aa3d622886cf7dd96a1ad475ba914d5ab1f`.

f-f:
Now that #448 went in, compilation performance should be more acceptable.
@wavewave can you try taking a look at this and see if it's good?

FintanH:
@f-f We would like to test this but to update past the sha we're using we would need #453 to be fixed.

Gabriel439:
@FintanH: Yeah, #453 close to being fixed.  ETA is about 1-2 days

FintanH:
Thanks @Gabriel439! Very much appreciated :)

Gabriel439:
I have a pull request up with the fix now: https://github.com/dhall-lang/dhall-haskell/pull/471

You should be able to test memory usage off of that branch

FintanH:
Unfortunately we're still seeing the same issue.

```
[2018-06-18T16:57:55Z]     Configuring dhall-1.14.0...
[2018-06-18T16:57:55Z]     Preprocessing library for dhall-1.14.0..
[2018-06-18T16:57:55Z]     Building library for dhall-1.14.0..
[2018-06-18T16:57:55Z]     [ 1 of 13] Compiling Dhall.Context    ( src/Dhall/Context.hs, .stack-work/dist/x86_64-linux-nix-9ab91c4/Cabal-2.0.1.0/build/Dhall/Context.o )
[2018-06-18T16:57:55Z]     [ 2 of 13] Compiling Dhall.Core[boot] ( src/Dhall/Core.hs-boot, .stack-work/dist/x86_64-linux-nix-9ab91c4/Cabal-2.0.1.0/build/Dhall/Core.o-boot )
[...]

PR #448, opened 2018-06-09
Split Dhall.Parser in smaller modules

zudov:
This is a good thing on it's own, but beside that it's very likely to help with #443


Gabriel439:
Thank you! :)

Gabriel439:
@eskimor: I don't know if you meant it as a joke but just in case I want to remind you that the Dhall contributing guidelines require that people treat others with dignity and respect:

https://github.com/dhall-lang/dhall-lang/blob/master/.github/CONTRIBUTING.md#do-i-belong-here

src/Dhall/Parser.hs
```diff
@@ -61,710 +62,9 @@ import qualified Text.Parser.Combinators
...
-
-    let indexed = do
-            _ <- Text.Parser.Char.char '@'
-            Text.Parser.Token.natural
-
-    n <- indexed <|> pure 0
-    return (V x n)
+import Dhall.Parser.Combinators
+import Dhall.Parser.Token
+import Dhall.Parser.Expression
```
Gabriel439: I think you need to add `Dhall.Parser.Expression` to version control because I don't see it in the set of changes

zudov: Oh yes, sorry, I've pushed it now (and also added the modules to cabal file).

Thanks for looking at it, it's still a work in progress, so not really ready for final review. Let's discuss it tomorrow.

zudov:
@eskimor @Gabriel439 Sorry about it, we were siting next to each other during that "review" and somehow ended up at bad joke.

FintanH:
Beautiful 👌

eskimor:
> https://github.com/dhall-lang/dhall-lang/blob/master/.github/CONTRIBUTING.md#do-i-belong-here

@zudov knew about that, but he said it was ok as I just insulted his code and not him. In any case, I'll update my comment as the PR is of course not all crap ;-)

eskimor:
Also I am going to be back to serious now, as I am on my way back home already. Have an awesome remaining Zurihac! See you guys at next awesome Haskell event!

src/Dhall/Parser/Combinators.hs
```diff
@@ -0,0 +1,161 @@
+{-# LANGUAGE CPP                        #-}
+{-# LANGUAGE DeriveDataTypeable         #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE RecordWildCards            #-}
+module Dhall.Parser.Combinators where
```
Gabriel439: Maybe add a header to each module explaining the scope of what belongs in that module

zudov:
To summarize: I moved a bunch of code around without making any actual changes (to avoid regressions). The structure now goes as follows:

- `Dhall.Parser` :: The public parser interface, pretty much all definitions are exported and documented.
- `Dhall.Parser.Expression` :: High-level parsers that output `Expr` (or `Chunks`)
- `Dhall.Parser.Token` :: A bunch of parsers and scanners that capture small pieces of Dhall syntax outputting data-types from which `Expr`s can be constructed. Deals with things like literals, reserved words, comments, labels, filepaths, etc 
- `Dhall.Parser.Combinators` :: Utulities that don't directly relate to Dhall's syntax, but only help defining the rest of the parser.

`Expression` and `Token` modules are still quite big, so essentially instead of having one kitchen sink for everything, we now have a compound kitchen sink (`Dhall.Parser`) that consists of big sink for high-level dishes (`Dhall.Parser.Expression`), smaller sink for smaller dishes and cups (`Dhall.Parser.Token`) and a container with some cleaning devices (`Dhall.Parser.Combinators`).

As a next step I want to split it into more meaningful modules groupped by expression types, e.g. a module for imports, a module for literals, a module for records, a module for primitives, etc. Current structure is a very good stepping stone for that since now I can simply move stuff from `Expression` module and pull whatever needed from `Token` along with it.

In the end, this is ready to be merged and I'll continue with further localization and refactoring.

src/Dhall/Parser/Combinators.hs
zudov: :+1: They all have it, except this one. Adding now.

zudov: Added a missing one, and refined some others.

Gabriel439:
@zudov: Sounds good!  I'll go ahead and merge this, then.  For future reference, you have the commit bit and therefore you have permission to merge your pull requests yourself if somebody else approves the change.  The only exception I can think of is for pull requests that are waiting on a matching change to the language standard and even then it's not a big deal if you accidentally merge them prematurely

The worst that happens is that I can always open another pull request proposing any modifications that I had in mind

PR #449, opened 2018-06-09
WIP: Implement `as Location` import syntax

kirelagin:
A fix for dhall-lang/dhall-lang#71.

This PR adds two new types:

* `FilePath`
* `Url`

and literals for them. The syntax is

* `./some/path as Location` and
* `https://example.com/url as Location` respectively.

Trying to use an `env:` import with `as Location` is an error.

TODO:

- [ ] Update the language spec :).
- [x] ~Take headers which come from hashes (?) into account in diff.~
- [ ] Use a real `Uri` type for `Url` literals instead of current `Text`?

### DIFF
--- a/src/Dhall/Core.hs
+++ b/src/Dhall/Core.hs
@@ -195,7 +195,11 @@ instance Buildable ImportType where
         "env:" <> build env <> " "
 
 -- | How to interpret the import's contents (i.e. as Dhall code or raw text)
-data ImportMode = Code | RawText deriving (Eq, Ord, Show)
+data ImportMode
+    = Code
+    | RawText
+    | RawLocation
+    deriving (Eq, Ord, Show)
 
 -- | A `ImportType` extended with an optional hash for semantic integrity checks
 data ImportHashed = ImportHashed
@@ -227,8 +231,9 @@ instance Buildable Import where
     build (Import {..}) = build importHashed <> suffix
       where
         suffix = case importMode of
-            RawText -> "as Text"
-            Code    -> ""
+            Code        -> ""
+            RawText     -> "as Text"
+            RawLocation -> "as Location"
 
 instance Pretty Import where
     pretty import_ = Pretty.pretty (Builder.toLazyText (build import_))
@@ -352,6 +357,14 @@ data Expr s a
     | TextLit (Chunks s a)
     -- | > TextAppend x y                           ~  x ++ y
     | TextAppend (Expr s a) (Expr s a)
+    -- | >  FilePath                                ~  FilePath
+    | FilePath
+    -- | >  FilePathLit p                           ~  p
+    | FilePathLit FilePath
+    -- | >  Url                                     ~  Url
+    | Url
+    -- | >  UrlLit u                                ~  u
+    | UrlLit Text
     -- | > List                                     ~  List
     | List
     -- | > ListLit (Just t ) [x, y, z]              ~  [x, y, z] : List t
@@ -453,6 +466,10 @@ instance Monad (Expr s) where
     Text                 >>= _ = Text
     TextLit (Chunks a b) >>= k = TextLit (Chunks (fmap (fmap (>>= k)) a) b)
     TextAppend a b       >>= k = TextAppend (a >>= k) (b >>= k)
+    FilePath             >>= _ = FilePath
+    FilePathLit p        >>= _ = FilePathLit p
+    Url                  >>= _ = Url
+    UrlLit u             >>= _ = UrlLit u
     List                 >>= _ = List
     ListLit a b          >>= k = ListLit (fmap (>>= k) a) (fmap (>>= k) b)
     ListAppend a b       >>= k = ListAppend (a >>= k) (b >>= k)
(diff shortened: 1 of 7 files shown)


kirelagin:
> Take headers which come from hashes (?) into account in diff.

We agreed that it does not make sense to specify a hash for a url used `as Location`, so we should throw an error. (Another reasonable option would be to ignore it completely, which is how it works at the moment.)

kirelagin:
It turns out, I overlooked one part of url syntax, namely, `using` for specifying request headers. I think, the right thing to do is to prohibit them in url values.

I have actually implemented it already, will push later because no wifi.

kirelagin:
I am not really sure what to do with urls. I am now starting to think that adding a dependency on some library for handling urls might be not worth it. What do you think?

src/Dhall/Pretty/Internal.hs
```diff
@@ -1084,6 +1094,10 @@ buildExprF DoubleShow =
     "Double/show"
 buildExprF Text =
     "Text"
+buildExprF FilePath =
+    "FilePath"
+buildExprF Url =
+    "Url"
```
Gabriel439: Perhaps `Url` should be capitalized as `URL` since it is an acronym

src/Dhall/Import.hs
```diff
@@ -508,128 +508,154 @@ instance Show HashMismatch where
...
+-- | Exception thrown when @importType@ and @importMode@ are incompatible
+data IllegalImportType = EnvAsLocation deriving (Typeable)
 
-            return (path, text)
+instance Exception IllegalImportType
 
-        URL prefix file suffix maybeHeaders -> do
-            m <- needManager
+instance Show IllegalImportType where
+    show EnvAsLocation = _ERROR <> ": env cannot be imported as a location"
```
Gabriel439: `s/env/an environment variable/` and `s/as location/❰as Location❱/`

src/Dhall/Diff.hs
```diff
@@ -1088,6 +1104,19 @@ diffExprF l@(TextLit {}) r =
     mismatch l r
 diffExprF l r@(TextLit {}) =
     mismatch l r
+diffExprF (FilePathLit aL) (FilePathLit aR) =
+    diffText (Text.pack aL) (Text.pack aR)
+diffExprF l@(FilePathLit _) r =
+    mismatch l r
+diffExprF l r@(FilePathLit _) =
+    mismatch l r
+-- TODO: Diff maybeHeaders?
```
Gabriel439: As we discussed offline, we can probably have import resolution fail if the user specifies both `using` and `as Location`

Gabriel439:
Oh, one other thing: don't forget to mention the `as Location` feature somewhere in the tutorial

src/Dhall/Import.hs
kirelagin: ✅

src/Dhall/Core.hs
```diff
@@ -352,6 +357,14 @@ data Expr s a
     | TextLit (Chunks s a)
     -- | > TextAppend x y                           ~  x ++ y
     | TextAppend (Expr s a) (Expr s a)
+    -- | >  FilePath                                ~  FilePath
+    | FilePath
```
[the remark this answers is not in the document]
kirelagin: Well, I don’t know, are you sure? I feel that it will be somewhat more ambiguous. Also, currently it nicely corresponds to the Haskell type, as other primitive types (e.g. `Integer`, `Text`).

src/Dhall/Pretty/Internal.hs
kirelagin: Oh, sorry, I always capitalise this way, as most coding guidelines suggest using camel case with acronyms, so I didn’t even think about it.

Anyway, I can capitalise the text representation of the type, but if I change the constructor itself, then it will conflict with `URL` from `ImportType`. I guess, I can add `Import` as a prefix to constructors of `ImportType`. What do you think?

kirelagin:
By the way, I finally read some code around the parts I changed (😄) and I should point out that new relative path literals (somewhat) violate referential transparency: the check that local files cannot be imported from a remote one is done only when actually reading, so it is not performed for `as Location`.

On the one hand, it makes perfect sense for the cases when, e.g. some sort of default configuration is loaded from a remote URL and paths to other files in it are resolved relative to current directory and it probably doesn’t harm too much, since the referred files are not interpreted and even not read. On the other hand, your notion of referential transparency seems to be a good idea. I’m not sure what to do.

kirelagin:
Wait, no, it is resolved relative to the URL of the file that contains it, not to current directory 😕.
Status update: I have no idea how it works...

Gabriel439:
@kirelagin: What use case do you have in mind for this?  I ask because in https://github.com/dhall-lang/dhall-lang/issues/71 one of the ideas was that the import would actually be checked to see if it was valid

The reason I mention this is that if you validate the import tagged with `as Location` then the hash and headers become relevant again and the referential transparency check gets invoked

kirelagin:
Well, I have a felling (although, I have no evidence to back this claim) that software typically wants to just read a path from its config and then deal with any issues itself (e.g. if a database does not exist, create it) instead of getting an exception from its configuration “parser”.

Gabriel439:
Yeah, the standard is the big issue.  This has to be standardized before the Haskell implementation can be changed

You don't necessarily need to put up a pull request with the changes to the semantics and grammar.  Just open an issue against the `dhall-lang` repository with an informal description of the proposed change so that people can discuss it and the motivation behind it

Gabriel439:
I'm closing this since it's drifted quite a bit from `master` and this still needs to be standardized before implementation

PR #451, opened 2018-06-10
Hash and format subcommands

Fingerzam:
Added subcommands for `hash` and `format`. The implementation is mostly just copied from `dhall-format` and `dhall-hash`. Fixes #436.

Issue #437, opened 2018-06-07
Add `dhall freeze` subcommand


Gabriel439:
It would be nice if users could more easily take advantage of [semantic integrity checks](http://www.haskellforall.com/2017/11/semantic-integrity-checks-are-next.html) by automatically pinning all imports in a file.  For example, if a user had the following file:

```haskell
-- ./example.dhall

    let replicate =
          https://raw.githubusercontent.com/dhall-lang/Prelude/c79c2bc3c46f129cc5b6d594ce298a381bcae92c/List/replicate

in  replicate 5
```

... then they could pipe that through `dhall freeze` to pin the `replicate` import with a semantic integrity check:

```bash
$ dhall freeze < ./example.dhall
    let replicate =
          https://raw.githubusercontent.com/dhall-lang/Prelude/c79c2bc3c46f129cc5b6d594ce298a381bcae92c/List/replicate sha256:b0e3ec1797b32c80c0bcb7e8254b08c7e9e35e75e6b410c7ac21477ab90167ad

in  replicate 5
```

... or, like `dhall-format`, there could be an `--inplace` option, too:

```bash
$ dhall freeze --inplace ./example.dhall
```

joneshf:
Yes! This would be amazing!

gilligan:
Trying to move up from trivial PRs/issues to more interesting ones so I like this one but I would need some advise on how to approach this.

Would someone be willing to give me some guidance?

ocharles:
@gilligan Sure! I think the rough idea is to parse Dhall files into expressions, but not resolve imports. Then, you'll need to write a function that walks the resulting `Expr` and looks for import statements. Whenever you see an import that *doesn't* have an integrity hash, you would need to fetch that file, normalise it, and calculate the integrity hash. Then, you would replace the users original import with an import that uses the integrity hash.

What that hash is, I'm not sure of, but it sounds like @Gabriel439 is suggesting sha256, whereas I was suggesting a Dhall expression hash (I think such a thing exists!?)

gilligan:
Thanks, very helpful!

gilligan:
Since i'm bouncing between "ah, this is trivial" and "Huh, and now what?" i'll have to ask for some more help I'm afraid :) 

I guess my main problem is that I don't understand the `Expr` data type and I thought that i could just be naive about it and navigate around it but it doesn't seem to be working out. 

So what I am looking at right now is:

- Some dhall input  (`Text`)
[...]

PR #452, opened 2018-06-10
Make format and dhall subcommands of dhall executable

### LINKED ISSUE #436: Make `dhall-format` and `dhall-hash` subcommands instead of standalone executables
For command-line consistency, it would be nice for `dhall-format` and `dhall-hash` to become `dhall format` and `dhall hash` (i.e. `format` and `hash` subcommands of the `dhall` executable)

gilligan:
I really just moved code around a bit - fixes #436 

I just realized that Tutorial.hs also has to be adjusted since it is using `dhall-format` and `dhall-hash`. Another thing is with `--inplace` now being a global switch. Is that Ok @Gabriel439 or should that rather be moved down to be a switch to the `format` subcommand?


gilligan:
Oops, just saw the upstream changes and fixing..

src/Dhall/Main.hs
```diff
@@ -65,6 +68,13 @@ parseOptions = Options <$> parseMode <*> parseExplain <*> parsePlain
             <>  Options.Applicative.help "Disable syntax highlighting"
             )
 
+    parseInplace =
```
Gabriel439: Yeah, I think this should be a `format`-specific option (See https://github.com/dhall-lang/dhall-haskell/pull/451 for an example of how to do that)

Note that it's still worth keeping the intermediate `parseInplace` utility as I imagine that other utilities might reuse it in the future (like the `dhall freeze` command described in https://github.com/dhall-lang/dhall-haskell/issues/437)

gilligan: :ok_hand:

gilligan:
@Gabriel439 i also updated Tutorial.hs, let me know if that works for you.

@fingerzam I suppose your PR was actually ready before mine. I somehow missed it. Sorry. I actually took the format subcommand code from you now but I moved hash and format to their own modules..

Gabriel439:
As I mentioned in #436 I have decided to merge this pull request.  Thanks to both of you, though, for taking the time to do this 🙂

PR #454, opened 2018-06-10
Add template-haskell-2.12.0.0 to lts-6

jneira:
Cause picking 2.10 (default version in lts 6) makes dhall not compile: in that version [dataToExpQ](https://github.com/dhall-lang/dhall-haskell/blob/master/src/Dhall/TH.hs#L37) is in `Language.Haskell.TH.Quote` and not in `Language.Haskell.TH.Syntax`

### DIFF
--- a/src/Dhall/TH.hs
+++ b/src/Dhall/TH.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE CPP             #-}
 
 {-| This module provides `staticDhallExpression` which can be used to resolve
     all of an expression’s imports at compile time, allowing one to reference
@@ -25,6 +26,10 @@ module Dhall.TH where
 
 import Control.Monad
 import Data.Typeable
+#if MIN_VERSION_template_haskell(2,11,0)
+#else
+import Language.Haskell.TH.Quote
+#endif
 import Language.Haskell.TH.Syntax
 
 import qualified Data.Text as Text


jneira:
@sellout we should set the min bound but the stack file change still is needed (cause stack build would fail)
Let me add the change in the cabal file

Gabriel439:
@jneira: If you guard the import with `CPP` you can make the file build against multiple versions of `template-haskell`

jneira:
@Gabriel439 yeah, i tried it in my first fix. i'll push the changes

jneira:
I was reluctant to add cpp macros but i suppose some times is inevitable 😉

jneira:
@Gabriel439 master rebased

quasicomputational:
I think maybe we can get away without CPP here by unconditionally importing from `Quote`. Let me test locally and I'll report back

Gabriel439:
@quasicomputational: Oh yeah.  I didn't realize that `dataToExpQ` still exported from `Language.Haskell.TH.Quote` in both older and newer versions of `template-haskell`

Gabriel439:
@jneira: I believe GHC will correctly handle the case where the same identifier is imported from two different modules

Gabriel439:
Also, if we fully qualified the identifiers then there would also not be any issue

quasicomputational:
Works fine without the CPP. When things are simply re-exported, GHC is smart enough not to complain (usually). I've opened an alternative PR for it, #456.

jneira:
@quasicomputational :+1: ...closing this

sellout:
@jneira / @quasicomputational – thanks for cleaning up my mess.

PR #465, opened 2018-06-12
Add missing Diff dependency to doctests' build-depends

quasicomputational:
They were breaking without this for me when running with `new-test`.


ocharles:
I think the standard build environment has diff available.

quasicomputational:
Probably a difference in build tools. I don't know how to work Nix, but `stack test` on HEAD works though `cabal new-test` fails. It wouldn't surprise me at all if Hydra came back legitimately green.

PR #474, opened 2018-06-15
First draft at implementing union type combination

aleator:
This is not yet for merging, but it tries to implement the \\\ - operator for combining union types.

It's bit late here so I kinda just fumbled my way through this without thinking all that much...


Gabriel439:
Just noting that the matching change to the standard is here: https://github.com/dhall-lang/dhall-lang/pull/176

aleator:
Ok.. How does it look like now?

Gabriel439:
I'll close this since it's pretty out of date.  When we're ready to revisit this we can open a new pull request

PR #475, opened 2018-06-17
Optionally optimize for speed by replacing slow construct with native function

gip:
The aim of this PR is not to be merged. I think `dhall-haskell` should be the reference implementation in Haskell and be kept as simple as possible. 

However, I'm wondering if it would be possible to add hooks in the implementation that could be used by packages built on top of this one to optimize the normalization of some construct. I'm using Dhall as an intermediate representation and my application ended up generating a function similar to the one below. It was applied to large natural numbers:
```
λ(x : Natural) → λ(y : Natural) →
  Natural/fold y Natural (
    λ(n : Natural) → (
      Natural/fold n
                   { prev : Natural, next : Natural }
                   (λ(p : { prev : Natural, next : Natural }) → { prev = p.next, next = p.next + 1})
                   { prev = 0, next = 0 }).prev
  ) x 
```

When applied to (for instance) `9999` and `8888`, normalization takes forever as seen below. The optimized version is necessary to get acceptable performance. Of course I could try to change the code generated in the first place, but being able to speed up normalization would be great -- well implemented native functions are as safe as running the Dhall interpreter. 
```
Gilless-MacBook-Pro:dhall-haskell gilles$ time stack exec dhall normalize < ./t_9999_8888.dhall
1111

real	2m28.098s
user	2m25.813s
sys	0m1.425s
Gilless-MacBook-Pro:dhall-haskell gilles$ time stack exec dhall normalizeOpt < ./t_9999_8888.dhall
1111

real	0m0.269s
user	0m0.163s
sys	0m0.096s
```

### DIFF
--- a/dhall.cabal
+++ b/dhall.cabal
@@ -210,7 +210,8 @@ Library
         Dhall.Repl
         Dhall.TH,
         Dhall.Tutorial,
-        Dhall.TypeCheck
+        Dhall.TypeCheck,
+        Dhall.Optimizer
     Other-Modules:
         Dhall.Pretty.Internal,
         Dhall.Parser.Expression,
@@ -240,6 +241,7 @@ Test-Suite tasty
         Tutorial
         TypeCheck
         Util
+        Optimization
     Build-Depends:
         base                      >= 4        && < 5   ,
         deepseq                   >= 1.2.0.1  && < 1.5 ,
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -215,7 +215,7 @@ inputFromWith filename (Type {..}) ctx n txt = do
             _ ->
                 Annot expr' expected
     _ <- throws (Dhall.TypeCheck.typeWith ctx annot)
-    case extract (Dhall.Core.normalizeWith n expr') of
+    case extract (Dhall.Core.normalizeWith Nothing n expr') of
         Just x  -> return x
         Nothing -> Control.Exception.throwIO InvalidType
 
@@ -243,7 +243,7 @@ inputExprWith ctx n txt = do
     expr  <- throws (Dhall.Parser.exprFromText "(input)" txt)
     expr' <- Dhall.Import.loadWithContext ctx n expr
     _ <- throws (Dhall.TypeCheck.typeWith ctx expr')
-    pure (Dhall.Core.normalizeWith n expr')
+    pure (Dhall.Core.normalizeWith Nothing n expr')
 
 -- | Use this function to extract Haskell values directly from Dhall AST.
 --   The intended use case is to allow easy extraction of Dhall values for
--- a/src/Dhall/Core.hs
+++ b/src/Dhall/Core.hs
@@ -34,6 +34,7 @@ module Dhall.Core (
     -- * Normalization
     , alphaNormalize
     , normalize
+    , normalizeOpt
     , normalizeWith
     , Normalizer
     , judgmentallyEqual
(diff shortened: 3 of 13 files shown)


Gabriel439:
@gip: Just out of curiosity, what are you using subtraction for?  The first thing I want to discern is if there is a built-in worth adding to the language that provides equivalent power and efficiency

The second thing I want to note is that *in general* the goal of the Haskell implementation is not to serve as a reference implementation.  If it were then I would reject a lot of feature requests to add customization hooks :)

I do think at some point there should be a reference implementation that is much simpler, but I think that should be a separate project.

For this specific case, though, I would like to focus on the original problem you were trying to solve.  Usually the way I do things is:

* First, I recommend that people use the Dhall API to customize the set of built-ins (i.e. use `typeWith` and `normalizeWith`)
* If I see enough people performing the same customization then I try to see if there's some way to upstream it into the language standard

I also want to note that the only concern here is that the optimization might be brittle.  Adherence to the standard is not a concern here since it is still functionally equivalent, which is all that the standard requires.

gip:
Thanks for you comments.

> Just out of curiosity, what are you using subtraction for? The first thing I want to discern is if there is a built-in worth adding to the language that provides equivalent power and efficiency

It may be far from the initial design goal but basically, I'll be using Dhall as the language to implement smart contracts. It's ideal in few respects: statically typed, interpreter easy to write, tiny language. For safety it's important that everything happens inside the language. But I ran into the lack of substraction. The solution I came up with makes my programs terminate formally but not practically. 

> The second thing I want to note is that in general the goal of the Haskell implementation is not to serve as a reference implementation. If it were then I would reject a lot of feature requests to add customization hooks :)

I think that when a community is large then it makes sense for a separate reference implementation to exist. We are far from the size of WebAssembly at the moment :) It's my understanding that having a single reference repo that covers most of the use cases that exist today is probably the best approach.

> First, I recommend that people use the Dhall API to customize the set of built-ins

So that would mean that the AST needs to be extended? I'm far from convinced that extending the AST each time somebody wants something new is the way to go. I think the power of such a language is it's stability accross multiple platforms and languages. Extension of the AST will lead to fragmentation -- or the quick conclusion by very opiniated engineers that the language can't do what they want :)

Gabriel439:
@gip: If you use the Haskell API then you don't need to extend the AST to add new high-performance built-ins.

The first trick is that your builtin will be encoded in the AST as a free-variable.  Let's assume that you name the built-in for subtraction `Natural/subtract`.  The way you would use the built-in in your program is just refer to it like a free variable, i.e.:

```haskell
Natural/subtract 4 2
```

... so it will be parsed using the `Var` constructor and the AST after parsing (modulo `Noted` constructors) would be:

```haskell
App (App (Var (V "Natural/subtract" 0)) (NaturalLit 4)) (NaturalLit 2)
```

Then you use `Dhall.TypeCheck.typeWith` to type-check your expression with a custom context.  That context will specify that our new `Natural/subtract` built-in has type `Natural → Natural → Natural` so the type-checker will accept the above program.

Then once you type-check that you can normalize the program using `Dhall.Core.normalizeWith` which lets you provide a custom normalization hook for replacing all occurrences of `Natural/subtract x y` with machine subtraction.

I should probably write this up into a post on the wiki because I get this question a lot :)

ocharles:
I'm sure I don't have to remind you @Gabriel439, but it's worth emphasizing that rather than postulating a new builtin, you can also just supply `Natural/subtract` as an argument to your Dhall expression, which means that all other Dhall tools work out-of-the-box (e.g., `dhall repl`, `dhall type`, etc).

Gabriel439:
@ocharles: Oh yeah, I completely forgot to mention that :)

That's right.  You can assume built-ins using a lambda, like this:

```haskell
λ(Natural/subtract : Natural → Natural → Natural) → ...
```

Gabriel439:
Closing this since I added a new page to the Dhall wiki explaining how to use the Haskell API to customize the language with new built-in functions: https://github.com/dhall-lang/dhall-lang/wiki/How-to-add-a-new-built-in-function

PR #477, opened 2018-06-20
fix sha256 missing space in pretty print output 

justinwoo:
Looks like some SHA256 formatting got moved about or something.


Gabriel439:
Looks great!  I think you just need to strip the trailing newline from the expected test output

justinwoo:
Pushed up a fix and it's fixed now.

Gabriel439:
Awesome!  Thank you :)

PR #482, opened 2018-06-22
Implement 'with-http' cabal flag for disabling HTTP support

DanielG:
I've been wanting to try using dhall for some of my projects for a while now but it having HTTP support for imports always stopped me, both from a dependency closure size and security perspective. Support for attaching hashes to imports certainly made the security concerns less pressing but I still highly dislike depending on the gazillion http related libraries which I'm not even going to use.

Hence this PR.

I just quickly threw this together and haven't tested it yet beyond making sure it compiles with the new flag on and off. Let me know if you agree with the general approach here and I'd be happy to spend some more time on polishing this.


Gabriel439:
@DanielG: Yeah, this is totally reasonable to me.  I'm happy to merge whenever you're ready

DanielG:
Ok I've made sure it compiles with GHC 8.0 and 8.4, with and without `http-client >0.5` and re-added the export for `PrettyHttpException` (which doesn't seem to be used anywhere?). I'm not sure things still work with 7.10 since I can't get that GHC version to work on my machine anymore (that PIE issue and all that).

Do be aware that the exported API has changed in a PVP major version bump kind of way since I had to change `PrettyHttpException` so it doesn't reference the `HttpException` type.

Anyways, should be good to go from my end.

PR #485, opened 2018-06-24
Fix missing space in import suffix formatting

f-f:
Before:

```bash
❯ dhall-format <<< "env:AAA as Text"
env:AAAas Text
```


f-f:
@Gabriel439 can I merge or should I wait after the release?

Gabriel439:
@f-f: Definitely merge.  The reason there is a waiting period of up to 7 days is so that people have a chance to flush any pending changes

PR #486, opened 2018-06-25
Add freeze command

gilligan:
This PR adds a `freeze` command as described in #437 

Given 
```
    let replicate =
          https://raw.githubusercontent.com/dhall-lang/Prelude/c79c2bc3c46f129cc5b6d594ce298a381bcae92c/List/replicate

in  replicate 5
```

the freeze command will produce:

```
    let replicate =
          https://raw.githubusercontent.com/dhall-lang/Prelude/c79c2bc3c46f129cc5b6d594ce298a381bcae92c/List/replicate sha256:b0e3ec1797b32c80c0bcb7e8254b08c7e9e35e75e6b410c7ac21477ab90167ad 

in  replicate 5
```

---------

The PR doesn't include tests or updates to `Tutorial.hs` - can those be done separately or should those be included to make this eligible for merge?

Thanks to @Gabriel439 and @ocharles for the help on this.

### DIFF
--- a/src/Dhall/Freeze.hs
+++ b/src/Dhall/Freeze.hs
@@ -4,11 +4,14 @@ module Dhall.Freeze (freeze) where
 
 import Dhall.Core
 import Dhall.Import (load, hashExpression)
-import Dhall.Parser (exprAndHeaderFromText)
-import Data.Text.Prettyprint.Doc (Pretty)
-import Dhall.Pretty (prettyExpr)
+import Dhall.Parser (exprAndHeaderFromText, Src)
+import Dhall.Pretty (annToAnsiStyle)
 
+import System.Console.ANSI (hSupportsANSI)
+import Data.Monoid ((<>))
+import Data.Maybe (fromMaybe)
 import Data.Text
+
 import qualified Data.Text.Prettyprint.Doc                 as Pretty
 import qualified Data.Text.Prettyprint.Doc.Render.Terminal as Pretty
 import qualified Control.Exception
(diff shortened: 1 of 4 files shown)


src/Dhall/Main.hs
```diff
@@ -139,6 +142,19 @@ parseMode =
             parserWithHelper = Options.Applicative.helper <*> parser
             parser = Format <$> optional parseInplace
 
+    freezeSubcommand =
```
Gabriel439: I think you can simplify this as:

```haskell
freezeSubcommand = subcommand "freeze" "Add hashes to all import statements of an expression" parseFreeze
  where
    parseFreeze = Freeze <$> optional parseInplace
```

src/Dhall/Freeze.hs
```diff
@@ -0,0 +1,73 @@
...
+opts =
+    Pretty.defaultLayoutOptions
+        { Pretty.layoutPageWidth = Pretty.AvailablePerLine 80 1.0 }
+
+readInput :: Maybe FilePath -> IO Text
+readInput = maybe fromStdin Data.Text.IO.readFile
+    where 
+        fromStdin = System.IO.hSetEncoding System.IO.stdin System.IO.utf8 >> Data.Text.IO.getContents
+
+hashImport :: Import -> IO Import
```
Gabriel439: I think you should also export this utility since it is handy for people using the Haskell API

gilligan:
Added a paragraph on freeze to the Tutorial, please take a look.

src/Dhall/Freeze.hs
gilligan: Done.

src/Dhall/Tutorial.hs
```diff
@@ -1551,6 +1551,24 @@ import Dhall
 -- behavior-preserving.  This provides an easy way to detect refactoring errors
 -- that you might accidentally introduce.  The hash not only protects you
 -- from attackers, but also protects against human error, too!
+--
+-- If you have a file which either doesn't already use hashed imports,
+-- or you changed some of the imports and want to update the hashes you can use the
+-- freeze command to either add or updated hashes:
```
Gabriel439: `s/updated/update/`

src/Dhall/Tutorial.hs
```diff
@@ -1551,6 +1551,24 @@ import Dhall
...
+--
+-- > cat foo.dhall
+-- ''
+--    let replicate = ./r.dhall
```
Gabriel439: Maybe use the GitHub URL for the import so that users can reproduce the example and also so that I can later add the example to the test suite

PR #495, opened 2018-06-29
Tutorial.hs: add small section about backticking labels

Profpatsch:
Unfortunately, I was not able to test the output of this, since haddock in the fixed nixpkgs segfaults on me; see https://github.com/haskell/haddock/pull/846


Gabriel439:
I'll close this as the backticks will no longer be necessary once I finish implementing https://github.com/dhall-lang/dhall-lang/pull/437

PR #496, opened 2018-06-30
Use `doctest` to test `Dhall.Tutorial` module

### LINKED ISSUE #493: Doctests for Dhall/Tutorial.hs
The tutorial’s doctests are not run at the moment.

I imagine the reason is that it does not fare well with dhall imports?
If it’s possible to specify the “file folder” that dhall sees somehow, the example files can be put into a folder and the doctests pointed at it.

Gabriel439:
Fixes #493

### DIFF
--- a/doctest/Main.hs
+++ b/doctest/Main.hs
@@ -3,4 +3,10 @@ module Main where
 import qualified Test.DocTest
 
 main :: IO ()
-main = Test.DocTest.doctest [ "-isrc", "src/Dhall.hs", "src/Dhall/Import.hs" ]
+main =
+    Test.DocTest.doctest
+        [ "-isrc"
+        , "src/Dhall.hs"
+        , "src/Dhall/Import.hs"
+        , "src/Dhall/Tutorial.hs"
+        ]
--- a/src/Dhall/Tutorial.hs
+++ b/src/Dhall/Tutorial.hs
@@ -482,7 +482,7 @@ import Dhall
 -- *** Exception: 
 -- ↳ ./baz: 
 -- ...
--- ...Error...: Missing file /Users/gabriel/proj/dhall-haskell/baz:
+-- ...Error...: Missing file .../baz:
 -- ...
 --
 -- This is because the parser thinks that @./baz:@ is a single token due to
--- a/src/Dhall/Tutorial.hs
+++ b/src/Dhall/Tutorial.hs
@@ -168,6 +168,16 @@ module Dhall.Tutorial (
 import Data.Vector (Vector)
 import Dhall
 
+-- $setup
+--
+-- >>> :set -XOverloadedStrings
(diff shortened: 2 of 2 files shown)


src/Dhall/Tutorial.hs
```diff
@@ -168,6 +168,16 @@ module Dhall.Tutorial (
 import Data.Vector (Vector)
 import Dhall
 
+-- $setup
+--
+-- >>> :set -XOverloadedStrings
+-- >>> writeFile "makeBools" "λ(n : Bool) → [ n && True, n && False, n || True, n || False ]"
```
Profpatsch: Wait, where does the file go? I can’t find anything about the context in which doctests are run in its documentation.

src/Dhall/Tutorial.hs
```diff
@@ -2808,3 +2822,14 @@ import Dhall
 -- this guide:
 --
 -- <https://github.com/dhall-lang/dhall-lang/wiki/How-to-translate-recursive-code-to-Dhall How to translate recursive code to Dhall>
+
+-- $cleanup
+--
+-- >>> System.Directory.removeFile "bool"
```
Profpatsch: Is all this run in the current folder?

src/Dhall/Tutorial.hs
Gabriel439: doctests are run from the root folder of the project, so the file is added to the root foldr

src/Dhall/Tutorial.hs
Gabriel439: Yes

src/Dhall/Tutorial.hs
Profpatsch: Do you think we could switch to a `TMP` folder first?

Gabriel439: I tried that and it doesn't work.  `ghci` unloads all modules when you `:cd` to another directory so it breaks `doctest`

Profpatsch:
I opened an [issue](https://github.com/sol/doctest/issues/204) about temporary files in doctest. Other than that, I think we can merge.

src/Dhall/Tutorial.hs
sol: As a side note, I think technically you can run `doctest` in a temporary directory, but you would need to then specify `-i` accordingly so that `doctest` will find your sources.

doctest/Main.hs
```diff
@@ -3,4 +3,10 @@ module Main where
 import qualified Test.DocTest
 
 main :: IO ()
-main = Test.DocTest.doctest [ "-isrc", "src/Dhall.hs", "src/Dhall/Import.hs" ]
+main =
```
sol: More specifically: Create a temporary directory here...

sol: ...and `cd` into it...

PR #497, opened 2018-07-01
Fix documentation for Natural/Integer caveats

markus1189:
I noticed that the docs in the `Caveats` section are wrong about `Integer` and `Natural` regarding the prefixes.  Should be fixed now ;)


sellout:
It looks like you’re using an older version of Dhall locally, as the current version should match the docs as they are.

markus1189:
Ooops indeed, good to know :sweat_smile:

PR #503, opened 2018-07-04
Use temporary directory for tutorial example files

Gabriel439:
... as suggested by @sol

### DIFF
--- a/doctest/Main.hs
+++ b/doctest/Main.hs
@@ -4,7 +4,7 @@ import Data.Monoid ((<>))
 import System.FilePath ((</>))
 
 import qualified System.Directory
-import qualified System.IO.Temp
+import qualified Test.Mockery.Directory
 import qualified Test.DocTest
 
 main :: IO ()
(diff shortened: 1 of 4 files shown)


doctest/Main.hs
```diff
@@ -1,12 +1,31 @@
...
-        , "src/Dhall.hs"
-        , "src/Dhall/Import.hs"
-        , "src/Dhall/Tutorial.hs"
-        ]
+main = do
+    pwd    <- System.Directory.getCurrentDirectory
+    prefix <- System.Directory.makeAbsolute pwd
+
+    System.IO.Temp.withSystemTempDirectory "doctest" $ \directory -> do
+        System.Directory.setCurrentDirectory directory
```
sol: This works.  Using `Test.Mockery.Directory.inTempDirectory` would be slightly more concise.

https://hackage.haskell.org/package/mockery-0.3.5/docs/Test-Mockery-Directory.html

PR #467, opened 2018-06-12
Add `inputFromFile` and `inputExprFromFile` (#408)

gilligan:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/408

- `inputFromFile` reads file and invokes Dhall.input
- `inputExprFromFile` reads file and invokes Dhall.inputExpr

---

That was a quick one, did I miss something? :sweat_smile:

PR #508, opened 2018-07-13
Add various functions explicitly taking a root directory.

### LINKED ISSUE #507: CWD independence
dhall's API should be convenient to work with without switching the current working directory. However, currently the `input*` family of functions will only resolve imports relative to the CWD; these would be very useful to solve dhall-lang/dhall-to-cabal#114. There are hacks around this (e.g., parse the file, then munge all imports to include the desired prefix), but they're obviously hacks and doing the right thing ought to be easy, not hard.

quasicomputational:
Now the importing code is not tied to the current working directory:
it can resolve imports relative to any directory passed to it.

Fixes #507.

Notes: I haven't added every flavour of `Dir` function that I could have: I only put in the ones relevant to dhall-to-cabal, besides the most general one of each family that contains the actual implementation. These functions are getting a bit dense and we might consider some kind of alternative strategy here (e.g., an `InputSettings` type, where today's `input` becomes equivalent to `input defaultInputSettings`).

I didn't add an `importFile` function as suggested in #507 and previously implemented in #467.

I put the root directory in the `Status` object because caching relative imports with different root directories isn't safe. However, I note that it should be immutable, which is currently only enforced by convention: the `StateT` would allow modification of the root directory, despite that never being the correct thing to do.

### DIFF
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -219,7 +219,7 @@ inputFromWith filename ty ctx n txt =
     to, a file to mention in errors as the source, a custom typing
     context, and a custom normalization process.
 
-@since 1.6
+@since 1.16
 -}
 inputDirFromWith
     :: FilePath
@@ -283,7 +283,7 @@ inputExprWith ctx n txt = do
 {-| Extend `inputExpr` with a directory to resolve imports relative to,
     custom typing context and normalization process.
 
-@since 1.6
+@since 1.16
 -}
 inputExprDirWith
     :: FilePath
--- a/src/Dhall/Import.hs
+++ b/src/Dhall/Import.hs
@@ -547,7 +547,7 @@ loadWith from_import ctx n expr =
 -- context and `Import`-resolving callback in arbitrary `MonadCatch`
 -- monad, relative to a given directory.
 --
--- @since 1.6
+-- @since 1.16
 loadDirWith
     :: MonadCatch m
     => FilePath


quasicomputational:
One more wrinkle: the `_root` field in `Status` is a relative path. If `setCurrentWorkingDirectory` is called somewhere else, the cached imports in the `Status` will be wrong. I tried to fix this by calling `makeAbsolute` before `loadStaticWith` but I ran into trouble, since `loadWith` doesn't have a `MonadIO` constraint. I suppose that this is already problematic, so this PR isn't making the situation worse.

src/Dhall/Import/Types.hs
```diff
@@ -17,7 +17,11 @@ import Dhall.TypeCheck (X)
 
 -- | State threaded throughout the import process
 data Status = Status
-    { _stack   :: [Import]
+    { _root :: FilePath
```
Gabriel439: You don't need to add a `_root` field.  Instead, you can make `_stack` a `NonEmpty` list of `Import`s and set the first one to the directory with a file named `"."`

src/Dhall.hs
```diff
@@ -203,9 +212,32 @@ inputFromWith
...
+inputFromWith filename ty ctx n txt =
+  inputDirFromWith "." filename ty ctx n txt
+
+{-| Extend 'inputFrom' with a root directory to resolve imports relative
+    to, a file to mention in errors as the source, a custom typing
+    context, and a custom normalization process.
+
+@since 1.6
+-}
+inputDirFromWith
```
Gabriel439: Are you sure that you want this function to take two `FilePath` arguments.  Couldn't you infer the starting directory from the path of the source file?

quasicomputational: That's not guaranteed to be a path to the file; e.g., it's also `<input>` or `<stdin>` if that's where the data is actually coming from.

quasicomputational:
OK, I removed `_root` and now it all goes through `canonicalizeImport`, which is definitely a bit conceptually nicer but it had a bunch of knock-on refactorings. I've also added a basic smoke test for this new capability.

src/Dhall.hs
Gabriel439: Yeah, and also thinking more about your answer in the other issue I agree that it's necessary to disentangle the two concepts, so don't worry about this after all

src/Dhall/Import.hs
```diff
@@ -225,23 +226,25 @@ instance Show ReferentiallyOpaque where
...
-        imports' = zip [0..] (drop 1 (reverse (canonicalizeAll imports)))
-
+        indent n import_ =
+            "\n" ++ replicate (2 * n) ' ' ++ "↳ " ++ Dhall.Pretty.Internal.prettyToString import_
+        canonical = NonEmpty.toList (canonicalizeAll imports)
+        -- The immediate error's source is displayed by 'e', so we
+        -- need to not display the first (innermost) import, and the
+        -- final (outermost) import is fake to establish the base
+        -- directory. Also, we need outermost-first.
+        toDisplay = drop 1 (reverse (drop 1 canonical))
```
Gabriel439: The `doctest` error appears to be due to the new `drop 1` here.  According to the comment this appears to be intentional, but are you sure that the innermost import is already displayed by `e`?

quasicomputational: I thought it was! :( I'm having a look now too. That reasoning was based on the previous code having a `drop 1` in it but I might have confused myself while trying to get this function to work.

quasicomputational: Oh, it was on the wrong side of the `reverse` anyway for that. Now I don't know why the previous code had the `drop 1` in it at all.

Gabriel439: @quasicomputational: I think the reason it had the other `drop 1` is because it would implicitly insert `./.` as the root of the the chain of imports before your change and the `drop 1` removed it so it wouldn't show up in the chain of imports

quasicomputational:
Oh, I think I understand what was going on: the previous code was injecting its own fake import and that had to be dropped. We're still doing that, just more systematically.

src/Dhall.hs
```diff
@@ -203,9 +212,32 @@ inputFromWith
...
     -- ^ The decoded value in Haskell
-inputFromWith filename (Type {..}) ctx n txt = do
+inputFromWith filename ty ctx n txt =
+  inputDirFromWith "." filename ty ctx n txt
+
+{-| Extend 'inputFrom' with a root directory to resolve imports relative
+    to, a file to mention in errors as the source, a custom typing
+    context, and a custom normalization process.
+
+@since 1.6
```
Gabriel439: Did you mean `@since 1.15` or `@since 1.16`?

quasicomputational: Oops, yes. Might be contagious!

Gabriel439:
You're welcome!  Thank you, too 🙂

PR #510, opened 2018-07-14
Add `$` as an allowed character in quoted labels

f-f:
Implements https://github.com/dhall-lang/dhall-lang/pull/184


f-f:
@Gabriel439 thanks! Could we cut a minor release with this so we can use it right away?

Gabriel439:
@f-f: Here you go: https://github.com/dhall-lang/dhall-haskell/pull/511

PR #511, opened 2018-07-14
Version 1.15.1 → 1.16.0

Gabriel439:


Gabriel439:
You're welcome!  Note that I'm going to wait 3 days before releasing just to give people a chance to complete any pending pull requests

sellout:
Just since you’ve done it twice now, I figure I’ll point it out … you keep writing 1.5 instead of 1.15. Happened on the .1 release and now this one. 😄

quasicomputational:
#508 made a breaking change in the `Status` type, so the version needs to be 1.16 (or it needs to be backed out then re-applied after the release).

f-f:
Since the language standard now [has a version](https://github.com/dhall-lang/dhall-lang/releases/tag/v1.0.0) we could state in the `CHANGELOG` and in the `README` that this `1.16.0` release supports version `v1.0.0` of the language.

Gabriel439:
Sorry for the delay in cutting this release.  The reason for the delay is that I've been working on getting the fully static build of Dhall working (See: https://github.com/dhall-lang/dhall-lang/issues/192#issuecomment-406127492) since I wanted to include a static binary in this release.  I opening up a pull request for this soon and merging tomorrow so that I can cut the release.

PR #512, opened 2018-07-15
Proof-of-concept refactoring of the zoo of different input functions

quasicomputational:
This isn't necessarily meant to be merged as-is, but more as a
suggestion that this API can take. BC breaks aside, there are two
things I would like to bring up:

* Argument order. If `InputSettings` went last, we could use the tight
  precedence of record updates to write this:

  ```
     expr <- inputExprWithSettings txt defaultSettings
       { sourceName = "foobar" }
  ```

  However, we then can't use monad operators so conveniently. I'm not
  sure which is more important.

* A putative `inputFile`, which would do the Right Thing for
  `rootDirectory` and `sourceName` and hence would not take them as
  parameters, but which would still take a context and a normaliser.

### DIFF
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -20,6 +20,8 @@ module Dhall
     -- * Input
       input
     , inputWithSettings
+    , inputFile
+    , inputFileWithSettings
     , inputExpr
     , inputExprWithSettings
     , rootDirectory
@@ -95,6 +97,7 @@ import GHC.Generics
 import Lens.Family (LensLike', view)
 import Numeric.Natural (Natural)
 import Prelude hiding (maybe, sequence)
+import System.FilePath (takeDirectory)
 
 import qualified Control.Applicative
 import qualified Control.Exception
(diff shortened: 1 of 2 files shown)


quasicomputational:
I think the `fromFile` approach isn't quite sufficient, because then users still have to write `inputWithSettings (fromFile path) (readFile path)`. I also struggle to imagine that `fromFile` would be useful outside that pattern; better to wrap it up in a convenient `inputFile`, I think.

I've pushed some more commits to explore a slightly different design:
* a lensy interface to settings, hiding the record fields and constructors (which is good for avoiding compatibility breaks if we add any more fields)
* factoring out an `EvaluateSettings` type, which is what `inputFile` will take, and using a typeclass so that lenses to get at its fields work equally well on `EvaluateSettings` and `InputSettings`
* and then using that to put together a convenient `inputFile` and `inputFileWithSettings`.

Gabriel439:
Yeah, on more reflection I agree that having a dedicated `inputFile` might make more sense

quasicomputational:
Having slept on it, I'm happy with this lensy design if you are. Want to merge this for 1.16, or hold off in case anyone has any better ideas?

Gabriel439:
@quasicomputational: I'll go ahead and merge since we're already making a breaking change to the `input*` family of functions, so we might as well bundle the two breaking changes togeher

PR #521, opened 2018-07-22
Version 1.16.0 → 1.16.1

Gabriel439:


Gabriel439:
This is mainly to fix a test failure I discovered when updating downstream packages to depend on the Hackage version.  A test file wasn't declared in the `.cabal` file so it wasn't bundled in the upload to Hackage.

Here is an example occurrence of the test failure:

https://hydra.dhall-lang.org/build/2359/nixlog/4/tail

PR #522, opened 2018-07-23
Fix `isNormalized`

### LINKED ISSUE #516: Reversed logic in `BoolAnd` etc. cases of `isNormalized`
This logic (and similar pieces) looks backwards to me:

https://github.com/dhall-lang/dhall-haskell/blob/99db7f715b1515fd9033e2e82985aa9f606be398/src/Dhall/Core.hs#L1776-L1782

This is what I would expect:
```haskell
        case x of
            BoolLit _ -> False
            _ ->
                case y of
                    BoolLit _ -> False
                    _ -> True
```
because `&&` can be normalized if one or both of its operands are literals.

Gabriel439:
Fixes #516

00c79f5d654ae151f7171456c1786d89caa512d3 and
d287f020895e7b6f13f61eb58030e5207fe2a38a added additional simplifications to
Dhall, but without a matching change to `isNormalized`.  This causes
`isNormalized` to falsely report that these expressions are normalized when
in fact they can still be normalized.  This change fixes that by updating
`isNormalized` to match the behavior of `normalize`


Gabriel439:
Note that I'm waiting for https://github.com/dhall-lang/dhall-haskell/pull/519 to be merged because I'd like to build upon that in this pull request to add a `QuickCheck` property test verifying that `isNormalized` and `normalize` are consistent with one another.  I'm just putting up the code for people to review until then

PR #526, opened 2018-07-27
Improve parsing performance

Gabriel439:
This uses `takeWhile{,1}P` as a faster alternative to `many . satisfy` to
speed up parsing large labels

This also uses the `many` re-exported by `megaparsec` which is more efficient
for types that implement `MonadPlus`


ocharles:
Any numbers to go with this?

PR #528, opened 2018-07-28
Improve parsing performance on happy path

Gabriel439:
This uses `noted = id` on the first attempt to parse an expression and then
uses the real `noted` if parsing fails to provide better error messages


Gabriel439:
Thank god for the doctests.  These point out a fundamental issue with this pull request, which is that we need the source annotations even when parsing succeeds so that type-checking error messages can be properly localized.  I'll close this for now

Issue #525, opened 2018-07-26
Combinators for `InputType`s for Haskell records


eamsden:
Currently (1.16.1) the `dhall` package provides combinators to construct `Type` values for Haskell record types (the `record` and `field` combinators and the `Applicative` instance for `RecordType`) but no corresponding way to generate `InputType` values for Haskell record types.

I would like to implement the following API, along with any feedback requested by the maintainers:

```
data FieldType a =
  { fieldName :: Text
  , fieldSelector :: a -> Expr Src X
  , fieldType :: Expr Src X
  } deriving (etc...)

inputRecord :: [FieldType] -> InputType a

inputField :: (Inject b) => Text -> (a -> b) -> FieldType a
(.=) :: (Inject b) => Text -> (a -> b) -> FieldType a

inputFieldWith :: InputType b -> Text -> (a -> b) -> FieldType a
```
(Any suggestions for an operator name which doesn't conflict with Aeson would be welcome.)

Gabriel439:
One way to do this for symmetry with the `RecordType` API would be to do something like this:

```haskell
data RecordInputType a = ...

instance Contravariant RecordInputType
instance Divisible     RecordInputType

inputField :: Text -> InputType a -> RecordInputType a

inputRecord :: RecordInputType a -> InputType a
```

Then if you build on top of [the `Divisible` API](http://hackage.haskell.org/package/contravariant-1.5/docs/Data-Functor-Contravariant-Divisible.html) to define these utilities:

```haskell
divide0 :: Divisible f => f a
divide0 = conquer

divide1 :: Divisible f => (a -> b) -> f b -> f a
divide1 = contramap

divide2 :: Divisible f => (a -> (b, c)) -> f b -> f c -> f a
divide2 = divide

divide3 :: Divisible f => (a -> (b, c, d)) -> f b -> f c -> f d -> f a
divide3 k fb fc fd = contramap adapt (divided (divided fb fc) fd)
  where
    adapt a = ((b, c), d)
      where
        (b, c, d) = k a

divide4 :: Divisible f => (a -> (b, c, d, e)) -> f b -> f c -> f d -> f e -> f a
divide4 k fb fc fd = contramap adapt (divided (divided (divided fb fc) fd) fe)
  where
    adapt a = (((b, c), d), e)
      where
        (b, c, d, e) = k a

...
```

... and then you use those to build `InputType`s like this:

```haskell
[...]

PR #530, opened 2018-07-30
API to construct `InputType`s for Record types.

eamsden:
Adds the `RecordInputType` type and associated functions `inputFieldWith`, `inputField`, and `inputRecord`.

These `inputField{,With}` functions and the `Contravariant` and `Divisible` operations create `RecordInputType` values which describe how a Haskell record type may be encoded into a Dhall expression. The `inputRecord` function converts a `RecordInputType` to an `InputType`.

An example (using the same type as the example for `RecordType`) is given in the documentation comment for `RecordInputType.

See #525


src/Dhall.hs
```diff
@@ -1333,3 +1340,80 @@ field key valueType =
...
+> }
+
+    Our injector has type 'InputType' @Project@, but we can't build that out of any
+    smaller injectors, as 'InputType's cannot be combined (they are only 'Contravariant's).
+    However, we can use an 'InputRecordType' to build an 'InputType' for @Project@:
+
+> injectProject :: InputType Project
+> injectProject =
+>   inputRecord
+>     (  adapt >$< inputField "name" (inject @Text)
```
Gabriel439: I believe the type applications are not necessary since they should be inferred by the context

src/Dhall.hs
```diff
@@ -1333,3 +1340,80 @@ field key valueType =
...
+>              >*< inputField "description" (inject @Text)
+>              >*< inputField "stars" (inject @Natural)
+>     )
+>   where
+>     adapt (Project{..}) = (projectName, (projectDescription, projectStars))
+
+-}
+
+-- | Infix 'contramap'
+(>$<) :: Contravariant f => (a -> b) -> f b -> f a
```
Gabriel439: I just realized that the `>$<` operator already exists in the `contravariant` package:

http://hackage.haskell.org/package/contravariant-1.5/docs/Data-Functor-Contravariant.html

... so we can reuse that

eamsden: I can drop that definitiion and just rexport the contravariant definition. Give me a few minutes.

eamsden:
@quasicomputational maybe we should use the `inject` prefix? I was naming after `InputType`, which they are intended to aid construction of.

calling them `recordField{,With}` doesn't really give an idea of their distinction from `field`

src/Dhall.hs
eamsden: I think you are right. And actually, given the way the API is written, they should either be `inputFieldWith` or the second parameter should be dropped entirely. The example as written woudn't typecheck!

Gabriel439: Note that we have `doctest` enabled on this module, so if you transform the example into a doctest then it will be checked 🙂

eamsden: I might do that in a separate PR, as it would involve modifying the $setup block to bring the `Project` type into scope.

Gabriel439: Yeah, it's not urgent to do in this PR

src/Dhall.hs
```diff
@@ -1333,3 +1340,87 @@ field key valueType =
...
+>   where
+>     adapt (Project{..}) = (projectName, (projectDescription, projectStars))
+
+-}
+
+-- | Infix 'divided'
+(>*<) :: Divisible f => f a -> f b -> f (a, b)
+(>*<) = divided
+
+infixr 4 >*<
```
eamsden: I think this should be `infixl 4 >*<` to work properly with the `>$<` exported from contravariant, right?

Gabriel439: It should still be `infixr` if you want the tuple nesting to be `(a, (b, c))`, but the precedence needs to be elevated to 5

eamsden:
@Gabriel439 does this need anything else before it can be merged?

Gabriel439:
@eamsden: No, nothing needs to be done.  I just thought you had additional changes planned, but if you think this is ready then I'll go ahead and merge

PR #531, opened 2018-07-31
Add options and command-line support for protocol versions

Gabriel439:
The immediate goal of this pull request was to add command-line support
for a `--protocol-version` flag so that users could pin to a specific
binary protocol versions tring to avoid a forced upgrade when they
install a newer version of the Dhall interpreter.

However, in the process I ended up cleaning up the `Dhall.Import`
`load*` family of functions by consolidating everything into two
functions:

* `load` - Exact same type signature and semantics as before for the
  happy path with sensible defaults

* `loadWith` - The ultra-customizable version where all the options are
  stuffed into the existing `Status` record for ease of extension later

I also fixed a bug along the way where `decodeWithVersion_1_0` and
`encodeWithVersion_1_0` were serializing and deserializing the wrong
version string.

### DIFF
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -307,7 +307,6 @@ inputWithSettings settings (Type {..}) txt = do
                set Dhall.Import.protocolVersion _protocolVersion
             .  set Dhall.Import.normalizer      _normalizer
             .  set Dhall.Import.startingContext _startingContext
-            .  set Dhall.Import.rootDirectory   _rootDirectory
 
     let status = transform (Dhall.Import.emptyStatus ".")
 
@@ -401,7 +400,6 @@ inputExprWithSettings settings txt = do
                set Dhall.Import.protocolVersion _protocolVersion
             .  set Dhall.Import.normalizer      _normalizer
             .  set Dhall.Import.startingContext _startingContext
-            .  set Dhall.Import.rootDirectory   _rootDirectory
 
     let status = transform (Dhall.Import.emptyStatus ".")
 
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -28,6 +28,7 @@ module Dhall
     , sourceName
     , startingContext
     , normalizer
+    , protocolVersion
     , defaultInputSettings
     , InputSettings
     , defaultEvaluateSettings
@@ -89,18 +90,20 @@ import Data.Text (Text)
 import Data.Typeable (Typeable)
 import Data.Vector (Vector)
 import Data.Word (Word8, Word16, Word32, Word64)
+import Dhall.Binary (ProtocolVersion(..))
 import Dhall.Core (Expr(..), Chunks(..))
 import Dhall.Import (Imported(..))
 import Dhall.Parser (Src(..))
 import Dhall.TypeCheck (DetailedTypeError(..), TypeError, X)
 import GHC.Generics
-import Lens.Family (LensLike', view)
+import Lens.Family (LensLike', set, view)
 import Numeric.Natural (Natural)
 import Prelude hiding (maybe, sequence)
 import System.FilePath (takeDirectory)
 
 import qualified Control.Applicative
 import qualified Control.Exception
+import qualified Control.Monad.Trans.State.Strict as State
 import qualified Data.Foldable
 import qualified Data.Functor.Compose
 import qualified Data.Functor.Product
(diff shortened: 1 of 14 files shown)


src/Dhall.hs
```diff
@@ -282,12 +298,21 @@ inputWithSettings
...
+
+    let EvaluateSettings {..} = _evaluateSettings
+
+    let transform =
+               set Dhall.Import.protocolVersion _protocolVersion
+            .  set Dhall.Import.normalizer      _normalizer
+            .  set Dhall.Import.startingContext _startingContext
+            .  set Dhall.Import.rootDirectory   _rootDirectory
+
+    let status = transform (Dhall.Import.emptyStatus ".")
```
quasicomputational: This doesn't look right. I am not sure why `Status` has grown the `_rootDirectory` field since it's not actually being used (?), but, when that gets set, it doesn't do the right thing since the root import is always going to be relative to `.` since that's what's passed to `emptyStatus`.

Note that the test suite isn't exercising this code, and that the test suite is invoking `emptyStatus` with the root directory rather than `.`.

src/Dhall/Import.hs
```diff
@@ -665,17 +622,17 @@ loadStaticWith from_import ctx n expr₀ = case expr₀ of
...
+                        Right _   -> return (Dhall.Core.normalizeWith (getReifiedNormalizer _normalizer) expr'')
+                    zoom cache (State.put $! Map.insert here expr''' _cache)
                     return expr'''
 
     case hash (importHashed import_) of
         Nothing -> do
             return ()
         Just expectedHash -> do
-            let actualHash = hashExpression expr
+            let actualHash = hashExpression _protocolVersion expr
```
quasicomputational: So the hashes guarding all imports, both direct and transitive, are governed by this one protocol version. Doesn't this risk ossifying the protocol version across the ecosystem?

src/Dhall.hs
Gabriel439: Whoops, good catch.  I'll fix this

src/Dhall/Import.hs
Gabriel439: @quasicomputational: How about the following set of changes:

* Change `decode` to not require a version number
    * Instead, it just infers the version from the binary representation, which already includes the version string
    * In retrospect, I'm not sure why I made it require a `ProtocolVersion` argument :stuck_out_tongue: 
* Change semantic integrity checks to allow the user to optionally specify the protocol version string
    * i.e. `sha256:${HASH}:${PROTOCOL_VERSION}`
    * If the protocol version is unspecified then the interpreter assumes the latest protocol version string
* Remove the `--protocol-version` flag

Do you think that would work?

quasicomputational: That sounds reasonable. The other thing it occurs to me to do would be to allow many hashes and then have the importer pick the one with a version that it knows about, but that's probably putting the cart before the horse and making things too complex.

Any solution here that keeps dependents and dependencies from having to upgrade their hashes in lockstep would work, and, so long as decoders understand a few previous versions of the protocol version, this should work fine.

Gabriel439: Actually, I just realized that adding support for `sha256:${HASH}:${PROTOCOL_VERSION}` would be implementing a new feature without standardizing it.  What I will do in the interim is still remove support for the `--protocol-version` flag as I mentioned before, but wait to add support for the `:${PROTOCOL_VERSION}` extension to the semantic integrity check until it has been standardized.

This implies that I will delay standardizing any change that changes the binary protocol until we have first standardized support for pinning the protocol via the extension to the semantic integrity check.

Gabriel439: Actually, thinking on this more, we probably want to keep the `--protocol-version` flag, for two reasons:

* We minimally require the flag for the `dhall freeze` and `dhall hash` subcommands since the protocol version to use when hashing the expression(s) cannot be inferred from the input

* Even if we do eventually support specifying the protocol version on a per-import basis, we still need to specify a default protocol version when the protocol version is not pinned for an import.  Rather than hard-coded the default to be the latest version I think it makes sense to give users the ability to override the default protocol version from the command line.

So in other words, we'll still allow users to specify the protocol version on a per-import basis once that is standardized, but even after that we probably want a way for users to specify the default protocol version from the command line when otherwise unspecified.

PR #533, opened 2018-08-04
Cache imports protected by a semantic integrity check

Gabriel439:
... as standardized in https://github.com/dhall-lang/dhall-lang/pull/208

For example, given this file:

```haskell
http://prelude.dhall-lang.org/package.dhall sha256:2a84d3c5a420e2549f1dfb145fa2a6956481d99f022a77bc31152c29a008dbfa
```

The first time you interpret the file it is retrieved, parsed, and normalized
like normal:

```bash
$ time dist/build/dhall/dhall <<< './test.dhall'
{ `Bool` :
    { and :
…
    }
}

real    0m6.524s
user    0m0.720s
sys     0m0.158s
```

... and the second time you get a significant speedup by hitting the
local cache:

```bash
$ time dist/build/dhall/dhall <<< './test.dhall'
{ `Bool` :
    { and :
…
    }
}

real    0m0.162s
user    0m0.063s
sys     0m0.070s
```

The main contributors to the speed up are:

* Retrieving a single local file is faster than retrieving multiple
  remote files
* Decoding binary is faster than parsing text
* The expression is cached in normal form, so there is no need to
  re-normalize

### DIFF
--- a/benchmark/parser/Main.hs
+++ b/benchmark/parser/Main.hs
@@ -53,7 +53,9 @@ benchExprFromBytes name bytes = bench name (whnf f bytes)
         term <- case Codec.Serialise.deserialiseOrFail bytes of
             Left  _    -> Nothing
             Right term -> return term
-        Dhall.Binary.decode term
+        case Dhall.Binary.decode term of
+            Left  _          -> Nothing
+            Right expression -> return expression
 
 main :: IO ()
 main = do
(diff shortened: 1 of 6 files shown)


FintanH:
Much excite!

Gabriel439:
Also, the locally cached Prelude is `5 KB`:

```
$ cat ~/.cache/dhall/2a84d3c5a420e2549f1dfb145fa2a6956481d99f022a77bc31152c29a008dbfa | wc -c
5040
```

... and if space ever became an issue we could eventually compress things further:

```
$ cat ~/.cache/dhall/2a84d3c5a420e2549f1dfb145fa2a6956481d99f022a77bc31152c29a008dbfa | gzip | wc -c
1249
```

PR #535, opened 2018-08-04
Stack support

jneira:
* Use lts-12.4 in the default yaml file
* Keep lts-11 in it own config file `stack-lts-11.yaml`, adding the needed extra-deps
* Add them to `stack-lts-6.yaml` too
* Add an import to Data.Monoid.(<>) in `Dhall.Binary` to make it compilable with base < 4.11 (and lts-11)


Gabriel439:

PR #536, opened 2018-08-05
Remove `ansi-terminal` dependency

Gabriel439:
We were only using `ansi-terminal` for a single function, so this change
internally vendors that function to remove the dependency on
`ansi-terminal`


ocharles:
What is the benefit here? You still require ansi-terminal transitively through prettyprint-ansi-terminal anyway.

Gabriel439:
@ocharles: Ah, good point.  Never mind.

PR #541, opened 2018-08-09
Avoid instance derivation error with stack-lts-6

jneira:
Hi, when trying to build master with stack-lts-6.yaml i've got an error:
```
C:\ws\eta\dhall\dhall-haskell\src\Dhall\Binary.hs:765:19:
    No instance for (Exception Text) 
      arising from the first field of ‘UnsupportedProtocolVersionString’
        (type ‘Text’)
    Possible fix: 
      use a standalone 'deriving instance' declaration,
        so you can specify the instance context yourself
    When deriving the instance for (Exception DecodingFailure)
```
I guess this is a bug resolved for ghc > 7.10: https://stackoverflow.com/questions/39379006/what-is-the-difference-between-deriveanyclass-and-an-empty-instance

### DIFF
--- a/src/Dhall/Binary.hs
+++ b/src/Dhall/Binary.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE DeriveAnyClass     #-}
-{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE RecordWildCards    #-}
 {-# LANGUAGE OverloadedStrings  #-}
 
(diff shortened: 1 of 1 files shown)


src/Dhall/Binary.hs
```diff
@@ -761,7 +762,9 @@ data DecodingFailure
     = CannotDecodeProtocolVersionString Term
     | UnsupportedProtocolVersionString Text
     | CBORIsNotDhall Term
-    deriving (Eq, Exception)
+    deriving (Eq)
+
+deriving instance Exception DecodingFailure
```
Gabriel439: I don't think you need `StandaloneDeriving` here.  I think this can be:

```haskell
instance Exception DecodingFailure`
```

jneira: Oh, i followed the ghc error advice without too much thinking, i'll try that simpler option

jneira:
@Gabriel439 much better in that way, thanks!

Gabriel439:
Thank you! 🙂

PR #542, opened 2018-08-09
Set lower bound of directory to 1.2.7.1 to make it compatible with cabal-1.22

jneira:
Hi, in the proccess of set dhall as the default config for etlas we had a clash between the upper bound of directory (< 1.3 trough the etlas dep on cabal 1.22 ) and the lower bound of dhall (>= 1.3) so we had to force `allow-newer:true` in the `stack yaml` for etlas. That flag in stack is global and could cause problems in the resolution later.
I've built dhall with lts-6, lts-12 and cabal and i've not seen any problem with this bound but i wonder if the lower bound 1.3 (although is 1.2.2.0 for doctest) was deliberate.

### DIFF
--- a/dhall.cabal
+++ b/dhall.cabal
@@ -178,7 +178,7 @@ Library
         contravariant                              < 1.6 ,
         cryptonite                  >= 0.23     && < 1.0 ,
         Diff                        >= 0.2      && < 0.4 ,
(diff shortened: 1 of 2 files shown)


Gabriel439:
The lower bound was not deliberate and lowering it is fine

PR #543, opened 2018-08-11
Build inside of source directory produced by `cabal sdist`

Gabriel439:
This ensures that we don't accidentally forget to package data files
when uploading to Hackage to prevent build failures

... as suggested by @ocharles in https://github.com/dhall-lang/dhall-to-cabal/pull/121


ocharles:
Why filter the input to cabal sdist? To try and avoid extraneous recompilation? With my IFD method you don't need to do that (in fact the hash of the cleaned directory is exactly a content hash and nothing more)

Gabriel439:
@ocharles: This is mainly to avoid storing multiple nearly-duplicate copies of the local source directory underneath the `/nix/store`.  I'm referring specifically to the input directory that is provided to `cabal sdist`, not the output directory that it emits.

ocharles:
I believe those will get GCed as there is no root that points to them - they are used only at evaluation time and then nothing references them.

Gabriel439:
@ocharles: The other thing to keep in mind is that if you don't minimally filter out `dist` it can grow quite large (~250 MB on my machine at the moment), so then it takes a while to copy that to the store on each copy and wastes space quickly.

ocharles:
I don't really run nix builds locally so I suppose I don't have that concern - this stuff is only for CI which wouldn't see a large `dist` directory. If you're happy with this, we're all good :+1:

PR #550, opened 2018-08-24
Use System.Directory to get cache dirs

### LINKED ISSUE #549: Cache imports not working in windows by default
Hi, i've tested the [new feature of cache imports](https://github.com/dhall-lang/dhall-haskell/commit/d010464d02836025251b2d061fe93b56625daa3b) using in master version and it seems that it is not working in (at least some versions) of windows.
In Windows 7 the `HOME` environment variable used in the code is not set by default, the more similar one would be `%USERPROFILE%`. 
Setting `HOME` at system configuration level makes it work but no setting it in the dos console.

jneira:
To fix #549

### DIFF
--- a/src/Dhall/Import.hs
+++ b/src/Dhall/Import.hs
@@ -540,13 +540,14 @@ getCacheFile hash = do
                     liftIO (Directory.setPermissions directory private)
 
     let tryXdgCacheHome = do
-            Just cacheDirectory <- liftIO (System.Environment.lookupEnv "XDG_CACHE_HOME")
+            cacheDirectory <- liftIO $ Directory.getXdgDirectory Directory.XdgCache ""
+            
             assertDirectory cacheDirectory
 
             return cacheDirectory
 
     let tryHome = do
-            Just home <- liftIO (System.Environment.lookupEnv "HOME")
+            home <- liftIO Directory.getHomeDirectory
 
             assertDirectory home
 
--- a/src/Dhall/Import.hs
+++ b/src/Dhall/Import.hs
@@ -539,25 +539,12 @@ getCacheFile hash = do
 
                     liftIO (Directory.setPermissions directory private)
 
-    let tryXdgCacheHome = do
+    cacheDirectory <- do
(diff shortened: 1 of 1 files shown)


jneira:
In Windows it picks for xdg cache dir  `%LOCALAPPDATA%`(`C:\Users\user\AppData\Local`)

src/Dhall/Import.hs
```diff
@@ -540,13 +540,14 @@ getCacheFile hash = do
...
 
     let tryXdgCacheHome = do
-            Just cacheDirectory <- liftIO (System.Environment.lookupEnv "XDG_CACHE_HOME")
+            cacheDirectory <- liftIO $ Directory.getXdgDirectory Directory.XdgCache ""
+            
             assertDirectory cacheDirectory
 
             return cacheDirectory
 
     let tryHome = do
```
Gabriel439: @jneira: You can delete the `tryHome` alternative since `Directory.getXdgDirectory` already falls back to `~/.cache` if the `XDG_CACHE_HOME` environment variable is unset

src/Dhall/Import.hs
```diff
@@ -539,26 +539,13 @@ getCacheFile hash = do
 
                     liftIO (Directory.setPermissions directory private)
 
-    let tryXdgCacheHome = do
+    cacheDirectory <- do
```
Gabriel439: You can also simplify this a bit further by inlining the inner `do` block into the outer `do` block.  In other words, instead of this:

```haskell
cacheDirectory <- do
    cacheDirectory <- liftIO $ Directory.getXdgDirectory Directory.XdgCache ""

    assertDirectory cacheDirectory

    return cacheDirectory
```

... you can do this:

```haskell
cacheDirectory <- liftIO $ Directory.getXdgDirectory Directory.XdgCache ""

assertDirectory cacheDirectory
```

jneira: yeah, the nesting do is unnecesary

Gabriel439:
Thanks for doing this! 🙂

PR #555, opened 2018-08-27
Make Lint handle some transitive cases.

clavoie-takt:
As it stands, the linter doesn't completely solve

```
$ cat Lint.dhall 
   let a = 5
in let b = a + 4
in 5
```

because it doesn't eliminate the a:

```
$ stack exec dhall lint < Lint.dhall 
let a = 5 in 5
```

After this fix, it does:

```
$ stack exec dhall lint < Lint.dhall 
5
```


Gabriel439:
Whoops!  Thank you for fixing that 🙂

PR #563, opened 2018-09-06
Add kind polymorphism.

sellout:
I wanted to propose this before I spend too much time on it, so let me know if
it’d be acceptable in some form. (And it would also require a change to the
language spec.)

Here’s an example:

./polyfunctor
```dhall
  λ(j : Kind)
→ λ(k : Kind)
→ λ(c : j → j → Type)
→ λ(d : k → k → Type)
→ λ(f : j → k)
→ { map : ∀(a : j) → ∀(b : j) → c a b → d (f a) (f b) }
```

then
```dhall
./polyfunctor
Type
Type
(λ(a : Type) → λ(b : Type) → a → b)
(λ(a : Type) → λ(b : Type) → a → b)
Optional
```
normalizes to
```dhall
{ map : ∀(a : Type) → ∀(b : Type) → (a → b) → Optional a → Optional b }
```
and
```dhall
./polyfunctor
(Type → Type)
(Type → Type)
(λ(a : Type → Type) → λ(b : Type → Type) → ∀(i : Type) → a i → b i)
(λ(a : Type → Type) → λ(b : Type → Type) → ∀(i : Type) → a i → b i)
```
normalizes to
```dhall
  λ(f : (Type → Type) → Type → Type)
→ { map :
        ∀(a : Type → Type)
      → ∀(b : Type → Type)
      → (∀(i : Type) → a i → b i)
      → ∀(i : Type)
      → f a i
      → f b i
  }
```

### DIFF
--- a/src/Dhall/Binary.hs
+++ b/src/Dhall/Binary.hs
@@ -149,6 +149,8 @@ encode_1_0 (Const Type) =
     TString "Type"
 encode_1_0 (Const Kind) =
     TString "Kind"
+encode_1_0 (Const Sort) =
+    TString "Sort"
 encode_1_0 e@(App _ _) =
     TList ([ TInt 0, f₁ ] ++ map encode_1_0 arguments)
   where
@@ -469,6 +471,8 @@ decode_1_0 (TString "Type") =
     return (Const Type)
 decode_1_0 (TString "Kind") =
     return (Const Kind)
+decode_1_0 (TString "Sort") =
+    return (Const Sort)
 decode_1_0 (TString x) =
     return (Var (V x 0))
 decode_1_0 (TList [ TString x, TInt n ]) =
--- a/src/Dhall/Core.hs
+++ b/src/Dhall/Core.hs
@@ -93,22 +93,25 @@ import qualified Data.Text.Prettyprint.Doc  as Pretty
 
 {-| Constants for a pure type system
 
-    The only axiom is:
+    The axioms are:
 
 > ⊦ Type : Kind
+> ⊦ Kind : Sort
 
     ... and the valid rule pairs are:
 
 > ⊦ Type ↝ Type : Type  -- Functions from terms to terms (ordinary functions)
 > ⊦ Kind ↝ Type : Type  -- Functions from types to terms (polymorphic functions)
 > ⊦ Kind ↝ Kind : Kind  -- Functions from types to types (type constructors)
+> ⊦ Sort ↝ Kind : Kind  -- Functions from kinds to types (polymorphic type constructors)
+> ⊦ Sort ↝ Sort : Sort  -- Functions from kinds to kinds (kind constructors)
 
     These are the same rule pairs as System Fω
 
     Note that Dhall does not support functions from terms to types and therefore
     Dhall is not a dependently typed language
 -}
-data Const = Type | Kind deriving (Show, Eq, Data, Bounded, Enum, Generic)
+data Const = Type | Kind | Sort deriving (Show, Eq, Data, Bounded, Enum, Generic)
 
 instance Pretty Const where
     pretty = Pretty.unAnnotate . prettyConst
--- a/src/Dhall/Pretty/Internal.hs
+++ b/src/Dhall/Pretty/Internal.hs
@@ -334,6 +334,7 @@ prettyChunks (Chunks a b) =
 prettyConst :: Const -> Doc Ann
 prettyConst Type = builtin "Type"
 prettyConst Kind = builtin "Kind"
+prettyConst Sort = builtin "Sort"
 
 prettyVar :: Var -> Doc Ann
 prettyVar (V x 0) = label (Pretty.unAnnotate (prettyLabel x))
--- a/src/Dhall/TypeCheck.hs
+++ b/src/Dhall/TypeCheck.hs
@@ -54,15 +54,19 @@ traverseWithIndex_ k xs =
 
 axiom :: Const -> Either (TypeError s a) Const
 axiom Type = return Kind
-axiom Kind = Left (TypeError Dhall.Context.empty (Const Kind) Untyped)
+axiom Kind = return Sort
+axiom Sort = Left (TypeError Dhall.Context.empty (Const Sort) Untyped)
 
 rule :: Const -> Const -> Either () Const
+rule Type Type = return Type
+rule Kind Type = return Type
+rule Sort Type = return Type
+rule Kind Kind = return Kind
+rule Sort Kind = return Kind
+rule Sort Sort = return Sort
 -- This forbids dependent types. If this ever changes, then the fast
 -- path in the Let case of typeWithA will become unsound.
-rule Type Kind = Left ()
-rule Type Type = return Type
-rule Kind Kind = return Kind
-rule Kind Type = return Type
+rule _    _    = Left ()
 
 {-| Type-check an expression and return the expression's type if type-checking
     succeeds or an error if type-checking fails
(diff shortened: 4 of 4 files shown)


sellout:
Just to have a non-endofunctor:
```dhall
./polyfunctor
(Type -> Type)
Type
(λ(a : Type → Type) → λ(b : Type → Type) → ∀(i : Type) → a i → b i)
(λ(a : Type) → λ(b : Type) → a → b)
```
normalizes to
```dhall
  λ(f : (Type → Type) → Type)
→ { map :
        ∀(a : Type → Type)
      → ∀(b : Type → Type)
      → (∀(i : Type) → a i → b i)
      → f a
      → f b
  }
```
(which is the “kind” of thing I would use in dada for mapping over fixed-point operators).

Gabriel439:
@sellout: Yes, I'll accept this.  The main benefit of this (which you appear to have noticed while implementing this) is that it allows records of types with mixed kinds (i.e. `{ foo = List, bar = Bool }`) since they can now be encoded using kind-polymorphism.  That in turns means that most packages will only need to export at most two records: one for terms and one for types.

It also doesn't affect existing user code

Profpatsch:
> is that it allows records of types with mixed kinds (i.e. { foo = List, bar = Bool }) since they can now be encoded using kind-polymorphism

If we can do that step, what’s the problem with doing the same for values as well? That the type system would need to become dependent in the process?

Gabriel439:
@Profpatsch: No, separating types and terms in records is independent of whether or not the type system is dependent.  In other words, even if we turned on dependent types we would still need to keep types and terms separate.

Basically all of Dhall's higher-level features (like builtins, records, lists, unions, etc.) can be encoded in terms of lower-level features (i.e. pure lambda calculus) using Dhall's own type system.

For example, a record of the form:

```haskell
{ foo = 1, bar = True }
```

... could be encoded like this:

```haskell
λ(result : Type) → λ(makeRecord : ∀(foo : Natural) → ∀(bar : Bool) → result) → makeRecord 1 True
```

... and accessing a field of a record like that:

```haskell
λ(r : { foo : Natural, bar : Bool }) → r.foo
```

... could be encoded like this:

```haskell
  λ(r : ∀(result : Type) → (∀(foo : Natural) → ∀(bar : Bool) → result) → result)
→ r Natural (λ(foo : Natural) → λ(bar : Bool) → foo)
```

The encoding might be very clumsy to use and might be less efficient, but it's theoretically possible to translate all of Dhall's features to raw System Fω.

The reason these encodings are important is that they establish that Dhall's higher-level features are still sound and non-Turing-complete since they can be translated to a non-Turing-complete language (System Fω in this case).

The reason this feature is sometimes confused with dependent types is because if you don't have a sound way to translate records to the underlying lambda calculus then records become a backdoor that you can use to break the type system.  One example of that breaking the type system was using records to make code accidentally dependently typed, but that's not the only possible way that it can break the type system.

As @sellout just mentioned, if you want to have records that mix types and terms, then what you actually need is universe polymorphism in order to be able to encode records of types and terms in lambda calculus.  Universe polymorphism means that instead of `Type`, `Kind`, and `Sort` we would have `Type 0`, `Type 1`, `Type 2`, ... or more generally we could parametrize the constant by a natural number.  If you have that then you gain the ability to mix records of types and terms in a sound way.  If I remember correctly, a record like this:

```haskell
{ foo = 1, bar = Natural }
```

... would translate to:

```haskell
λ(n : Natural) → λ(result : Type n) → λ(makeRecord : ∀(foo : Natural) →∀(bar : Type) → result) →makeRecord 1 Natural
```

@sellout: I'm not sure about that yet.  Let me spend some time reasoning through how it's encoded and how that would affect the type-checking rules for records (and how to do it in a way that doesn't break existing user code).

Gabriel439:
@sellout: So the issue with using universe polymorphism is that you can't hide it from the user when type-checking record types.  The issue you run into is that the type of a record type would end up being polymorphic over the universe and you wouldn't be able to hide that universe-polymorphism from the end user for a mixed record of terms and types.

For example, under the current type system, the type of the type of a record of terms is `Type` and the type of the type of a record of types is `Kind`.  If you introduce universe polymorphism then the type of a mixed record of terms and types would be `∀(n : Natural) → Type n`, which means that the user would need to explicitly instantiate the record ahead of time to a specific universe in order to use it anywhere that expected either a `Type 0` or a `Type 1`.  Also, once you instantiated it to such a universe you would only be able to extract either terms (if you instantiated it to `Type 0`) or types (if you instantiated it to `Type 1`) but no longer both.

Let me motivate this with a concrete example with the following type-level function:

```haskell
-- ./Map.dhall
--
-- Simulate a map as an association list

λ(type : { key : Type, value : Type }) → List { key : type.key, value : type.value }
```

```
$ dhall <<< './Map.dhall { key = Text, value = Natural }'
List { key : Text, value : Natural }
```

The interpreter (correctly) does not flag `./Map.dhall` as a dependent type because the type system infers that the type of `{ key : Type, value : Type }` is `Kind` so the `./Map.dhall` is treated as an ordinary type constructor that just happens to permit initializing its type variables using record syntax instead of positional type arguments.

However, now imagine that you have a record with these contents:

```haskell
{ key = Text, value = Natural, x = 0 }
  : { key : Type, value : Type, x : Natural }
  : ∀(n : Natural) → Type n
```

... which could potentially be used as either a record of terms or types.  The user would need to explicitly instantiate it to a record of types before they could pass it to `./Map.dhall` so that you don't accidentally create a dependent-types backdoor where they could thread terms like the field `x` into the type (not that dependent types are bad, but this is an example of what could accidentally could go wrong if you don't make the universe explicit for such a mixed record).  More generally, if you don't require the user to be explicit about the choice of universe for mixed records then I believe a similar trick could be used as a backdoor to synthesize the unsound `Type n : Type n`.

So I think the best we'll be able to do for the foreseeable future is your original proposal to add `Sort` and then for users to export one record for terms and another record for types.  I can then document the two-export idiom in the Dhall wiki as a work-around for the type system limitation.

There's another related issue I considered alongside this one, which is the proposal you raised earlier for dependent records.  I think that even if we don't support universe polymorphism it doesn't preclude the possibility of dependent records.  The reason why is that in the vast majority of use cases for dependent records (i.e. modules with scoped access to the ordered record fields) you usually always want the final result of the scoped access to be a term (or another module), so the type of the type of a dependent record can be fixed to `Type` and you can treat it as a record of terms for the purposes of passing the module around opaquely.

sellout:
> If you introduce universe polymorphism then the type of a mixed record of terms and types would be `∀(n : Natural) → Type n`

Is this necessarily true? I may be out of my depth here, but I was thinking that effectively `Type` would be an alias for `U 0`, and `Kind` for `U 1`, and just as
```dhall
{ key : Type, value : Type, x : Natural } : Kind
```
isn’t valid with kind polymorphism, universe polymorphism wouldn’t necessarily allow
```dhall
{ key : U 0, value : U 0, x : Natural } : U 1
```

With universe polymorphism, I was hoping to be able to to do something roughly like
```dhall
-- pattern functor for lists
  λ(level : Level)
→ λ(a : U level)
→ λ(b : U level)
→   < Nil : ({} : U level)
    | Cons : { head : a, tail : b }
    >
  : U level
```
to enable HLists, etc.

Gabriel439:
@sellout: So I looked into this more closely and it looks like in Agda the way they do this is that the universe level represents an upper bound on the type of the type of the value that you can extract from a record.  Specifically, if the type of the type of `A` is `Universe m` and the type of the type of `B` is `Universe n` then the type of the type of `A x B` is `Universe (max m n)`.  That implies that the type of the type of a record is the type of the type of the "largest" field that you could possibly extract.

However, I think that if you want this for heterogeneous lists then it sounds like what you want is metaprogramming which I think would be solved better by access to the abstract syntax tree rather than implementing it with the language.  The reason why is that if you implement heterogeneous lists within the language you are going to lose all the syntactic niceties and features of Dhall records which you could keep if you access the record structure through the Haskell API.

phadej:
I think this is bad idea.

```haskell
% cat example.dhall
λ(j : Kind) → λ(a : Kind → Kind) → λ(b : Kind) → a (a b)

% cat example.dhall | dhall type
∀(j : Kind) → ∀(a : Kind → Kind) → ∀(b : Kind) → Kind
```

https://en.wikipedia.org/wiki/System_U#Girard's_paradox

> The definitions of System U and U− allow the assignment of polymorphic kinds to generic constructors in analogy to polymorphic types of terms in classical polymorphic lambda calculi, such as System F. An example of such a generic constructor might be[2]:353 (where k denotes a kind variable)

*as above*

> This mechanism is sufficient to construct a term with the type `forall (p : Type) -> p`, which implies that every type is inhabited.

---

*A simplification of Girard's paradox* by Antonius J. C. Hurkens
have an example of the bottom value. I tried to make it, but I'm too tired, here's the start

```haskell
   let bottom = forall (any : Type) -> any
in let not = \(p : Type) -> bottom
in let pow = \(X : Kind) -> X -> Type
in let U = forall (X : Kind) -> (pow (pow X) -> X) -> pow (pow X)

in let tau : pow (pow U) -> U
         = \(t : pow (pow U)) ->
           \(X : Kind) ->
           \(f : pow (pow X) -> X) ->
	   \(p : pow X) ->
	   (t (\(x : U) -> p (f ((x X) f))))

in let sigma : U -> pow (pow U)
             = \(s : U) ->
               ((s U) (\(t : pow (pow U)) -> tau t))

in let Delta : pow U
             = \(y : U) ->
               not (forall (p : pow U) -> ((sigma y) p) -> (p (tau (sigma y))))

in let Omega : U
             = tau (\ (p : pow U) -> forall (x : U) -> (((sigma x) p) -> p x))

in \(t : pow (pow U)) -> sigma (tau t)
```

Gabriel439:
@phadej: @sellout's proposal does not entail Girard's paradox.  To be precise, neither kind polymorphism nor universe polymorphism imply Girard's paradox.

I'm familiar with the issues with System U and `Type : Type`, but that's not what is being proposed here.

sellout:
Now that the spec change is merged (dhall-lang/dhall-lang#238), what’s the the plan for this? More changes to be made?

I’ve written tons of code using this branch:
- [dhall-bhat](https://github.com/sellout/dhall-bhat/tree/kind-polymorphism)
- [dada](https://github.com/sellout/dada/tree/kind-polymorphism)

sellout:
And I noticed it just failed … but I can’t figure out what I broke from the log.

sellout:
@f-f Good point. I can copy over some stuff from the other repos I linked.

Gabriel439:
@sellout: Once the corresponding change to the spec is merged you have the greenlight to merge this.  As far as I can tell the only outstanding request is to add tests like @f-f suggested

Gabriel439:
I ran into one issue when testing this branch, which is that a record like `{ x = List }` still does not type-check.  However, I realized that technically the standard had the same issue so fixing it requires a change to the standard:

https://github.com/dhall-lang/dhall-lang/pull/241

Gabriel439:
This looks good to me.  Did you have any other changes planned?

sellout:
@Gabriel439 Yeah, I noticed (when writing tests) that `Sort` doesn’t parse as a reserved word. I think I lost that when the parsing restructuring went in. But adding the obvious places didn’t fix that (it still says “Error: Unbound variable: Sort”), so I gotta sit and stare a bit tonight. But I’ll merge it today, with some new tests.

Gabriel439:
@sellout: If you're busy, just add a failing test and I can fix things so that the test passes

sellout:
@Gabriel439 Ok, I pushed the failing test, along with other stuff.

PR #570, opened 2018-09-11
Add `--ascii` flag

### LINKED ISSUE #562: Don't auto replace with lambdas
We are starting to use dhall for creating configurations for our customers.

It is so good!  Thanks for making it.  

One thing I worry about being intimidating is the λ 's that appear everwhere as a reformat for `\`.

If there could be a reformat option to disable this replacement I think it would be good for people trying to use dhall with a non-fp audience.  

I would be glad to help with it if someone can point me to the relevant section perhaps?

Gabriel439:
Fixes #562

This flag changes causes the pretty-printer to render ASCII instead of
Unicode


quasicomputational:
That's a pretty fearsome diff in the pretty-printer. I've glanced at it and it all looks sensible to me, though this code ought to be very amenable to tests and I guess some smoke checks wouldn't hurt?

src/Dhall/Main.hs
```diff
@@ -95,23 +96,17 @@ parseOptions :: Parser Options
 parseOptions =
         Options
     <$> parseMode
-    <*> parseExplain
-    <*> parsePlain
+    <*> switch "explain" "Explain error messages in more detail"
+    <*> switch "plain" "Disable syntax highlighting"
+    <*> switch "ascii" "Format code using only ASCII characters"
```
PierreR: Something like `flag Unicode ASCII` might be more idiomatic ?

basile-henry: `Text` literals are still formatted in `Unicode` right?

tests/format/asciiB.dhall
```diff
@@ -0,0 +1 @@
+\(a : Type) -> forall (b : a) -> a
```
basile-henry: Does this test work with a missing newline at the end of the file?

src/Dhall/Main.hs
Gabriel439: Yes

tests/format/asciiB.dhall
Gabriel439: No.  We have to strip trailing newlines because it looks for an exact match with the rendered string (which doesn't have a trailing newline)

src/Dhall/Main.hs
basile-henry: It didn't seem obvious to me from the description. But I guess you wouldn't really be able to make the `Text` literals `ascii` only anyway.
If I'm the only one who thought the description was weird leave it like as it is! :smile:

Gabriel439: Would "Format code using only ASCII syntax" have been more clear?

basile-henry: How about "Format code using ASCII syntax for Dhall built-ins"? Or is that too long?

tests/format/asciiB.dhall
sellout: @Gabriel439 Would it be ok to change the test matcher to append a newline to the rendered string? Or, equivalently, to strip the last character from the loaded file? It would be nice for the test files to have the usual final newline.

src/Dhall/Main.hs
Gabriel439: The issue with "built-in" is that it means either built-in expressions like `Natural/even` or `Optional/fold` or built-in operators like `+`, but not syntactic elements like `λ`/`→`

basile-henry: In that case your version with "syntax" instead of "characters" is better. Sorry for being so picky about the wording. 😅

PR #571, opened 2018-09-11
Change `Prelude/` directory to be a submodule

Gabriel439:
... as suggested by @f-f

Now that there is a separate git repository for the Prelude, we can use it
as a submodule instead of mirroring the files between the two repositories


Gabriel439:
Hmmm.  It looks like the use of submodules causes evaluation errors under the current Nix logic:

```
command `nix-prefetch-git /var/lib/hydra/scm/git/12a6923f1f6a30e5b8cd6617716519885ff93d2bb618d20552eb1e3d5de8a621 03d56ed227ee199bb8fcb257985164193e2cedda' failed with exit status 256 at /nix/store/rak0ad5jqs1jy20kaq8zws22044blh6d-hydra-2017-11-21/libexec/hydra/lib/Hydra/Helper/Nix.pm line 428.
```

I might have to postpone this until I figure out how to fix that problem

ocharles:
We use sub modules just fine. Are you sure the submodule commit you're referring to actually exists on the remote?

Gabriel439:
@ocharles: Is that separate from pushing the branch?

ocharles:
Nope, it looks like everything is correct here, as I can follow the commit via GitHub's UI. I would follow `journalctl -u hydra-evaluator -u hydra-queue-runner -f`. IIRC, there's a log message that shows up there that doesn't show up in the web UI. Should shine some light on why `nix-prefetch-git` is failing.

Gabriel439:
Closing this in favor of making the Prelude a remote import as suggested in https://github.com/dhall-lang/dhall-lang/issues/240#issuecomment-427471013

Issue #568, opened 2018-09-09
Make tests more language-agnostic


f-f:
I'd love to to possibly reuse *all* the tests from this implementation in [dhall-clj](https://github.com/f-f/dhall-clj), and in the end have a kind-of-agnostic shared test-suite to validate implementations on.
(current status: I'm using all the normalization an typecheck tests, and it's working great)

If this sounds like a good idea, I would like to:
- contribute new tests here ([I'm monitoring the test coverage](https://coveralls.io/github/f-f/dhall-clj?branch=master), and some places - e.g. the typechecker - are a bit uncovered). 
  Related: should we track coverage also here?
- adapt the current tests to be more agnostic. This includes stuff like:
  - applying existing conventions all over, like in #567
  - move tests from Haskell source to Dhall files - I noticed some tests check against an AST built in code instead of parsing it from source
  - make existing tests more precise: e.g. the `parser` tests are either "parse" or "not parse", while it would be helpful to apply the  `A/B` convention and have a "correct" parse to check against.
    A possible way to do this would be to: dump the AST to file as outputted by the Haskell implementation to the `B` files, and use it in the Haskell tests by `read`ing it. Then other implementations can "parse" this AST (it's basically a lisp, acceptably easy to parse IMO) and reuse it.

joneshf:
> move tests from Haskell source to Dhall files - I noticed some tests check against an AST built in code instead of parsing it from source

If you mean stuff like: https://github.com/dhall-lang/dhall-haskell/blob/b5c005ee2bd83f40deaec52e8a472c876d9db572/tests/Normalization.hs#L208-L231

I think that's checking contexts in Haskell. I think `simpleCustomization` is validating normalization for a single addition to the context. I think `nestedReduction` is validating normalization can "see through" multiple additions to the context. Assuming that's a correct interpretation, is there a way to check the spirit of those tests without using the host language (Haskell)?

I think what these tests are showing is that there are different categories of testing happening (sometimes in the same file). Maybe more changes like you did in #567 can help flesh out the categories.

Gabriel439:
I think the language-agnostic test files should probably be relocated to the `dhall-lang` repository since those can be viewed as part of the standard since they are conformance tests.  The exception would be test files for features that have not been standardized (like formatting code)

[...]

PR #585, opened 2018-09-17
Make fold tests unambiguous wrt associativity

f-f:
As noted in https://github.com/dhall-lang/dhall-haskell/issues/568#issuecomment-421777944

### DIFF
--- a/tests/normalization/examples/List/fold/1B.dhall
+++ b/tests/normalization/examples/List/fold/1B.dhall
@@ -1,1 +1,1 @@
-λ(nil : Natural) → 2 + 3 + 5 + nil
+λ(nil : Natural) → 2 + (3 + (5 + nil))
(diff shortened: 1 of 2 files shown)


Gabriel439:
@f-f: These tests are based on the example interpreter outputs from the Prelude documentation, such as:

https://github.com/dhall-lang/Prelude/blob/4e565d6ed6a5bcb5a8afe1f6829fec59cf34f829/Natural/fold#L13

You probably want to keep these files the same (since they need to match the exact output of the interpreter in the example), but instead add a new conformance test for associativity

f-f:
@Gabriel439 +1 for adding new tests to test the new associativity

However, I'm not sure these files can stay as they are when we make operators left-associative. 
The reason is that the existing (unchanged) test file will be parsed/normalized to the AST of `λ(zero : Natural) → 125 * zero`, which is not the same as the AST of `λ(zero : Natural) → 5 * (5 * (5 * zero))`, so these tests will fail

So we'd have to either:
- use `foldl` instead of `foldr` when we fold, so we get the same ASTs again
- output parentheses when the operations are supposed to be right-associative (like in this case with `foldr`), so we'll get these new test files as output

f-f:
Cool 👍 
I'll close this one

PR #595, opened 2018-09-20
Build on older versions of GHC

Gabriel439:
CI only tests GHC 8.4, so it missed the fact that `Dhall.Util` does not build
on GHC 8.0 or GHC 8.2 due to missing a `Data.Monoid` import, which this fixes.

I also plan to separately fix CI to test building on GHC 8.0 to prevent this from
recurring

### DIFF
--- a/src/Dhall/Util.hs
+++ b/src/Dhall/Util.hs
@@ -7,6 +7,7 @@ module Dhall.Util
     , snipDoc
     ) where
 
+import Data.Monoid ((<>))
 import Data.Text (Text)
 import Data.Text.Prettyprint.Doc (Doc)
 import Dhall.Pretty (Ann)


jneira:
Would be possible to test ghc 7.10.3? Eta is built with this version...

Gabriel439:
@jneira: I just tried this yesterday and the problem I ran into is that at least one of Dhall's transitive dependencies (`basement`) only builds on GHC 8.0 or newer.  I think that probably the more reliable way to verify this is to build with Eta directly if we get `etaPackages` working because that's a more accurate check than building with GHC 7.10.

jneira:
mmm that would be nice too, but i am afraid that etlas are built with ghc 7.10.3 and now it has dhall as transitive dependency. 
Eta itself is 7.10.3 but it has some extensions from ghc >= 8 and base <= 4.11.1.  It would be possible then that eta could build dhall but ghc 7.10 could not build dhall for etlas. 🤔
So far we have been able to build etlas with dhall using [this stack file](https://github.com/typelead/etlas/blob/master/stack.yaml), that uses `basement-0.0.6`.

Gabriel439:
@jneira: I see what you mean now.  Yeah, in that case, I'll try to get it to build against `ghc-7.10` then and pin `basement` to an older version

PR #599, opened 2018-09-26
Allow template-haskell-2.15 (GHC-8.6)

phadej:


phadej:
Note: there **is** an install plan with GHC-8.6.1 which compiles and passes tests just fine. `containers-0.5.5.11` is installable. http://hackage.haskell.org/package/containers

PR #601, opened 2018-09-26
Compress operator parsing hierarchy

Gabriel439:
This groups operators that work on different types into a single precedence
level to improve parsing performance.  Parsing the precedence of these groups
of operators incorrectly doesn't matter since any expression that mixes them is
ill-typed

This improves performance by about 15% on the benchmark expression from #108

Before:

```
time                 334.4 ms   (308.6 ms .. 375.3 ms)
                     0.998 R²   (0.995 R² .. 1.000 R²)
mean                 349.8 ms   (340.4 ms .. 364.3 ms)
std dev              14.35 ms   (5.761 ms .. 19.71 ms)
```

After:

```
time                 289.9 ms   (274.6 ms .. 297.9 ms)
mean                 309.3 ms   (301.0 ms .. 318.3 ms)
std dev              11.08 ms   (8.083 ms .. 14.31 ms)
```

### DIFF
--- a/src/Dhall/Parser/Expression.hs
+++ b/src/Dhall/Parser/Expression.hs
@@ -158,67 +158,61 @@ nonEmptyOptional embedded = do
     return (OptionalLit b (pure a))
 
 operatorExpression :: Parser a -> Parser (Expr Src a)
-operatorExpression = importAltExpression
+operatorExpression = precedence0Expression
 
 makeOperatorExpression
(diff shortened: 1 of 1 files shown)


Gabriel439:
Note that I also tried using `Text.Parser.Expression.buildOperatorExpression` as well and that made things slower (even slower than the original code)

Gabriel439:
This also improves heap residency, which goes down from ~5.5 MB to ~4 MB

PR #603, opened 2018-09-27
Introduce InputText

phadej:
InputText (something to be parsed) and InputChunk (something parsed)
help to differentiate from all other Text we have

As a bonus this allows to experiment with different underlying
representations.

Operationally there shouldn't be any differences, but one:
`readInputText` doesn't care about users' LOCALE settings,
and assumes input to be UTF8 _always_.
That's different from `Data.Text.readTextIO`

Otherwise, this patches adds a lot differently named `coerce`,
so performance shouldn't be affected, given GHC does good job.

### DIFF
--- a/dhall.cabal
+++ b/dhall.cabal
@@ -233,6 +233,7 @@ Library
         Dhall.Parser.Expression,
         Dhall.Parser.Combinators,
         Dhall.Parser.Token,
+        Dhall.Parser.InputText,
         Dhall.Import.Types,
         Dhall.Util,
         Paths_dhall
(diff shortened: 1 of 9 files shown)


Gabriel439:
What optimizations do we expect to be able to obtain from differentiating between `InputText` and `InputChunk`?  The reason I ask is because I don't want to prematurely abstract without at least a motivating use case

phadej:
It might be beneficial to keep `InputChunk = Text` or `= [Char]` or `data InputChunk !Int [Char]`, but change `InputText = U.Vector Char`

phadej:
They also have different instances. `Semigroup InputText` is semi-code-smell.

Gabriel439:
@phadej: How about just `InputText` for now since we have a motivating use case for that already and then we can add `InputChunk` if we find that another representation is more optimal for that?

PR #605, opened 2018-09-27
Try parens as first primitive

phadej:
`((((((((((((((x))))))))))))))` is still slow, but now less  (from 2.0s -> 1.6s) on my machine


Gabriel439:
I have the general fix up here: https://github.com/dhall-lang/dhall-haskell/pull/606

PR #612, opened 2018-10-01
Add new `Dhall.Map` module

Gabriel439:
This change drops the dependency on `insert-ordered-containers` in favor of a
new `Dhall.Map` module which implements a limited subset of the functionality we
need.  In addition to simplifying the dependency tree this also improves
the performance of the `deep-nested-large-record` benchmark by 2.4x:

Before:

```
time                 1.412 s    (1.388 s .. 1.436 s)
mean                 1.398 s    (1.387 s .. 1.404 s)
std dev              10.02 ms   (2.502 ms .. 13.39 ms)
```

After:

```
time                 582.2 ms   (552.5 ms .. 628.4 ms)
mean                 565.9 ms   (560.3 ms .. 574.6 ms)
std dev              8.901 ms   (1.218 ms .. 11.75 ms)
```

### DIFF
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -38,7 +38,7 @@ module Dhall
 
     -- * Types
     , Type(..)
-    , RecordType
+    , RecordType(..)
     , InputType(..)
     , Interpret(..)
     , InvalidType(..)
@@ -67,7 +67,7 @@ module Dhall
 
     , Inject(..)
     , inject
-    , RecordInputType
+    , RecordInputType(..)
     , inputFieldWith
     , inputField
     , inputRecord
@@ -91,7 +91,6 @@ import Control.Monad.Trans.State.Strict
 import Data.Functor.Contravariant (Contravariant(..), (>$<))
 import Data.Functor.Contravariant.Divisible (Divisible(..), divided)
 import Data.Scientific (Scientific)
-import Data.Semigroup
 import Data.Sequence (Seq)
 import Data.Text (Text)
 import Data.Typeable (Typeable)
@@ -711,7 +710,7 @@ unit = Type extractOut expectedOut
         | Data.Foldable.null fields = return ()
     extractOut _ = Nothing
 
-    expectedOut = Record Dhall.Map.empty
+    expectedOut = Record mempty
 
 {-| Decode a `String`
 
@@ -864,7 +863,7 @@ instance GenericInterpret V1 where
       where
         extract _ = Nothing
 
-        expected = Union Dhall.Map.empty
+        expected = Union mempty
 
 instance (Constructor c1, Constructor c2, GenericInterpret f1, GenericInterpret f2) => GenericInterpret (M1 C c1 f1 :+: M1 C c2 f2) where
     genericAutoWith options@(InterpretOptions {..}) = pure (Type {..})
@@ -1117,9 +1116,9 @@ instance Inject Double where
 instance Inject () where
     injectWith _ = InputType {..}
       where
-        embed = const (RecordLit Dhall.Map.empty)
+        embed = const (RecordLit mempty)
 
-        declared = Record Dhall.Map.empty
+        declared = Record mempty
 
 instance Inject a => Inject (Maybe a) where
     injectWith options = InputType embedOut declaredOut
@@ -1269,9 +1268,9 @@ instance (GenericInject f, GenericInject g) => GenericInject (f :*: g) where
 instance GenericInject U1 where
     genericInjectWith _ = pure (InputType {..})
       where
-        embed _ = RecordLit Dhall.Map.empty
+        embed _ = RecordLit mempty
 
-        declared = Record Dhall.Map.empty
+        declared = Record mempty
 
 instance (Selector s, Inject a) => GenericInject (M1 S s (K1 i a)) where
     genericInjectWith opts@(InterpretOptions {..}) = do
@@ -1327,7 +1326,7 @@ newtype RecordType a =
   RecordType
     ( Data.Functor.Product.Product
         ( Control.Applicative.Const
-            ( MonoidMap
+            ( Dhall.Map.Map
                 Text
                 ( Expr Src X )
             )
@@ -1340,19 +1339,10 @@ newtype RecordType a =
     )
   deriving (Functor, Applicative)
 
-newtype MonoidMap k v = MonoidMap (Dhall.Map.Map k v)
-
-instance Ord k => Data.Semigroup.Semigroup (MonoidMap k v) where
-    MonoidMap l <> MonoidMap r = MonoidMap (Dhall.Map.union l r)
-
-instance Ord k => Monoid (MonoidMap k v) where
-    mempty = MonoidMap Dhall.Map.empty
-
-    mappend = (<>)
 
 -- | Run a 'RecordType' parser to build a 'Type' parser.
 record :: RecordType a -> Dhall.Type a
-record ( RecordType ( Data.Functor.Product.Pair ( Control.Applicative.Const ( MonoidMap fields ) ) ( Data.Functor.Compose.Compose extractF ) ) ) =
+record ( RecordType ( Data.Functor.Product.Pair ( Control.Applicative.Const fields ) ( Data.Functor.Compose.Compose extractF ) ) ) =
   Type
     { extract =
         extractF
@@ -1376,11 +1366,9 @@ field key valueType =
     RecordType
       ( Data.Functor.Product.Pair
           ( Control.Applicative.Const
-              ( MonoidMap
-                  ( Dhall.Map.singleton
-                      key
-                      ( Dhall.expected valueType )
-                  )
+              ( Dhall.Map.singleton
+                  key
+                  ( Dhall.expected valueType )
               )
           )
           ( Data.Functor.Compose.Compose extractBody )
@@ -1451,12 +1439,10 @@ instance Contravariant RecordInputType where
 instance Divisible RecordInputType where
   divide f (RecordInputType bInputTypeRecord) (RecordInputType cInputTypeRecord) =
       RecordInputType
-          (Dhall.Map.union
-              ((contramap $ fst . f) <$> bInputTypeRecord)
-              ((contramap $ snd . f) <$> cInputTypeRecord)
-          )
-
-  conquer = RecordInputType Dhall.Map.empty
+    $ Dhall.Map.union
+      ((contramap $ fst . f) <$> bInputTypeRecord)
+      ((contramap $ snd . f) <$> cInputTypeRecord)
+  conquer = RecordInputType mempty
 
 inputFieldWith :: Text -> InputType a -> RecordInputType a
 inputFieldWith name inputType = RecordInputType $ Dhall.Map.singleton name inputType
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -38,7 +38,7 @@ module Dhall
 
     -- * Types
     , Type(..)
-    , RecordType(..)
+    , RecordType
     , InputType(..)
     , Interpret(..)
     , InvalidType(..)
@@ -67,7 +67,7 @@ module Dhall
 
     , Inject(..)
     , inject
-    , RecordInputType(..)
+    , RecordInputType
     , inputFieldWith
     , inputField
     , inputRecord
@@ -92,6 +92,7 @@ import Data.Functor.Contravariant (Contravariant(..), (>$<))
 import Data.Functor.Contravariant.Divisible (Divisible(..), divided)
 import Data.Monoid ((<>))
 import Data.Scientific (Scientific)
+import Data.Semigroup
 import Data.Sequence (Seq)
 import Data.Text (Text)
 import Data.Typeable (Typeable)
@@ -711,7 +712,7 @@ unit = Type extractOut expectedOut
         | Data.Foldable.null fields = return ()
     extractOut _ = Nothing
 
-    expectedOut = Record mempty
+    expectedOut = Record Dhall.Map.empty
 
 {-| Decode a `String`
 
@@ -864,7 +865,7 @@ instance GenericInterpret V1 where
       where
         extract _ = Nothing
 
-        expected = Union mempty
+        expected = Union Dhall.Map.empty
 
 instance (Constructor c1, Constructor c2, GenericInterpret f1, GenericInterpret f2) => GenericInterpret (M1 C c1 f1 :+: M1 C c2 f2) where
     genericAutoWith options@(InterpretOptions {..}) = pure (Type {..})
@@ -1117,9 +1118,9 @@ instance Inject Double where
 instance Inject () where
     injectWith _ = InputType {..}
       where
-        embed = const (RecordLit mempty)
+        embed = const (RecordLit Dhall.Map.empty)
 
-        declared = Record mempty
+        declared = Record Dhall.Map.empty
 
 instance Inject a => Inject (Maybe a) where
     injectWith options = InputType embedOut declaredOut
@@ -1269,9 +1270,9 @@ instance (GenericInject f, GenericInject g) => GenericInject (f :*: g) where
 instance GenericInject U1 where
     genericInjectWith _ = pure (InputType {..})
       where
-        embed _ = RecordLit mempty
+        embed _ = RecordLit Dhall.Map.empty
 
-        declared = Record mempty
+        declared = Record Dhall.Map.empty
 
 instance (Selector s, Inject a) => GenericInject (M1 S s (K1 i a)) where
     genericInjectWith opts@(InterpretOptions {..}) = do
@@ -1327,7 +1328,7 @@ newtype RecordType a =
   RecordType
     ( Data.Functor.Product.Product
         ( Control.Applicative.Const
-            ( Dhall.Map.Map
+            ( MonoidMap
                 Text
                 ( Expr Src X )
             )
@@ -1340,10 +1341,19 @@ newtype RecordType a =
     )
   deriving (Functor, Applicative)
 
+newtype MonoidMap k v = MonoidMap (Dhall.Map.Map k v)
+
+instance Ord k => Data.Semigroup.Semigroup (MonoidMap k v) where
+    MonoidMap l <> MonoidMap r = MonoidMap (Dhall.Map.union l r)
+
+instance Ord k => Monoid (MonoidMap k v) where
+    mempty = MonoidMap Dhall.Map.empty
+
+    mappend = (<>)
 
 -- | Run a 'RecordType' parser to build a 'Type' parser.
 record :: RecordType a -> Dhall.Type a
-record ( RecordType ( Data.Functor.Product.Pair ( Control.Applicative.Const fields ) ( Data.Functor.Compose.Compose extractF ) ) ) =
+record ( RecordType ( Data.Functor.Product.Pair ( Control.Applicative.Const ( MonoidMap fields ) ) ( Data.Functor.Compose.Compose extractF ) ) ) =
   Type
     { extract =
         extractF
@@ -1367,9 +1377,11 @@ field key valueType =
     RecordType
       ( Data.Functor.Product.Pair
           ( Control.Applicative.Const
-              ( Dhall.Map.singleton
-                  key
-                  ( Dhall.expected valueType )
+              ( MonoidMap
+                  ( Dhall.Map.singleton
+                      key
+                      ( Dhall.expected valueType )
+                  )
               )
           )
           ( Data.Functor.Compose.Compose extractBody )
@@ -1440,10 +1452,12 @@ instance Contravariant RecordInputType where
 instance Divisible RecordInputType where
   divide f (RecordInputType bInputTypeRecord) (RecordInputType cInputTypeRecord) =
       RecordInputType
-    $ Dhall.Map.union
-      ((contramap $ fst . f) <$> bInputTypeRecord)
-      ((contramap $ snd . f) <$> cInputTypeRecord)
-  conquer = RecordInputType mempty
+          (Dhall.Map.union
+              ((contramap $ fst . f) <$> bInputTypeRecord)
+              ((contramap $ snd . f) <$> cInputTypeRecord)
+          )
+
+  conquer = RecordInputType Dhall.Map.empty
 
 inputFieldWith :: Text -> InputType a -> RecordInputType a
 inputFieldWith name inputType = RecordInputType $ Dhall.Map.singleton name inputType
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -90,6 +90,7 @@ import Control.Exception (Exception)
 import Control.Monad.Trans.State.Strict
 import Data.Functor.Contravariant (Contravariant(..), (>$<))
 import Data.Functor.Contravariant.Divisible (Divisible(..), divided)
+import Data.Monoid ((<>))
 import Data.Scientific (Scientific)
 import Data.Sequence (Seq)
 import Data.Text (Text)
--- a/src/Dhall.hs
+++ b/src/Dhall.hs
@@ -114,7 +114,6 @@ import qualified Control.Monad.Trans.State.Strict as State
 import qualified Data.Foldable
 import qualified Data.Functor.Compose
 import qualified Data.Functor.Product
-import qualified Data.HashMap.Strict.InsOrd
 import qualified Data.Scientific
 import qualified Data.Sequence
 import qualified Data.Set
@@ -126,6 +125,7 @@ import qualified Dhall.Binary
 import qualified Dhall.Context
 import qualified Dhall.Core
 import qualified Dhall.Import
+import qualified Dhall.Map
 import qualified Dhall.Parser
 import qualified Dhall.Pretty.Internal
 import qualified Dhall.TypeCheck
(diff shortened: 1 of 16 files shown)


sellout:
@Gabriel439 Thanks! Yeah, @FintanH tried this and we saved a significant amount of time (50–60s) on one of our benchmarks. Our branch that just eliminates insert-order completely (plain Map, no sorting) shaves like another 50–60s, but at a loss of the normalized ordering (and `format` ordering). So, this branch gets about half the benefit with no feature compromise. There might be a bit more we can squeeze out of it, but this is a huge improvement.

@FintanH may have more details to share, since he’s been doing all our benchmarking.

Gabriel439:
@sellout: Just out of curiosity: what is the total running time? (or, equivalently, what is the fractional performance improvement?)

src/Dhall/Map.hs
```diff
@@ -0,0 +1,258 @@
...
+-}
+member :: Ord k => k -> Map k v -> Bool
+member k (Map m _) = Data.Map.member k m
+
+{-| Combine two `Map`s, preferring keys from the first `Map`
+
+> union = unionWith (\v _ -> v)
+-}
+union :: Ord k => Map k v -> Map k v -> Map k v
+union (Map mL ksL) (Map mR ksR) = Map m ks
```
quasicomputational: Why not define in terms of `unionWith`?

src/Dhall/Map.hs
```diff
@@ -0,0 +1,258 @@
...
+
+    ks = ksL <|> filter (\k -> Data.Set.notMember k setL) ksR
+
+{-| Combine two `Map` on their shared keys, keeping the value from the first
+    `Map`
+
+> intersection = intersectionWith (\v _ -> v)
+-}
+intersection :: Ord k => Map k a -> Map k b -> Map k a
+intersection (Map mL ksL) (Map mR _) = Map m ks
```
quasicomputational: And, why not define in terms of `intersectionWith`?

FintanH:
@Gabriel439 @sellout Hey! So I did a couple of runs today and noticed the speedup was even better. It went from ~190s to ~100s. I've been adding some tweaks like inlining and I'm seeing some promising feedback. Just witnessed a run that was 55s :grin:

src/Dhall/Map.hs
```diff
@@ -0,0 +1,258 @@
...
+-}
+fromList :: Ord k => [(k, v)] -> Map k v
+fromList kvs = Map m ks
+  where
+    m = Data.Map.fromList kvs
+
+    ks = nub (map fst kvs)
+
+nub :: Ord k => [k] -> [k]
+nub = go Data.Set.empty
```
quasicomputational: Suggest renaming to `nubOrd` so that no-one else gets this confused with `Data.List.nub`.

src/Dhall/Map.hs
```diff
@@ -0,0 +1,258 @@
...
+import qualified Data.Set
+
+{-| A `Map` that remembers the original ordering of keys
+
+    This is primarily used so that formatting preserves field order
+
+    This is done primarily to avoid a dependency on @insert-ordered-containers@
+    and also to improve performance
+-}
+data Map k v = Map (Data.Map.Map k v) [k]
```
quasicomputational: `Seq` instead of a list improves the asymptotic of `union`.

sellout:
So, actually, my numbers might be off – we have a _lot_ of benchmark numbers floating around at the moment.

But this branch gives us a 102s run. The previous time I see (including the `denote` change, but still using InsOrd) is 196s. However, that doesn’t have some of the other post 1.17 improvements on it, AFAIK, which probably explains the discrepancy from the 60s improvement I mentioned before.

Pinging @FintanH to verify.

Other than this Map stuff (which is rapidly improving) our other bottlenecks look a lot like
```
typeWithA.loop          Dhall.TypeCheck    src/Dhall/TypeCheck.hs:(98,5)-(794,49)   12.7    6.1
normalizeWith.loop.kts' Dhall.Core         src/Dhall/Core.hs:1682:9-28              10.8   12.7
normalizeWith.loop.kts' Dhall.Core         src/Dhall/Core.hs:1676:9-28               6.5    7.4
alphaNormalize.kts₁     Dhall.Core         src/Dhall/Core.hs:1240:5-35               5.6    6.8
...
alphaNormalize.kts₁     Dhall.Core         src/Dhall/Core.hs:1232:5-35               3.4    4.0
```
the normalization cases are for `Union` and `Record`, and I’m guessing that’s also where we’re spending time in `typeWithA`, but need to break it down first.

src/Dhall/TypeCheck.hs
```diff
@@ -794,6 +790,16 @@ typeWithA tpa = loop
...
 
+newtype Process a b = Process { runProcess :: Either a b }
+
+instance Semigroup b => Semigroup (Process a b) where
+    Process eL <> Process eR = Process (liftA2 (<>) eL eR)
+
+instance Monoid b => Monoid (Process a b) where
+    mempty = Process (pure mempty)
+
+    mappend = (<>)
```
quasicomputational: This is going to break on GHC 8.2 because `Semigroup =/> Monoid`.

src/Dhall/Map.hs
```diff
@@ -0,0 +1,258 @@
...
+    deriving (Data, Eq, Foldable, Functor, Show, Traversable)
+
+instance Ord k => Data.Semigroup.Semigroup (Map k v) where
+    (<>) = union
```
quasicomputational: Might I suggest strategically avoiding defining `Semigroup` and `Monoid` for `Map`? These forgetful instances often do the wrong thing for `Data.Map.Map`; if you want the left-biased behaviour, it's often best to write `union` explicitly.

src/Dhall/Map.hs
Gabriel439: @quasicomputational: This was just in case `Data.Map.union` was more efficient implementation than `Data.Map.unionWith const`

src/Dhall/Map.hs
Gabriel439: Oh, now I remember why I added it.  `Dhall.RecordType` depends on `Map` being a `Monoid`.  I could add a newtype wrapper there just for this purpose, though

FintanH:
To shed some more light on the context of the above profile. There are a lot of unions in our AST and these grow every time we use a constructor. We think that's where the bottle neck is but how to optimize that is the million dollar question.

src/Dhall/Map.hs
Gabriel439: I'm going to add a benchmark first to double-check this because sometimes the better per-element constant factors of lists outweigh the asymptotic penalty

Gabriel439:
@FintanH: Give me a million dollars and I can fix it 😉 

In all seriousness, the next major step is just being more intelligent about where we are strict and lazy in normalization.  The logical conclusion of this is implementing the algorithm described in this book, since that algorithm prevents duplication of this sort of work:

https://dl.acm.org/citation.cfm?id=320040

Until that happens we might be able to play tricks by delaying expanding `constructors` fully until the end of normalization.

jneira:
Hi, i've tried to build this pr with the `stack build --stack-yaml stack-lts-6.yaml` ([after updating it](https://github.com/dhall-lang/dhall-haskell/pull/613)), the lts needed to build etlas with dhall as dependency and i am getting several errors about importing `<>` operator in the Typecheck module:
```console
C:\dhall-haskell\src\Dhall\TypeCheck.hs:3622:12:
    Ambiguous occurrence ‘<>’ 
    It could refer to either ‘Data.Monoid.<>’, imported from ‘Data.Monoid’ at src\Dhall\TypeCheck.hs:28:21-24 
     or ‘Pretty.<>’,  imported from ‘Data.Semigroup’ at src\Dhall\TypeCheck.hs:30:24-36
```

jneira:
For lts-6 the fix seems to be removing `import Data.Monoid ((<>))` in `Dhall.Typecheck` and `Dhall` modules. It seems ghc 7.10 is not able to see that `Data.Monoid` reexports the `Data.Semigroup` operator?

jneira:
Another possible caveat: `dhall-to-cabal` is using another Map ops in the `CabalToDhall` module: [`Map.difference`](https://github.com/dhall-lang/dhall-to-cabal/blob/master/lib/CabalToDhall.hs#L171),[`Map.filter`](https://github.com/dhall-lang/dhall-to-cabal/blob/master/lib/CabalToDhall.hs#L167),[`Map.mapMaybe`](https://github.com/dhall-lang/dhall-to-cabal/blob/master/lib/CabalToDhall.hs#L444)

src/Dhall/Map.hs
```diff
@@ -0,0 +1,258 @@
...
+
+{-| A `Map` that remembers the original ordering of keys
+
+    This is primarily used so that formatting preserves field order
+
+    This is done primarily to avoid a dependency on @insert-ordered-containers@
+    and also to improve performance
+-}
+data Map k v = Map (Data.Map.Map k v) [k]
+    deriving (Data, Eq, Foldable, Functor, Show, Traversable)
```
quasicomputational: Do we use the `Eq` instance? That's semi-dubious for this type, since we can reasonably want either strict equality of key order, or equality up to key order.

FintanH: It's used by `Expr` needing an `Eq` instance, since `Record` and `Union` are `Map`s under the hood

FintanH: I'm unsure that I follow. Would `Eq` not check that the list is in order as well?

quasicomputational: Have a look at the [definition of the instances for `InsOrdHashMap`](http://hackage.haskell.org/package/insert-ordered-containers-0.2.1.0/docs/src/Data-HashMap-Strict-InsOrd.html#InsOrdHashMap): it totally tosses away the indices of the elements. That's the code Dhall's currently using, meaning `Record` and `Union` compare for equality up to key insertion order, but this PR is going to change that.

FintanH: As far as I can see it only compares values inserted into the underlying `HashMap`. Maybe I'm misunderstanding something? :/

quasicomputational: Yes, that's what the `InsOrdHashMap` instance does. However, the derived `Dhall.Map.Map` instance is going to check for key ordering as well. To match the previous behaviour, we'd need to write something like:

```
instance (Eq k, Eq v) => Eq (Map k v) where
  Map a _ == Map b _ = a == b
```

FintanH: Right right, I've done that here actually :smile: https://github.com/FintanH/dhall-haskell/blob/fintan/new-map/src/Dhall/Map.hs#L55-L57

src/Dhall/TypeCheck.hs
```diff
@@ -21,11 +21,13 @@ module Dhall.TypeCheck (
     , TypeMessage(..)
     ) where
 
+import Control.Applicative (liftA2)
 import Control.Exception (Exception)
 import Data.Data (Data(..))
 import Data.Foldable (forM_, toList)
 import Data.Monoid ((<>))
 import Data.Sequence (Seq, ViewL(..))
+import Data.Semigroup (Semigroup(..))
```
FintanH: There is an ambiguous occurrence between the `<>` coming from here and the one above from `Data.Monoid`. I'd recommend removing the one from `Data.Monoid`

FintanH: Also happens in [`src/Dhall.hs`](https://github.com/dhall-lang/dhall-haskell/blob/gabriel/new_map_2/src/Dhall.hs#L93). This seems to be just for GHC-8.2 btw

src/Dhall/Map.hs
Gabriel439: @FintanH @quasicomputational: I prefer to use the new order-sensitive `Eq` instance.  In other words, I view a `Dhall.Map.Map` as an ordered association list that just happens to provide better time complexity for element lookup.

The main reason why is that this is more consistent with the standard, which no longer requires that record/union equality is order insensitive (See: https://github.com/dhall-lang/dhall-lang/pull/223).  Instead, whenever order-insensitivity is required the standard requires an explicit β-normalization step which (among other things) sorts the records/unions.

src/Dhall/TypeCheck.hs
jneira: For ghc-7.10.3 too 😉

src/Dhall/Map.hs
```diff
@@ -0,0 +1,312 @@
...
+import qualified Data.Set
+
+{-| A `Map` that remembers the original ordering of keys
+
+    This is primarily used so that formatting preserves field order
+
+    This is done primarily to avoid a dependency on @insert-ordered-containers@
+    and also to improve performance
+-}
+data Map k v = Map (Data.Map.Map k v) [k]
```
jneira: could the constructor be exported? it will let users extend it with new ops and automatic class deriving. `dhall-to-cabal` would need to add some operators and instances for `Semigroup` and `Monoid`

Gabriel439: I would like to avoid exporting the constructor so that we can easily change it if other representations end up being more efficient.  I'd prefer to just add whatever `dhall-to-cabal` needs, including the `Semigroup`/`Monoid` instance if necessary

jneira: Nice! I write down the list of operations used by dhall-to-cabal [here](https://github.com/dhall-lang/dhall-haskell/pull/612#issuecomment-426176497)

src/Dhall.hs
```diff
@@ -90,7 +90,6 @@ import Control.Exception (Exception)
 import Control.Monad.Trans.State.Strict
 import Data.Functor.Contravariant (Contravariant(..), (>$<))
 import Data.Functor.Contravariant.Divisible (Divisible(..), divided)
-import Data.Monoid ((<>))
```
jneira: `stack build --stack-yaml stack-lts6.yaml` fails with
```console
  C:\dhall-haskell\src\Dhall.hs:153:16:
  Not in scope: ‘<>’ 
    Perhaps you meant one of these:
      data constructor ‘Data.Sequence.:>’ (imported from Data.Sequence),
     ‘<|>’ (imported from Control.Applicative),
     ‘<*>’ (imported from Prelude)
```
Adding it again makes it work and the builds with `stack.yaml` and `stack-lts11.yaml` (`ghc-8.2.2`) work too.

jneira: It seems in a previous version the module imported `Data.Semigroup`, so it was necessary to remove `Dhall.Monoid.(<>)`, but without any of them `<>` is out of scope for previous versions of ghc

FintanH: I tried searching stackage.org and it seems `Data.Semigroup` wasn't available in `base` at the time. Is there a reason we're supporting and old lts? If it's to support stack + a certain version of ghc we can just specify which compiler to use in the yaml file instead of the lts.

jneira: i am afraid that (etlas)[https://github.com/typelead/etlas], that uses `dhall` through (dhall-to-etlas)[https://github.com/eta-lang/dhall-to-etlas] is being compiled with lts-6 and afaik it can't be compiled with newer lts's

jneira: But adding `import Data.Monoid ((<>))` seems to fix it, at least for `lts-6`, `lts-11` and `lts-12`, could it cause some problem now or in the future?

FintanH: Woof! Is there any plan to update that? I don't think it would change things in the future. I think we just need to import `Data.Semigroup (Semigroup)` for the class, and then we can import `Data.Monoid ((<>))` for the operator

jneira: Maybe @rahulmutt (owner of etlas) could answer it better than me :wink:
The main goal is [make dhall the default configuration format for etlas](https://github.com/typelead/etlas/issues/66)

FintanH: I've currently got a `stack build` going with `etlas` and this [yaml](https://github.com/FintanH/etlas/blob/fintan/update-lts/stack-ghc822.yaml)

FintanH: Ah ok. I'll move the discussion to those repos. In the meantime, I did a fix up of this branch for both stack yamls. I'll make a PR against this branch now

FintanH: https://github.com/dhall-lang/dhall-haskell/pull/616

src/Dhall/Core.hs
```diff
@@ -445,7 +442,75 @@ data Expr s a
...
+  fmap f (CombineTypes e1 e2) = CombineTypes (fmap f e1) (fmap f e2)
+  fmap f (Prefer e1 e2) = Prefer (fmap f e1) (fmap f e2)
+  fmap f (Merge e1 e2 maybeE) = Merge (fmap f e1) (fmap f e2) (fmap (fmap f) maybeE)
+  fmap f (Constructors e1) = Constructors (fmap f e1)
+  fmap f (Field e1 v) = Field (fmap f e1) v
+  fmap f (Project e1 vs) = Project (fmap f e1) vs
+  fmap f (Note s e1) = Note s (fmap f e1)
+  fmap f (ImportAlt e1 e2) = ImportAlt (fmap f e1) (fmap f e2)
+  fmap f (Embed a) = Embed (f a)
+  {-# INLINABLE fmap #-}
```
quasicomputational: I think it's worth adding a comment noting that this annotation is the whole reason why this instance isn't derived.

jneira:
Performance of compile dhall-to-cabal files has improved significantly :fireworks: :
```console
C:\dhall-to-cabal>timer "dhall-master < dhall-to-cabal.dhall > null"
00:00:02,44
00:00:23,93

C:\dhall-to-cabal>timer "dhall-newmap < dhall-to-cabal.dhall > null"
00:00:36,72
00:00:48,08
```
The new profile shows that cost has moved to `Dhall.Parser.Combinators` (although the new `Dhall.Map` is the second one):
```
	   dhall +RTS -p -RTS

	total time  =       14.47 secs   (14467 ticks @ 1000 us, 1 processor)
	total alloc = 22,183,225,456 bytes  (excludes profiling overheads)

COST CENTRE            MODULE                   SRC                                       %time %alloc

>>=                    Dhall.Parser.Combinators src\Dhall\Parser\Combinators.hs:94:5-48    18.3    1.7
traverseWithKey_       Dhall.Map                src\Dhall\Map.hs:302:1-77                  11.7   22.2
fmap                   Dhall.Map                src\Dhall\Map.hs:73:3-39                    9.8   15.9
typeWithA.loop         Dhall.TypeCheck          src\Dhall\TypeCheck.hs:(98,5)-(788,49)      8.9    0.7
normalizeWith.loop     Dhall.Core               src\Dhall\Core.hs:(1482,2)-(1833,22)        6.3    8.5
normalizeWith          Dhall.Core               src\Dhall\Core.hs:(1480,1)-(1833,22)        5.3   18.2
normalize              Dhall.Core               src\Dhall\Core.hs:1372:1-41                 3.6    4.3
denote                 Dhall.Core               src\Dhall\Core.hs:(1398,1)-(1462,39)        3.2    3.1
typeWithA.loop.process Dhall.TypeCheck          src\Dhall\TypeCheck.hs:(534,13)-(539,85)    2.9    0.0
exprFromUncachedImport Dhall.Import             src\Dhall\Import.hs:(580,1)-(673,45)        2.8    0.3
<|>                    Text.Megaparsec.Internal Text\Megaparsec\Internal.hs:195:3-16        2.3    2.0
fmap                   Dhall.Context            src\Dhall\Context.hs:32:15-21               2.0    6.6
pure                   Dhall.Parser.Combinators src\Dhall\Parser\Combinators.hs:75:5-24     1.9    0.9
shift                  Dhall.Core               src\Dhall\Core.hs:(758,1)-(920,29)          1.8    1.9
fmap                   Dhall.Parser.Combinators src\Dhall\Parser\Combinators.hs:68:5-41     1.3    1.2
lookup                 Dhall.Context            src\Dhall\Context.hs:(61,1)-(68,33)         1.2    0.1
getParserState         Dhall.Parser.Combinators src\Dhall\Parser\Combinators.hs:155:5-58    1.1    0.7
```

Gabriel439:
@jneira: Do those timings mean that this branch got slower compared to `master`?

jneira:
@Gabriel439 no, no, sorry for the poor stats but master took ~21 seconds and new_map ~11 seconds

src/Dhall/Core.hs
FintanH: https://github.com/dhall-lang/dhall-haskell/pull/620

jneira:
@Gabriel439 is there any plan for implement the Map operations used by dhall-to-cabal (difference, filter,...)

Gabriel439:
@jneira: Oops!  Yeah, I can add those

Gabriel439:
@jneira: Alright, I've pushed a branch of `dhall-to-cabal` that builds against this branch of `dhall`: https://github.com/dhall-lang/dhall-to-cabal/compare/gabriel/latest_dhall

I'm just going to add some doctests for `Dhall.Map` before merging

jneira:
@Gabriel439 Not sure if i should comment here but i've detected that performance of actual master version is slightly worse that the pr (at least [my branch created from it](https://github.com/eta-lang/dhall-haskell/tree/new_map)

```console
jneira@laptop MINGW64 /d/ws/eta/dhall/dhall-to-cabal
$ time dhall-newmap < dhall-to-cabal.dhall > nul

real    0m11.479s
user    0m0.000s
sys     0m0.031s

jneira@laptop MINGW64 /d/ws/eta/dhall/dhall-to-cabal
$ time dhall-master < dhall-to-cabal.dhall > nul

real    0m16.116s
user    0m0.000s
sys     0m0.015s
```
Those results match the benchmarks? 
cost of the branch is around 11 secs and master 16 :thinking:

jneira:
Profiles are significative:

```
	   dhall-master-prof +RTS -p -RTS

	total time  =       21.05 secs   (21055 ticks @ 1000 us, 1 processor)
	total alloc = 26,777,422,512 bytes  (excludes profiling overheads)

COST CENTRE            MODULE                   SRC                                       %time %alloc

toList.\               Dhall.Map                src\Dhall\Map.hs:462:33-51                 27.8    0.0
>>=                    Dhall.Parser.Combinators src\Dhall\Parser\Combinators.hs:94:5-48    11.2    1.4
fmap                   Dhall.Map                src\Dhall\Map.hs:78:3-39                    7.1   13.1
typeWithA.loop         Dhall.TypeCheck          src\Dhall\TypeCheck.hs:(98,5)-(788,49)      6.4    0.6
traverseWithKey        Dhall.Map                src\Dhall\Map.hs:(439,1)-(441,36)           6.1   15.2
normalizeWith.loop     Dhall.Core               src\Dhall\Core.hs:(1493,2)-(1844,22)        4.7    7.1
normalizeWith          Dhall.Core               src\Dhall\Core.hs:(1491,1)-(1844,22)        3.8   15.0
traverseWithKey.f'     Dhall.Map                src\Dhall\Map.hs:441:5-36                   3.8    7.2
toList                 Dhall.Map                src\Dhall\Map.hs:462:1-55                   3.0   10.6
normalize              Dhall.Core               src\Dhall\Core.hs:1383:1-41                 2.3    3.6
denote                 Dhall.Core               src\Dhall\Core.hs:(1409,1)-(1473,39)        2.2    2.6
typeWithA.loop.process Dhall.TypeCheck          src\Dhall\TypeCheck.hs:(534,13)-(539,85)    1.8    0.0
sort.ks                Dhall.Map                src\Dhall\Map.hs:169:5-24                   1.8    3.3
<|>                    Text.Megaparsec.Internal Text\Megaparsec\Internal.hs:195:3-16        1.7    1.7
exprFromUncachedImport Dhall.Import             src\Dhall\Import.hs:(596,1)-(689,45)        1.6    0.2
pure                   Dhall.Parser.Combinators src\Dhall\Parser\Combinators.hs:75:5-24     1.4    0.7
fmap                   Dhall.Context            src\Dhall\Context.hs:32:15-21               1.2    5.4
shift                  Dhall.Core               src\Dhall\Core.hs:(769,1)-(931,29)          1.0    1.5
```
vs
```
	   dhall-newmap-prof +RTS -p -RTS

	total time  =       14.40 secs   (14397 ticks @ 1000 us, 1 processor)
	total alloc = 22,008,482,176 bytes  (excludes profiling overheads)

COST CENTRE            MODULE                   SRC                                       %time %alloc

>>=                    Dhall.Parser.Combinators src\Dhall\Parser\Combinators.hs:94:5-48    17.5    1.7
traverseWithKey_       Dhall.Map                src\Dhall\Map.hs:302:1-77                  12.4   22.4
fmap                   Dhall.Map                src\Dhall\Map.hs:73:3-39                   10.5   16.0
typeWithA.loop         Dhall.TypeCheck          src\Dhall\TypeCheck.hs:(98,5)-(788,49)      9.3    0.7
normalizeWith.loop     Dhall.Core               src\Dhall\Core.hs:(1482,2)-(1833,22)        5.8    8.6
normalizeWith          Dhall.Core               src\Dhall\Core.hs:(1480,1)-(1833,22)        5.4   18.3
normalize              Dhall.Core               src\Dhall\Core.hs:1372:1-41                 3.8    4.4
denote                 Dhall.Core               src\Dhall\Core.hs:(1398,1)-(1462,39)        3.1    3.1
typeWithA.loop.process Dhall.TypeCheck          src\Dhall\TypeCheck.hs:(534,13)-(539,85)    2.7    0.0
exprFromUncachedImport Dhall.Import             src\Dhall\Import.hs:(580,1)-(673,45)        2.7    0.3
<|>                    Text.Megaparsec.Internal Text\Megaparsec\Internal.hs:195:3-16        2.6    2.0
fmap                   Dhall.Context            src\Dhall\Context.hs:32:15-21               2.1    6.6
pure                   Dhall.Parser.Combinators src\Dhall\Parser\Combinators.hs:75:5-24     1.8    0.9
fmap                   Dhall.Parser.Combinators src\Dhall\Parser\Combinators.hs:68:5-41     1.5    1.2
shift                  Dhall.Core               src\Dhall\Core.hs:(758,1)-(920,29)          1.5    1.9
getParserState         Dhall.Parser.Combinators src\Dhall\Parser\Combinators.hs:155:5-58    1.3    0.7
```

PR #616, opened 2018-10-03
Fix lts builds

FintanH:
* Merged `master`
* Fixed builds for both `stack build` and `stack build --stack-yaml stack-lts-6.yaml`


Gabriel439:
Thank you! 🙂

PR #621, opened 2018-10-04
Build against GHC 7.10.3

Gabriel439:
... as requested by @jneira so that Dhall will continue to work as a
dependency of Eta, which is built using GHC 7.10.3

This adds CI support for testing the build against GHC 7.10.3 and also
fixes issues building against that GHC version that were caught in the process

### DIFF
--- a/benchmark/parser/Main.hs
+++ b/benchmark/parser/Main.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP               #-}
 {-# LANGUAGE OverloadedStrings #-}
 
 module Main where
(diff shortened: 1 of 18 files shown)


jneira:
very thankful for this :pray:

quasicomputational:
When is it likely that Eta's going to upgrade and hence Dhall can drop the CPP?

jneira:
Well, there is a [pr in etlas about that](https://github.com/typelead/etlas/pull/76) and i've requested the review of eta/etlas owner @rahulmutt.

PR #622, opened 2018-10-05
Merge adjustment

sellout:
This is really two separate changes. Either could be taken individually, and the commits are divided that way.

The first removes the special-casing of `merge` annotation, while the second converts `merge` into a function on the union.

Either of these also affects the spec.


sellout:
Ah, took me forever to find the related issue somehow, but: dhall-lang/dhall-lang#106

src/Dhall/Binary.hs
```diff
@@ -154,6 +154,12 @@ encode_1_1 (Const Type) =
     TString "Type"
 encode_1_1 (Const Kind) =
     TString "Kind"
+encode_1_1 (App (Merge _T₀ t₀) u₀) =
+    TList [ TInt 6, t₁, u₁, _T₁ ]
```
sellout: This avoids breaking the encoding (and decoding) of the previous fully-applied `merge`.

src/Dhall/Binary.hs
```diff
@@ -262,16 +268,10 @@ encode_1_1 (Some t₀) =
...
     t₁ = encode_1_1 t₀
-encode_1_1 (Merge t₀ u₀ Nothing) =
-    TList [ TInt 6, t₁, u₁ ]
-  where
-    t₁ = encode_1_1 t₀
-    u₁ = encode_1_1 u₀
-encode_1_1 (Merge t₀ u₀ (Just _T₀)) =
-    TList [ TInt 6, t₁, u₁, _T₁ ]
+encode_1_1 (Merge _T₀ t₀) =
+    TList [ TInt 6, t₁, _T₁ ]
```
sellout: But the 2-element encoding conflicts – it used to mean no type annotation, but now it means partially-applied.

I couldn’t quickly come up with anything that wasn’t incompatible with the old version.

Gabriel439:
@sellout: I'm in favor of changing `merge` to support partial application as described in https://github.com/dhall-lang/dhall-lang/issues/106, but I'm against adding a new explicit type argument to it since that is a breaking change that makes the language more verbose.

I don't think a type argument is necessary to support partial application of `merge`.  You should be able to just require that if you partially apply `merge` to a record of handlers then the record should be non-empty and then you can infer the final type from the result type of the first handler.  If a user wants to merge an empty record of handlers, they can still eta-expand it, which is not that verbose:

```haskell
λ(x : <>) → merge {=} x : T
```

sellout:
> I don't think a type argument is necessary to support partial application of merge.

Yeah, I agree they’re independent. And considering `absurd` is the only case where the annotation is required, I’m good with that. My arguments for that change were
1. it’s consistent with the way we use types elsewhere (empty `List` being the one other real exception[†] now that `Some`/`None` are in) and
2. it removes some special-casing from the implementation.

>  If a user wants to merge an empty record of handlers, they can still eta-expand it

Are you saying we shouldn’t bother to support `merge {=} : <> → T`, but rather require η-expansion in that case?

[†]: It’s been brought up on the FP Slack that getting rid of the `Annot` special casing in general might be nice – like having `[...]` syntax be non-empty, and then using either `Nil T` or `None (List T)` for the empty case (the latter implying that `List` itself is non-empty). I think just speculation, but something to mull over. And the `Nil T` approach could be added like `None T` – without actually removing the old handling for the time being (while I think `None (List T)` is necessarily a breaking change).

Gabriel439:
@sellout: Yeah, my reasoning is that `merge {=} <> : T` is more similar in spirit to `[] : List T` than `None T`.

Both `merge` and `List` are designed so that that if the user were to always supply a type annotation then the syntax does not change abruptly between the empty and non-empty case.  The way I view this is that for `merge`/`List` there are countless non-empty cases (i.e. a merge with N handlers or a list of `N` elements, for all `0 < N`), so it's more unusual to behave differently for the empty case of `N == 0` because it feels like a special case in comparison to all of the possible non-empty examples.  Contrast that with `Optional` where the syntax for empty case is not really a special case if there are only two possible cases.

> Are you saying we shouldn’t bother to support `merge {=} : <> → T`, but rather require η-expansion in that case?

Yes.  This would be similar to the limitation for empty lists: if you want to create a polymorphic empty list you have to explicitly quantify it:

```
λ(a : Type) → [] : List a
```

If the user tried to write `merge {=}` or `merge {=} <>` they'd get a type error that would explain to them that a merge for for an empty union must be saturated and have a type annotation.

> It’s been brought up on the FP Slack that getting rid of the Annot special casing in general might be nice – like having `[...]` syntax be non-empty, and then using either `Nil T` or `None (List T)` for the empty case (the latter implying that List itself is non-empty). I think just speculation, but something to mull over. And the `Nil T` approach could be added like `None T` – without actually removing the old handling for the time being (while I think `None (List T)` is necessarily a breaking change)

If we were to do this I would prefer `Nil T` over `None (List T)` since in my view `Optional (List a)` would not be an ergonomic type and it would really need to be something like `Optional (NonEmptyList a)` since the word `List` does not connote non-emptiness.  However, going back to the above reasoning I try to minimize the syntactic difference for the empty case for lists to avoid it coming across as a special case.  I understand that having to add a type annotation for the empty list is a bit jarring compared to the non-empty case, but I believe that switching to `Nil` instead for the empty list would be even more jarring in comparison.

Gabriel439:
@sellout: Oh, sorry, I misunderstood part of your proposal.  Yes, I think supporting `merge {=} : <> → T` is fine so that users don't have to eta-expand the empty case.  For some reason I though you were proposing that `merge {=}` would have type `<> → ∀(T : Type) → T`, but now I understand that you are proposing that a user can annotate a partially applied `merge` with an annotation if the record of handlers is empty, which seems reasonable to me

sellout:
Closing – will submit a new PR with the correct change.

PR #625, opened 2018-10-07
Add null check

jneira:
As it is used at least by dhall-to-cabal...

### DIFF
--- a/src/Dhall/Map.hs
+++ b/src/Dhall/Map.hs
@@ -29,8 +29,9 @@ module Dhall.Map
       -- * Query
     , lookup
     , member
+    , null
     , uncons
-
+    
       -- * Combine
     , union
     , unionWith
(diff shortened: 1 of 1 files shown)


quasicomputational:
`Data.Foldable.null` ought to be good enough, I think? Even if there's an asymptotic win, we should just over-ride the default `Foldable` implementation and use that in client code (i.e., in `dhall-to-cabal`).

Gabriel439:
@quasicomputational: Yeah, `Data.Foldable.null` should do the right thing in this case

Gabriel439:
I also have a branch out that builds against the `Dhall.Map` changes: https://github.com/dhall-lang/dhall-to-cabal/compare/gabriel/latest_dhall

jneira:
Thanks for pointing it out, closing this

Issue #630, opened 2018-10-10
dhall thaw


FintanH:
I ran into a workflow issue while using `dhall freeze` with our configs.

1. Ran `dhall freeze` to freeze all files
2. Had to merge changes that were upstream
3. Now files didn't compile due to caching and difference in a record
	* A field was added (due to the merge) post-freeze which was not present in the cached version
4. Deleted cache to see that integrity checks were mismatching

Thus, I need to to `thaw` or `refreeze` all the files to get the new hash. This would also apply to when we want to develop on frozen files and then lock them again  incorporating the new changes.

jneira:
And what about if `dhall freeze` itself would update the existent hashes?

FintanH:
I feel like I'd be happy with that, because you mostly want the compile to fail if the integrity check failed. You want the freeze to always freeze the current state.

FintanH:
One workaround is via `find . -type f | xargs sed -i -r 's/ sha256:[a-f0-9][a-f0-9]*//g'`

Gabriel439:
Yeah, I agree that the ideal behavior is that `dhall freeze` just ignores any existing hashes and updates all of them.

joneshf:
Both sound useful! It seems like there's still a case for something like `dhall thaw`. Sometimes you do want to throw out the shas. It can be useful so other commands and executables can work. It can also be useful when going between `dhall` versions (as the algorithm changed recently).

I guess in both of those cases, there are different things that could happen: allow other commands/executables to ignore shas and backport re`freeze` behavior to older versions, respectively. Maybe I'm not supporting `thaw` enough, but those are two recent problems I ran into.

sellout:
I don’t have strong feelings on this, so feel free to disregard, but it seems like one might want more fine-grained control. I.e., not have `freeze` just update all hashes. I could see this being like `freeze` adds hashes to un-hashed things (the current implementation, yeah?), so you can selectively remove shas, then freeze, updating only thase. If you want to re-hash everything, you need to either remove shas in advance (like the `thaw` proposed), or maybe have a `freeze --force` to overwrite shas.

Alternatively, I could see passing a prefix like `freeze --prefix 'https://github.com/.../Prelude/'` that would update the shas for any imports that match the prefix, so you can update a single repo or something.

Gabriel439:
[...]

PR #637, opened 2018-10-13
Change `dhall freeze` to refreeze imports

Gabriel439:
Fixes: https://github.com/dhall-lang/dhall-haskell/issues/630

`dhall freeze` will now ignore the hash on existing imports and attempt to
refreeze them

### DIFF
--- a/src/Dhall/Freeze.hs
+++ b/src/Dhall/Freeze.hs
@@ -38,9 +38,16 @@ hashImport
     -> Import
     -> IO Import
 hashImport directory _standardVersion import_ = do
+    let unprotectedImport =
+            import_
+                { importHashed =
+                    (importHashed import_)
+                        { hash = Nothing
+                        }
+                }
     let status = set standardVersion _standardVersion (Dhall.Import.emptyStatus directory)
 
-    expression <- State.evalStateT (Dhall.Import.loadWith (Embed import_)) status
+    expression <- State.evalStateT (Dhall.Import.loadWith (Embed unprotectedImport)) status
 
     case Dhall.TypeCheck.typeOf expression of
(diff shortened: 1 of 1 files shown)


Profpatsch:
Uuh, I don’t think that is a good idea. My intuition tells me an action like `freeze` should be “locally idempotent” by default, and require an option like `freeze --update` to query for existing frozen dependencies (comparable to stateful package managers never updating pinned packages without explicit user consent).

If I add some import to my code and run `dhall freeze`, I’d expect it to not touch my existing frozen imports and only freeze the unfrozen ones.

Gabriel439:
@Profpatsch: How about updating them if import resolution fails (i.e. the semantic integrity check is for the wrong version)?

Gabriel439:
I added support so that `dhall refreeze` will only update a semantic integrity check if the check fails to resolve

Profpatsch:
Hm, it might be good if the tool also prints a warning if the hashsums don’t match. This would point to an issue with upstream (e.g. because the link points to a moving target).

On second thought, I can’t imagine a situation in which mismatching hashes don’t point to a problem. So it might be best to collect the list an print an error like this:

```
The upstream of these imports has changed from their local integrity checks:
<list of import URLs with their line numbers>

<something about using unchanging URLs here>

Once you are sure the links are correct, run `dhall refreeze --force` and I will update all existing integrity checks.
```

Gabriel439:
@Profpatsch: I support the idea of listing the hash mismatches as it refreezes them.

I'm not as sure that we need the user to supply a `--force` flag to confirm the change, for two reasons:

* It's not clear when the user would want to preserve broken hashes
* I presume the user is using version control so that they can roll back changes if they make a mistake

PR #649, opened 2018-10-19
Doc should be Prefer

FintanH:
The doc said `CombineRight`. I'm guessing a name change was made but the doc wasn't updated.


Gabriel439:
Whoops!  You're right.  Thanks 🙂

PR #667, opened 2018-10-31
Encode doubles as `Double`, do bounds checks, and add support for NaN and Infinity values

drvirgilio:
See discussion in dhall-lang issue [#252](https://github.com/dhall-lang/dhall-lang/issues/252). I'm not sure exactly on where the parser for "[-]Infinity" should go so I kind of guessed.


drvirgilio:
I merged the changes to the standard so I think this is ready to be merged.

dhall/src/Dhall/Pretty/Internal.hs
```diff
@@ -306,8 +305,8 @@ prettyNumber = literal . Pretty.pretty
 prettyNatural :: Natural -> Doc Ann
 prettyNatural = literal . Pretty.pretty
 
-prettyScientific :: Scientific -> Doc Ann
-prettyScientific = literal . Pretty.pretty . show
+prettyDouble :: Double -> Doc Ann
+prettyDouble = literal . Pretty.pretty . show
```
PierreR: @Gabriel439 I believe `show` is not needed here.

Gabriel439:
@f-f: I'll go ahead and merge this now and add tests later, because I need to sync this repository with the `dhall-lang` test suite first.  The other reason why is that I would like to add the tests to `dhall-lang` first and then mirror them to this repository rather than the other way around

Gabriel439:
Actually, there are test failures on this branch that are being hidden by a CI misconfiguration on my part.  I'm going to fix that first in a separate pull request and then fix the tests on this branch

Gabriel439:
I have a pull request up against this branch with fixes to the test suite: https://github.com/drvirgilio/dhall-haskell/pull/1

drvirgilio:
Finally got around to making a few changes:
1.  Added tests which check a few boundary conditions such as the point where `Integer/toDouble` rounds up to `Infinity`.
2.  Found a workaround for `fromInteger`'s weird rounding behavior by using `read . show`.
3.  Made the changes suggested by @Gabriel439

Gabriel439:
@drvirgilio: Thanks for doing this! 🙂

PR #668, opened 2018-10-31
Implement --prefix for dhall freeze

### LINKED ISSUE #663: Allow `dhall freeze` to only freeze specific imports
I've found that while working I often want to freeze a subset of imports - in particular, imports from `prelude.dhall-lang.org`. What about adding `--host` to `dhall freeze`, as one solution?

ocharles:
Fixes #663

### DIFF
--- a/dhall/src/Dhall/Freeze.hs
+++ b/dhall/src/Dhall/Freeze.hs
@@ -6,9 +6,11 @@ module Dhall.Freeze
     ( -- * Freeze
       freeze
     , hashImport
+    , allowPrefix
     ) where
 
 import Control.Exception (SomeException)
+import Control.Monad.IO.Class (MonadIO)
 import Data.Monoid ((<>))
 import Data.Maybe (fromMaybe)
 import Data.Text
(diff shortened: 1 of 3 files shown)


dhall/src/Dhall/Freeze.hs
```diff
@@ -122,7 +129,28 @@ freeze inplace _standardVersion = do
...
+allowPrefix prefix i =
+  case importType (importHashed i) of
+      Dhall.Core.Missing ->
+          return False
+
+      Dhall.Core.Env{} ->
+          return False
+
+      Dhall.Core.Remote (Dhall.Core.URL scheme authority file query fragment _maybeHeaders) ->
+        return (prefix `isPrefixOf` Dhall.Import.remoteToURL scheme authority file query fragment)
```
Gabriel439: I think it should match a logical prefix against the canonicalized path if possible.

By logical prefix I mean that if a person specifies that the prefix is `"./foo"` then it should match `"./foo/bar"` but not `"./foo.dhall"`.  Similarly, a prefix of `"."` should not match the path `"../foo"` (in fact, the prefix should give a parse error since it's an incomplete path prefix and needs to minimally be `"./"`).

The reason for canonicalizing the path before matching the prefix against it is so that a prefix of `"./"` does not match `"./../foo"`.

ocharles: But the whole point is that I want to freeze only stuff that's on `prelude.dhall-lang.org` - it seems that your suggestion would have me lose that distinction?

Gabriel439: @ocharles: I think it would still work because you could freeze `"https://prelude.dhall-lang.org/"` which is a valid logical prefix

Gabriel439:
I'll close this out as it is a bit out of date at this point

PR #669, opened 2018-11-01
Build against `ghc-8.6`

Gabriel439:
This switches to `ghc-8.6` as the new default used for both CI and local
development


quasicomputational:
I dunno about this, considering the scary bug 8.6.1 has. 8.6.2 ought to be along very shortly indeed with some really important fixes, though.

quasicomputational:
Yeah, that makes sense. Being a row in a build matrix ought to be fine, but I think we ought to be suspicious of binaries produced by 8.6.1 or relying on it for the development process.

PR #671, opened 2018-11-05
Update stack/cabal config to sync with new structure

jneira:
* Tested with `stack build` (from root and each subproject)
  * Not sure if nix section is needed at top-level
  * For all subproject `dhall` is a local extra dep
* Tested with `cabal v2-build` (from root and each subproject)
  * You can build all packages from root with `cabal v2-build all` or choose each one with `cabal v2-build dhall-bash`
  * ~~For all subprojects `dhall` is a local package~~


jneira:
Thinking twice about it, it seems that each subpackage should be buildable in isolation, to let users build them from source distributions (with `cabal get` or `stack unpack`)

jneira:
Added stack.yaml for lts-6 and lts-11
Unfortunately(?) `dhall-json` uses `ApplicativeDo`, not supported in `ghc-7.10`

jneira:
Well, i've added the[ stack.yaml at the top](https://github.com/dhall-lang/dhall-haskell/blob/24ce36401f6ac147047a13a47b4a6bc8f458af01/stack.yaml) but i thought that keep existent yaml files could be useful to build  each one individually with its own config settings. I'll delete them.
@Gabriel439 About lts-6, we would like to build dhall-json and other subprojects with etlas/eta and eta is between ghc-7.10 and ghc-8.*. So if lts-6 can build them surely eta will do too. However, coincidentally, eta has `ApplicativeDo` backported from ghc-8.

Gabriel439:
@jneira: Let me review more closely before you delete them.  I just want to make sure that we can provide convenient instructions for people to build each individual repository from the top-level

jneira:
Ok, i've restored them. To be fair, there is a duplication between the top and the submodules and you can build each one from root with `stack build dhall-json` f.e. 
Maybe the best option could be keep only the top one and add it to submodules if needed.

jneira:
Ok, i've deleted them. 
@Gabriel439 Do you prefer to make the docs changes in this pr or in another one?

Gabriel439:
@jneira: Great work!  Thank you :slightly_smiling_face:

PR #670, opened 2018-11-02
Use an Ordered Set for Project

DanBurton:
Addresses #664

PR #672, opened 2018-11-06
Test that formatting preserves projection order

DanBurton:
Tested that these fail on a commit prior to merging #670 like so:

```bash
git checkout dc5493d87f7f98b71b457c0163edab043a704b91
stack test
```

The result, as formerly expected, was that the projected fields were coming out in sorted order:

```
    preserve the original order of projections:                                     FAIL
      tests/Format.hs:102:
      The formatted expression did not match the expected output
      expected: "e.{ foo, bar, baz }"
       but got: "e.{ bar, baz, foo }"
```

The tests pass when run against the latest master commit.

### DIFF
--- a/dhall/tests/Format.hs
+++ b/dhall/tests/Format.hs
@@ -34,6 +34,10 @@ formatTests =
             "fieldOrder"
         , should
             Unicode
+            "preserve the original order of projections"
+            "projectionOrder"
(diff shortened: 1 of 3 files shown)


Gabriel439:
Good catch!  I forgot about the test

PR #675, opened 2018-11-10
Add support for multi-`let`

Gabriel439:
... as standardized in: https://github.com/dhall-lang/dhall-lang/pull/266

This also adds `dhall lint` support for consolidating nested `let` expressions


Gabriel439:
Note that in the process of implementing this I realized that I forgot to standardize how to encode/decode multi-`let` expressions, so I will finish that before merging this

joneshf:
Any thoughts on how we might mitigate that happening in the future?

PR #684, opened 2018-11-15
Allow custom normalization to be entered top-down

ocharles:
As discussed in https://github.com/dhallix/nix-derivation/issues/8


ocharles:
Heh, I wasn't sure if you'd like the marching indentation that this causes, but if you're happy with it then we're all good! Can always fix up later.

PR #687, opened 2018-11-16
Allow importing to use arbitrary monads

ocharles:
Continuing to try and solve https://github.com/dhallix/nix-derivation/issues/8 - I found out that I can't actually use the `normalizeWithM` work with importing as it doesn't let me choose `m`. This work is trying to do the bare minimum to allow me to choose a more interesting monad.


Gabriel439:
I'm going to close this since I think it would be obsoleted by large incoming performance improvements.  There is a PhD student who has in-progress work that is very promising.  That would enable the simpler approach of just processing the normalized expression tree

ocharles:
I agree, and that's very exciting news!

PR #688, opened 2018-11-17
Update tests to match the filesystem layout from the standard

Gabriella439:
This begins updates the filesystem layout for the tests to match
the new layout from the standard test suite (See:
https://github.com/dhall-lang/dhall-lang/pull/265)

This is still missing a part of the standard requirement, which is checking
parsing and import test results against an expected output (instead of
just checking that they succeed).  I plan to add that in a subsequent
pull request.  This is mainly to unblock other features that require using
the new standard layout.


jneira:
@Gabriel439 sorry for comment after merging but i've noticed that `B` cases in `dhall/tests/parser` have the `.dhall` extension but their contents are json (?). Shouldnt they have the `.json` extension instead?

Gabriel439:
@jneira: Yeah, I noticed that, too.  This is an issue with the upstream test suite in dhall-lang as well

PR #691, opened 2018-11-21
Version 1.18.0 → 1.19.0

Gabriel439:


jneira:
@Gabriel439 mmm, i am afraid you didnt see my comment about adding a ref in changelog

Gabriel439:
@jneira: Which comment was that?

Also, it's not too late to fix the change log.  The GitHub changelog is mutable and I can also fix the changelog so that the fix will appear in the next release to Hackage

PR #694, opened 2018-11-21
Fixed mixed field access

### LINKED ISSUE #692: The new union access is not transitive across record field
This dhall definition works as intended:
```
let Scope = < Public : {} | Private : {} >
in Scope.Public {=}
```
However, puting the union in a record field doesnt work (at least as i expected):
```
let Scope = < Public : {} | Private : {} >
let types = { Scope = Scope }
in types.Scope.Public {=}
```
It throws:
```

 Error : Not a record or a union

   types.Scope.Public
(stdin):3:4
```

Gabriel439:
Fixes #692

The standard permits a user to access a constructor from a type stored inside
a record, but the Haskell implementation had a mistake which prevented this.
Specifically, the Haskell implementation was not normalizing the union type
as the standard specified before attempting to access the constructor, leading
to an unexpected type error.

### DIFF
--- a/dhall/src/Dhall/TypeCheck.hs
+++ b/dhall/src/Dhall/TypeCheck.hs
@@ -791,12 +791,12 @@ typeWithA tpa = loop
                     Just t' -> return t'
                     Nothing -> Left (TypeError ctx e (MissingField x t))
             Const Type -> do
-                case r of
-                  (Note _ (Union kts)) ->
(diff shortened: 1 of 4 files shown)


Gabriel439:
Note that I plan to upstream the regression test I added to the standard, but I want to merge this fix before cutting the release today.

Issue #642, opened 2018-10-16
Loading from Cache Typechecks Again


FintanH:
It seems like we're not getting any wins from caching when it comes to caching. Correct me if I'm wrong (very likely 😄) but when we run `dhall <<< ./file` it will [resolve](https://github.com/dhall-lang/dhall-haskell/blob/master/src/Dhall/Main.hs#L273) the expressions, this is where loading from the cache comes in, but it still [typechecks](https://github.com/dhall-lang/dhall-haskell/blob/master/src/Dhall/Main.hs#L275) everything that was loaded. Could there be a way to keep the the inferred types of the cached files as well?

sellout:
So, we could potentially always cache `Annot`, but perhaps add a `checked` boolean to `Annot`, such that `typeWithA` doesn’t `loop` over `Annot` nodes that have been loaded from cache (i.e., `checked = True`)?

Gabriel439:
We could disable type-checking for things imported from the cache if we can lock down permissions such that only the user has read-write access to it.  We already type-check things before we cache them anyway

FintanH:
> We already type-check things before we cache them anyway

Ya, the hope is not to do that work the second time when loading from the cache 😁

Gabriel439:
Oh, I remember why we type-check things now.  Type-checking is necessary so that we can αβ-normalize the expression to verify the semantic integrity check.  So if we disable type-checking for imported expressions that implies disabling signature verification, too.

However, that could be an option.  Again, we can trust the cache if we can lock it down and assume cached expressions have been verified to match the semantic integrity check.

FintanH:
Does the semantic integrity check not happen before the place I pointed out above?

Gabriel439:
@FintanH: Oh, sorry.  I misunderstood which type-checking step you are referring to.

Yeah, this would require more work to prevent type-checking the resolved expression at least once.

Before deciding on a particular strategy, I would like to know:

* How large is the expression being type-checked (in bytes for the cached form)?
* How long does it take to type-check the expression after decoding it from the cache?

The reason I ask the first question is that I want to see if the problem is due to the expression having a very large normal form, because if that is the case maybe we can apply some language design to lead to more compact normal forms.

The second I ask the second question is to see if type-checking is in the right ballpark for performance or not, because I usually only resort to caching as a last resort for resolving performance issues.

FintanH:
[...]

Issue #580, opened 2018-09-16
Performance issues with many deeply nested imports


f-f:
One of the examples in the `dhall-kubernetes` repo ([this one](https://github.com/dhall-lang/dhall-kubernetes/blob/import-hash-bug/examples/deployment.dhall) - if you'd like to test it note that it's on a separate branch) takes roughtly 25s to evaluate on my machine.  
It is not a big deal for me, but after showing it in [the talk's demo](https://twitter.com/fabferrai/status/1040523416112128000) the other day the single most common question I got afterwards was "why does it take so long?" (N = ~20 people)

I think (I did not profile the thing actually, I'll do it asap) the problem comes from the fact that the Kubernetes types and defaults are nesting many imports (e.g. try to evaluate the [Deployment type](https://github.com/dhall-lang/dhall-kubernetes/blob/master/types/io.k8s.api.apps.v1.Deployment.dhall), it's quite huge), and Dhall will typecheck and normalize the imported expression before substituting it in the AST of the current expression (that then will be typechecked and normalized again, and so on for all the levels of nesting).

This work is done because [we are caching the imports in memory](https://github.com/dhall-lang/dhall-lang/blob/master/standard/semantics.md#duplicate-imports) (so we have to typecheck and normalize every import to possibly reuse it), but I wonder if we could skip some of this work (e.g. by resolving all nested imports first, then composing everything together, and then typechecking and normalizing the now-pure expression)

quasicomputational:
Minimising the re-typechecking/normalising we do is a very worthy goal. This might also help `dhall-to-cabal` significantly.

sellout:
This may be my fault. I made a couple changes a number of months ago that
1. cached imports more reliably and
2. pre-normalized an import when caching.

At the time, for my use cases, pre-normalizing sped things up like 10×. I don’t recall exactly where the benefit came from – perhaps just pre-resolving would give us the same benefit? Or maybe it’s just different styles of import patterns (but I feel like deeply-nested repeated imports was exactly our problem case).

I wonder if something like an annotation indicating that a sub-tree has already been normalized would eliminate the cost of re-traversing?

Gabriel439:
@f-f: Have you tested what performance you get when you don't prenormalize imported expressions?

[...]

PR #698, opened 2018-11-23
Generate dot graph to visualize import graph

basile-henry:
This PR adds a small option to `dhall resolve` that outputs the dependency graph at the import level in `dot` format.

It is a similar feature to [`stack`'s dependency visualization](https://docs.haskellstack.org/en/stable/dependency_visualization/) but at the module/import level as opposed to the package level.

If the feature is of any interest, I'd love to improve it and make it more user friendly! :smile: 

Here is what it currently looks like on `./Prelude/package.dhall` (in `.png` as GitHub doesn't seem to support `.svg` uploads):

```sh
dhall resolve --dot <<< "./Prelude/package.dhall" |
  dot -Tpng > dhall-prelude.png
```

dhall-prelude

There is currently a small issue with diamond dependencies:

diamond

`./c.dhall` is pointed at twice. I have some idea on how to fix that, and I'll try to implement that, but if you have some advice on how I should go about it, I'm all ears!
I'm also open to hearing all your suggestions to make the graph look better! :smile:


dhall/src/Dhall/Import.hs
```diff
@@ -766,10 +767,25 @@ loadWith expr₀ = case expr₀ of
...
                     zoom stack (State.put imports)
 
+                    zoom dot (State.modify $ \getSubDot -> do
+                        parentNode <- _dot
+
+                        -- Get current node back from sub-graph
+                        hereNode <- getSubDot
+
+                        -- Add edge between parent and here
+                        () <- parentNode .->. hereNode
```
Gabriel439: Minor suggestion: I don't think the `() <-` is necessary

basile-henry: Ah yeah of course :man_facepalming:

basile-henry:
@Gabriel439 Is it possible the caching mechanism for import doesn't work as it should?

With:

```diff
diff --git a/dhall/src/Dhall/Import.hs b/dhall/src/Dhall/Import.hs
index 15760b5..75d0372 100644
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -198,6 +198,8 @@ import qualified Text.Megaparsec
 import qualified Text.Parser.Combinators
 import qualified Text.Parser.Token
 
+import Debug.Trace
+
 -- | An import failed because of a cycle in the import graph
 newtype Cycle = Cycle
     { cyclicImport :: Import  -- ^ The offending cyclic import
@@ -732,7 +734,7 @@ loadWith expr₀ = case expr₀ of
     expr <- if here `elem` canonicalizeAll imports
         then throwMissingImport (Imported imports (Cycle import_))
         else do
-            case Map.lookup here _cache of
+            case Map.lookup here $ traceShowId _cache of
                 Just expr -> return expr
                 Nothing   -> do
                     -- Here we have to match and unwrap the @MissingImports@

```

```sh
➜ tail -n +1 -- *.dhall
==> a.dhall <==
./c.dhall

==> b.dhall <==
./c.dhall

==> c.dhall <==
./d.dhall

==> d.dhall <==
42

==> diamond.dhall <==
{ a = ./a.dhall
, b = ./b.dhall
}

➜ cabal new-run -- dhall <<< "./diamond.dhall"
Up to date
fromList []
fromList []
fromList []
fromList []
fromList [(Import {importHashed = ImportHashed {hash = Nothing, importType = Local Here (File {directory = Directory {components = []}, file = "a.dhall"})}, importMode = Code},NaturalLit 42)]                     
fromList [(Import {importHashed = ImportHashed {hash = Nothing, importType = Local Here (File {directory = Directory {components = []}, file = "a.dhall"})}, importMode = Code},NaturalLit 42)]                     
fromList [(Import {importHashed = ImportHashed {hash = Nothing, importType = Local Here (File {directory = Directory {components = []}, file = "a.dhall"})}, importMode = Code},NaturalLit 42)]                     
{ a = 42, b = 42 }
```

I would expect to see other files than just `a.dhall` in the cache. The way the imports are being traversed (depth first) isn't making it easy, but this looks like a bug to me.

What do you think?

Gabriel439:
@basile-henry: Yes, it is!  I think you may have found the long-sought cause of Dhall's inefficient import resolution.  I have the fix up here: https://github.com/dhall-lang/dhall-haskell/pull/702

f-f:
Oh wow, nice :)

@Gabriel439 could this be the cause of #642? (I recall there was another issue like "we do a suspiciously big amount of operations when resolving imports", but cannot find it anymore)

Gabriel439:
@f-f: You might be thinking of https://github.com/dhall-lang/dhall-haskell/issues/580

f-f:
No, I think it was older than that, and I think we closed it when fixing some other bug related to imports. So it looks like we don't have any issue open for inefficient imports, which is good :)

basile-henry:
Bonus graph `dhall-to-cabal`:

dhall-to-cabal

Do you think I should use relative imports in the graph? Would that make it more readable? :smile:

dhall/src/Dhall/Import/Types.hs
```diff
@@ -39,9 +41,12 @@ data Status m = Status
     -- ^ Stack of `Import`s that we've imported along the way to get to the
     -- current point
 
-    , _cache :: Map Import (Expr Src X)
-    -- ^ Cache of imported expressions in order to avoid importing the same
-    --   expression twice with different values
+    , _dot :: (Int, Dot NodeId)
```
Gabriel439: Rather than use `_dot :: (Int, Dot NodeId)` I would suggest two separate fields, such as `_dot :: Dot NodeId` and `_nodeId :: Int`

PR #700, opened 2018-11-23
Add support for records of records of types

Gabriel439:
... as standardized in https://github.com/dhall-lang/dhall-lang/pull/300

Fixes https://github.com/dhall-lang/dhall-haskell/issues/692


Gabriel439:
Note that I will hold off merging this until after cutting the 1.19.1 release tonight

PR #713, opened 2018-11-26
Unique dependencies in dot graph

basile-henry:
Sometimes the same import is used multiple times in a given `dhall` file. This just adds extra noise in the generated graph.

For example here:

https://github.com/dhall-lang/dhall-to-cabal/blob/8967220bd3618c876bbc4f7b0c93d49fa47a30ee/dhall/unconditional.dhall#L21-L26

2018-11-26-152926_1196x79_scrot

It results in 4 connections between two nodes where 1 would be enough.

By making the generated graph `strict` we prune away all the redundant connection making the graph more readable.

With `dhall-to-cabal` as a *real world example*:

Before this PR:

old-dhall-to-cabal

After this PR:

new-dhall-to-cabal

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -9,6 +9,7 @@ extra-deps:
   - repline-0.2.0.0
   - serialise-0.2.1.0
   - neat-interpolation-0.3.2.4
+  - dotgen-0.4.2@sha256:309b7cc8a3593a8e48bee7b53020d5f72db156d58edf78a0214f58fbb84b292b
 nix:
   packages:
     - ncurses
--- a/dhall/src/Dhall/Main.hs
+++ b/dhall/src/Dhall/Main.hs
@@ -300,7 +300,7 @@ command (Options {..}) = do
                 State.runStateT (Dhall.Import.loadWith expression) status
 
             if   dot
-            then putStr . Text.Dot.showDot $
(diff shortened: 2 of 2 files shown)


stack.yaml
```diff
@@ -9,6 +9,7 @@ extra-deps:
   - repline-0.2.0.0
   - serialise-0.2.1.0
   - neat-interpolation-0.3.2.4
+  - dotgen-0.4.2@sha256:309b7cc8a3593a8e48bee7b53020d5f72db156d58edf78a0214f58fbb84b292b
```
Gabriel439: What is the reason for selecting this version of `dotgen`?

basile-henry: It's the latest one (and it was available on `nixpkgs`). Also according to the changelog it contains a fix to build with GHC 7.10: https://hackage.haskell.org/package/dotgen-0.4.2/changelog
Do you think we should relax the lower version bound?

The `stack.yaml` fix should have really gone in the previous PR, so I thought I would include it here, but it's not really the focus of this short PR. Would you want me to split it in two PRs?

Gabriel439: @basile-henry: It's fine if you include the change in this pull request.  I was just curious

You don't need to relax the lower version bound unless people specifically ask for it

PR #717, opened 2018-11-27
Add `--json` flag for `dhall {decode,encode}`

### LINKED ISSUE #714: Add encode-json command
Now that we have the parser tests in the `dhall-lang` repo, there is the issue of keeping them up to date as the standard evolves. This is slightly annoying currently, as the "B side" of a Dhall expression is its CBOR-JSON form; last time (when generating the tests) I tried using some thing from npm to translate the binary encoding from `dhall encode` to JSON, but that was kinda broken.

For this reason it would probably make sense to add a `dhall-to-json encode` command, that would render a Dhall expression in its CBOR-JSON form, so I guess we'd basically need a function `Codec.CBOR.Term -> Data.Aeson.Value`

Example output:
```bash
$ dhall-to-json encode <<< "λ(x : Bool) → x"
[
  1,
  "x",
  "Bool",
  "x"
]
```

Gabriel439:
Fixes #714

This adds a `--json` flag that `dhall decode` and `dhall encode` can use to
read/write the equivalent JSON representation of the CBOR.  This comes in
handy for the parsing compliance tests which use this CBOR-as-JSON as the
standard representation of the abstract syntax tree.

### DIFF
--- a/dhall/dhall.cabal
+++ b/dhall/dhall.cabal
@@ -296,10 +296,13 @@ Library
     Hs-Source-Dirs: src
     Build-Depends:
         base                        >= 4.8.2.0  && < 5   ,
+        aeson                       >= 1.0.0.0  && < 1.5 ,
+        aeson-pretty                               < 0.9 ,
         ansi-terminal               >= 0.6.3.1  && < 0.9 ,
         bytestring                                 < 0.11,
         case-insensitive                           < 1.3 ,
         cborg                       >= 0.2.0.0  && < 0.3 ,
+        cborg-json                                 < 0.3 ,
         containers                  >= 0.5.0.0  && < 0.7 ,
         contravariant                              < 1.6 ,
         cryptonite                  >= 0.23     && < 1.0 ,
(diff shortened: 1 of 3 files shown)


Gabriel439:
@f-f: Ah, CI caught a problem with this.  `cborg-json` requires a version of `base` newer than what GHC 7.10.3 supports (cc: @jneira)

jneira:
Mmm, `cborg-json-0.2.10` uses `base (>=4.7 && <4.13)` and `ghc-7.10.3`, `base = 4.8.2.0` so, in theory, they could be compatible...

Gabriel439:
Oh, I completely misunderstood what was going on.  It's actually the exact opposite problem: ghc-8.6 uses a version of `base` that is too new for `cborg-json`.  Let me see if I can get `cborg-json` to bump its upper bound on `base`

Gabriel439:
Looking more closely, `cborg-json` is not on Stackage so I need to see if the maintainers would be okay with adding it to Stackage

Gabriel439:
Here is the request to add `cborg-json` to Stackage: https://github.com/well-typed/cborg/issues/191

Gabriel439:
Alright, `cborg-json` was added to Stackage nightly and I fixed `dhall` to use the latest version which has the expanded bounds on `base` so this is good to merge now

Issue #707, opened 2018-11-25
Add :quit command for `dhall repl`


Gabriel439:
Currently the only way to exit from the `dhall repl` is by typing `Ctrl-D`.  We should add a `:quit` command for this purpose as well

Gabriel439:
Fixed by #719

PR #719, opened 2018-11-28
Add :quit repl command

kedashoe:
For #707 

Not sure what the "best" way to quit is, `throw Interrupt` seems a bit odd to me but I couldn't figure out a better way..


dhall/src/Dhall/Repl.hs
```diff
@@ -234,6 +235,10 @@ saveBinding (file : "=" : tokens) = do
   liftIO (System.IO.withFile file System.IO.WriteMode handler)
 saveBinding _ = fail ":save should be of the form `:save x = y`"
 
+cmdQuit :: ( MonadIO m, MonadState Env m ) => [String] -> m ()
+cmdQuit _ = do
+  liftIO (putStrLn "Goodbye.")
+  throw Interrupt
```
Gabriel439: You usually want to prefer `throwIO` instead of `throw` since `throw` can accidentally be triggered by evaluation whereas `throwIO` is evaluation-safe.  Or in other words:

```haskell
throw x `seq` () = throw x

throwIO x `seq` () = ()
```

```suggestion
  liftIO (throwIO Interrupt)
```

kedashoe: ah so technically it is possible it could throw without `putStrLn`ing? Updated.

Gabriel439: Yeah!  The `repline` loop has logic that checks if the last command was non-empty when an `Interrupt` was thrown.  If the command is empty then it says `Goodbye.` and if it was non-empty then it says nothing.  This is why you had to insert an explicit `Goodbye.` to match the behavior because `:quit` was a non-empty command.

Gabriel439:
Thank you for contributing this! 🙂

PR #729, opened 2018-12-02
Change how `dhall repl` handles prior definitions

Gabriel439:
The motivation for this change is so that more code type-checks
within a REPL session.

For example, before this change the following session fails:

```haskell
⊢ :let Either = λ(a : Type) → λ(b : Type) → < Left : a | Right : b >
⊢ :let Example = < Left : Natural | Right : Bool >

Example : Type

⊢ Example.Left

Example : Type

Error: Not a record or a union

Example.Left
(stdin):1:1
```

After this change, the same session works:

```haskell
⊢ :let Example = < Left : Natural | Right : Bool >

Example : Type

⊢ Example.Left

λ(Left : Natural) → < Left = Left | Right : Bool >

```

The difference is in how we incorporate prior definitions when
type-checking code.  Before this change prior definitions were added
to the context, which leads to type-checking failures because type
synonyms don't work when used in this way.  After this change we
incorporate them as extra `let` definitions (same as normalization).

### DIFF
--- a/dhall/src/Dhall/Repl.hs
+++ b/dhall/src/Dhall/Repl.hs
@@ -1,8 +1,9 @@
 -- | This module contains the implementation of the @dhall repl@ subcommand
 
-{-# language FlexibleContexts #-}
-{-# language NamedFieldPuns #-}
+{-# language FlexibleContexts  #-}
+{-# language NamedFieldPuns    #-}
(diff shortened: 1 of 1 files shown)


Gabriel439:
cc: @basile-henry since you may have run into this issue when testing union alternative completion in #727

PR #739, opened 2018-12-06
Add "Try Dhall" project

Gabriel439:
This adds two new Nix build products:

* `try-dhall-static` - The static assets for "Try Dhall"
* `try-dhall-server` - A script which serves the static assets for local
   debugging

This is powered by a new small `dhall-try` package which is also included
in this change.

try dhall


Gabriel439:
Also, if anybody wants to try this out before this is ready to go live, you can download the latest tarball of the static assets here:

https://hydra.dhall-lang.org/job/dhall-haskell/master/tarball-try-dhall/latest

Just unpack the tarball and open `index.html` in your browser to try out the demo

PR #740, opened 2018-12-07
Fix diff so that an empty list is equal to itself

madjar:
While using `dhall diff`, I realized that empty lists are considered to be different. This PR contains two commits:
- The first one adds a quickcheck test to ensure an expression is always the same as itself (as long as it typechecks ; things go weird when you have an invalid expression, I guess because `Diff` assume the expression is well-formed).
- The second one is the actual fix.

Sadly, the test required me to expose some previously private part of `Dhall.diff`. If that's a problem, feel free to cherry-pick the second commit.

### DIFF
--- a/dhall/src/Dhall/Diff.hs
+++ b/dhall/src/Dhall/Diff.hs
@@ -11,7 +11,9 @@
 
 module Dhall.Diff (
     -- * Diff
-      diffNormalized
+      Diff (..)
+    , diffExpression
+    , diffNormalized
     , Dhall.Diff.diff
     ) where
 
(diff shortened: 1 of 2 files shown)


madjar:
The story behind that: I'm converting a complex json file to dhall. To ensure nothing is lost, I converted it to dhall with https://gist.github.com/madjar/252c517644c0e13ef28a2a7ca71f5fa4, and I'm comparing every change against the original with `dhall diff`. I was a bit distressing to get an enormous diff when passing twice the same file :)

Gabriel439:
@madjar: Thank you!

madjar:
@basile-henry I don't see the issue with `Text`

```
> let x = TextLit (Chunks [] "") in same (diffExpression x x)
True
```

PR #747, opened 2018-12-10
Expand "Try dhall" into `dhall-lang.org` home page

Gabriel439:
This expands the "Try dhall" page to serve as a functional home page for
"dhall-lang.org" in the short term by making the following changes:

* Adding a navigation bar to the top that links to useful resources and
  official integrations
* Adding a quick summary explaining what Dhall is

Here is an example of what it looks like:

screenshot_20181209_181946

### DIFF
--- a/dhall-try/index.html
+++ b/dhall-try/index.html
@@ -34,6 +34,14 @@
       .nav-link {
         outline: 0;
       }
+
+      .navbar span {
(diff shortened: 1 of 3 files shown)


Gabriel439:
I made some additional changes to expand the Navbar:

try dhall - 3

PR #749, opened 2018-12-10
Add Unordered Map Traversal

### LINKED ISSUE #640: Better implemented Map operations
I would argue that https://github.com/dhall-lang/dhall-haskell/blob/master/src/Dhall/Map.hs#L463 i.e.:
```haskell
Data.Functor.void (traverseWithKey f m)
```

is wasteful work as `traverseWithKey` calls `toList` and `fromList` which isn't necessary. We could implement this as something more along the lines of:

```haskell
traverseWithKey_ f (Map m _) = Data.Function.void (Data.Map.traverseWithKey f m)
```

I would also think that https://github.com/dhall-lang/dhall-haskell/blob/master/src/Dhall/Map.hs#L449 could perform better if it was implemented as:

```haskell
traverseWithKey f (Map m ks) = fmap (\m' -> Map m' ks) (Data.Map.traverseWithKey f m)
```

FintanH:
Fixes #640 

* Added `unorderedTraverseWithKey_`
* Added benchmarks to confirm performance of different map traversals
* Use `unorderedTraverseWithKey_` in `TypeCheck.hs`


FintanH:
Oh hahaha I was wondering why I couldn't push after merging :joy: Just saw you merged master @Gabriel439

Gabriel439:
@FintanH: Thank you! 🙂

PR #754, opened 2018-12-13
Expand main page with high-level reasons to adopt Dhall

Gabriel439:
This adds four new sections to the page after the live demo which highlight
the common themes that I notice people use when communicating the value of
Dhall to others on social media:

* The first section emphasizes the element of delight in using the language for
  people who are into elegance and quality

* The second section focuses on more pragmatic people who are sick of YAML and
  just want a reasonable alternative that they can convince their manager to
  adopt

* The third section appeals to the LangSec crowd that wants an uncompromising
  and secure foundation for what they are buliding

* The last section targeted at the skeptic who thinks that Dhall is an ivory
  tower language not suited for real-world problems.

The second crowd (YAML emigrants) is the audience that I'm targeting the
most strongly at the moment, but I didn't want to lead with a negative reason
adopt by focusing on the limitations of YAML, so I put the section on delight
first so that we could start with a more positive tone.

This is probably the last change that I have planned before publishing this page
as the new `dhall-lang.org`

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,12 +104,12 @@ Building the website from source is currently only supported for Nix on Linux.
 You can build the static assets by running:
 
 ```bash
-$ nix-build --attr try-dhall-static
+$ nix-build --attr try-dhall
 ```
 
 ... then open `./result/index.html` in your browser.
 
-You can also download archive containing the pre-built website from CI using
+You can also download an archive containing the pre-built website from CI using
 this link:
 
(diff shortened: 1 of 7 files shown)


Gabriel439:
Almost forgot the recording:

try dhall - 4

basile-henry:
It looks great! :+1: 

I am however missing some images when looking at it after a `nix-build`

2018-12-13-114836_595x456_scrot

Issue #764, opened 2018-12-23
Space leak in dhall 1.19.1


vmchale:
I have encountered a space leak in Dhall 1.19.1 (I do not know the cause). The particular Dhall file I was checking was [pkg-set.dhall](https://github.com/vmchale/cpkg/blob/master/pkgs/pkg-set.dhall). 

Running `cat pkg-set.dhall | dhall-profile +RTS -h` yielded the following heap profile:

![dhall-profile](https://user-images.githubusercontent.com/13259982/50379868-c7f33800-061b-11e9-88b7-21f258bbbefc.png).

As you can see, this takes nearly 600MB to typecheck/normalize/pretty-print the expression! I assume this is a bug.

Thanks again.

PR #765, opened 2018-12-24
Prenormalize substituted expressions

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/764

This pre-normalizes expressions before substituting them to
avoid duplicated work.  This also fixes the space leak from the
above issue.

### DIFF
--- a/dhall/src/Dhall/Core.hs
+++ b/dhall/src/Dhall/Core.hs
@@ -1714,15 +1714,18 @@ normalizeWithM ctx e0 = loop (denote e0)
                         case res2 of
                             Nothing -> pure (App f' a')
                             Just app' -> loop app'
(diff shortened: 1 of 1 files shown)


Gabriel439:
Also, just in case people are curious why we didn't do this prenormalization before, it was due to the (now-defunct) `constructors` keyword.  Before, the `constructors` keyword could expand to a large record which was an expensive expression to substitute into or to be substituted into.  Deferring normalization improved the efficiency of `constructors`-related code, but at the expense of other code being slower.

However, since we don't generate records of constructors any longer there is less of a reason to do this deferred normalization.  There might still be some expressions where it's cheaper to defer the normalization work until after substitution, but this is significantly less likely than before.

PR #771, opened 2018-12-30
Update standard version to 5.0.0 in Binary.hs

MonoidMusician:
It looks like this was missed in the latest release?

As far as I can tell, this is the only place is needs to be changed.


Gabriel439:
Thanks for fixing this!  I will upload a new release soon for this fix

MonoidMusician:
Sounds good, thank you!

PR #775, opened 2019-01-07
`UnionType` and `UnionInputType`, analogues of `RecordType` and `RecordInputType`

mstksg:
Allows us to generate an `Input` for a union:

```
data Status = Queued Natural
            | Result Text
            | Errored Text

status :: Type Status
status =
  union
    ( (Queued  <$> constructor "Queued"  natural)    -- or `constructor "Queued" (Queued <$> natural)`
   <> (Result  <$> constructor "Result"  string )
   <> (Errored <$> constructor "Errored" string )
    )
```

And (potentially partially) create an `InputType`:

```
injectStatus :: InputType Status
injectStatus =
  inputUnion
    (  inputConstructor "Queued"  (\case Queued n  -> Just n; _ -> Nothing)
    <> inputConstructor "Result"  (\case Result t  -> Just t; _ -> Nothing)
    <> inputConstructor "Errored" (\case Errored e -> Just e; _ -> Nothing)
    )
-- (could potentially be simplified with a prism-based interface, but this would incur a profunctors dep)
```

I couldn't really think of a way to write `inputUnion` in a way that would make sure it was total without a "two-pass" solution, where the user provides the "injecting" logic and the "specify type" logic in two different places; one alternative was a quasi-one-pass system where the user specifies the total logic up-front, but the result can be optionally "validated" by providing a proof of totality before being used.  However, that requires encoding the number of branches in the type level, so I'm not sure if the benefit pays for the cost in complexity.

Let me know if this isn't something that fits in the vision of the library, or if there's anything I could fix :)

### DIFF
--- a/dhall/src/Dhall.hs
+++ b/dhall/src/Dhall.hs
@@ -10,6 +10,7 @@
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE StandaloneDeriving         #-}
 {-# LANGUAGE TypeOperators              #-}
+{-# LANGUAGE TupleSections              #-}
 
 {-| Please read the "Dhall.Tutorial" module, which contains a tutorial explaining
     how to use the language, the compiler, and this library
@@ -1579,7 +1580,7 @@ newtype UnionInputType a =
                 ( Expr Src X )
             )
         )
-        ( Op (Expr Src X) )
+        ( Op (Text, Expr Src X) )
         a
     )
   deriving (Contravariant)
@@ -1592,7 +1593,7 @@ newtype UnionInputType a =
 -- itself (since no instance is possible).
 (>|<) :: UnionInputType a -> UnionInputType b -> UnionInputType (Either a b)
 UnionInputType (Data.Functor.Product.Pair (Control.Applicative.Const mx) (Op fx))
-    >|< UnionInputType (Data.Functor.Product.Pair (Control.Applicative.Const my) (Op fy)) = 
+    >|< UnionInputType (Data.Functor.Product.Pair (Control.Applicative.Const my) (Op fy)) =
     UnionInputType
       ( Data.Functor.Product.Pair
           ( Control.Applicative.Const (mx <> my) )
@@ -1603,9 +1604,10 @@ infixr 5 >|<
 
 inputUnion :: UnionInputType a -> InputType a
 inputUnion ( UnionInputType ( Data.Functor.Product.Pair ( Control.Applicative.Const fields ) ( Op embedF ) ) ) =
-    InputType 
-      { embed =
-          embedF
+    InputType
+      { embed = \x ->
+          let (name, y) = embedF x
+          in  UnionLit name y (Dhall.Map.delete name fields)
       , declared =
           Union fields
       }
@@ -1622,7 +1624,7 @@ inputConstructorWith name inputType = UnionInputType $
               ( declared inputType )
           )
       )
-      ( Op ( embed inputType )
+      ( Op ( (name,) . embed inputType )
       )
 
 inputConstructor
--- a/dhall/src/Dhall.hs
+++ b/dhall/src/Dhall.hs
@@ -97,7 +97,7 @@ import Control.Exception (Exception)
 import Control.Monad.Trans.State.Strict
 import Control.Monad (guard)
 import Data.Coerce (coerce)
-import Data.Functor.Contravariant (Contravariant(..), (>$<))
+import Data.Functor.Contravariant (Contravariant(..), (>$<), Op(..))
 import Data.Functor.Contravariant.Divisible (Divisible(..), divided)
 import Data.Maybe (fromMaybe, listToMaybe, mapMaybe)
 import Data.Monoid ((<>))
@@ -124,6 +124,7 @@ import qualified Control.Monad.Trans.State.Strict as State
 import qualified Data.Foldable
 import qualified Data.Functor.Compose
 import qualified Data.Functor.Product
+import qualified Data.Monoid
 import qualified Data.Scientific
 import qualified Data.Sequence
 import qualified Data.Set
@@ -1423,8 +1424,8 @@ instance Monoid (UnionType a) where
     mempty = coerce (mempty :: Dhall.Map.Map Text (Type a))
 
 -- | Run a 'UnionType' parser to build a 'Type' parser.
-union :: Dhall.Map.Map Text (Type a) -> Type a
-union mp = Type
+union :: UnionType a -> Type a
+union (UnionType (Data.Functor.Compose.Compose mp)) = Type
     { extract  = extractF
     , expected = Union expect
     }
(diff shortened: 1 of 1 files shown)


dhall/src/Dhall.hs
```diff
@@ -1371,6 +1381,66 @@ field key valueType =
...
+  deriving (Functor)
+
+instance Semigroup (UnionType a) where
+    (<>) = coerce ((<>) :: Dhall.Map.Map Text (Type a) -> Dhall.Map.Map Text (Type a) -> Dhall.Map.Map Text (Type a))
+
+instance Monoid (UnionType a) where
+    mempty = coerce (mempty :: Dhall.Map.Map Text (Type a))
+
+-- | Run a 'UnionType' parser to build a 'Type' parser.
+union :: Dhall.Map.Map Text (Type a) -> Type a
```
Gabriel439: Should the type signature of `union` be:

```haskell
union :: UnionType a -> Type a
```

... in order to be compatible with `constructor`?

dhall/src/Dhall.hs
```diff
@@ -1453,4 +1523,84 @@ inputRecord (RecordInputType inputTypeRecord) = InputType makeRecordLit recordTy
...
+    UnionInputType x <> UnionInputType y = UnionInputType (x <> y)
+
+instance Monoid (UnionInputType a) where
+    mempty = UnionInputType mempty
+
+inputUnion :: UnionInputType a -> InputType a
+inputUnion (UnionInputType inputTypeUnion) = InputType makeUnionLit (Union declare)
+  where
+    declare = fst <$> inputTypeUnion
+    makeUnionLit x = fromMaybe (errorWithoutStackTrace unmatched)
```
Gabriel439: Would it be possible to avoid the partial implementation by changing `UnionInputType` to:

```haskell
data UnionInputType a = UnionInputType (Map Text (Expr Src X)) (a -> Maybe (Map Expr Src X))
```

dhall/src/Dhall.hs
mstksg: Ah right, thanks!  This was a remnant from an early draft, but this is the intended signature.

dhall/src/Dhall.hs
mstksg: This representation makes more sense, since we only consume injector in the asum'd form.  I don't think that this directly helps us avoid partiality.

However, this representation gives a hint about how we could make it non-partial; we might be able to write a variant of `inputUnion` that gives a fallback option that would be used if anything was `Maybe`.

mstksg: 17fd6c3 now incorporates this new representation!  It's actually nicer because we get a derivable Contravariant and almost a derivable Monoid instance (waiting for base-4.12, which has Monoid for `:*:`).

Still curious about what you had in mind for resolving the partiality issue, though! :)

ocharles:
At work, we are using what I suppose is a simpler variation of this:

```haskell
makeUnion :: Map.InsOrdHashMap LazyText.Text ( Dhall.Type a ) -> Dhall.Type a
makeUnion alts =
  let
    extract expr = do
      Expr.UnionLit ctor v _ <-
        return expr

      t <-
        Map.lookup ctor alts

      Dhall.extract t v

    expected =
      Expr.Union ( Dhall.expected <$> alts )

  in Dhall.Type { .. }
```

Gabriel439:
@mstksg: I believe you can fix the partiality using a trick similar to `RecordInputType`, except replacing `Divisible` with `Decidable`.

In other words, the user interface we supply looks like this:

```haskell
-- I'm not sure if this is the right representation to use, so focus more
-- on the type signatures than the implementation
newtype UnionInputType a = UnionInputType (Map Text (InputType a))

inputConstructorWith :: Text -> InputType a -> UnionInputType a
inputConstructorWith k v = UnionInputType (Data.Map.singleton k v)

inputConstructor :: Inject a => Text -> UnionInputType a
inputConstructor k = inputConstructorWith k inject

instance Contravariant UnionInputType where …
instance Divisible     UnionInputType where …
instance Decidable     UnionInputType where …

(>|<) :: Decidable f => f a -> f b -> f (Either a b)
(>|<) = Data.Functor.Contravariant.chosen

-- I haven't thought through if this can be implemented.  Feel free to
-- change the implementation of `UnionInputType` if this function isn't
-- total
inputUnion :: UnionInputType a -> InputType a
inputUnion = …
```

... so that the user can write this:

```haskell
injectStatus :: InputType Status
injectStatus =
          adapt
    >$< inputConstructor "Queued"
    >|< inputConstructor "Result"
    >|< inputConstructor "Errored"
  where
    adapt (Queued  n) = Left (Left n)
    adapt (Result  t) = Left (Right t)
    adapt (Errored e) = Right e
```

mstksg:
A Decidable-inspired interface might work!  Although I don't think we could actually use `Decidable`, because the `Divisible` instance might not necessarily make sense (this is the reason also why I use `Monoid` instead of `Alternative` for `UnionType`).  On that note, we could actually maybe be able to unify `RecordType` and `UnionType`, to some extent, by providing a type that is *both* `Applicative` and `Alternative`, and also unify `RecordInputType` and `UnionInputType` by providing a type that is *both* `Divisible` and `Decidable`.  These would basically allow the user to construct/define records of unions of records of records of unions of (etc.), of arbitrary nesting.  This isn't actually too bad, but it might be difficult to predict the resulting shape for non-complex usage.

I'll see if I can whip up a "Decidable-inspired" interface (using an analog of `chosen`), and also maybe play around with arbitrary nesting so we could unify `RecordType` and `UnionType` perhaps, if that might be cleaner.

mstksg:
The Decidable-inspired method (implemented in cc140c3) works and ensures totality! :D 

However, as feared, it is impossible to actually use `Decidable`, because `Decidable` requires `Divisible`.  The issue is `conquer`; isn't any way to implement:

```
conquer :: forall a. UnionInputType a
```

because this would allow the user to construct something silly like a `UserInputType Natural`, but without specifying any constructor names.  `inputUnion (conquer @Natural)` is unimplementable in a total way.

However, dropping in a custom `>|<` detached from `chosen` works, and we can re-use `>$<` for `Contravariant`.

One small note: it's *possible* to implement `lose :: forall a. (a -> Void) -> UnionType a`; however, I'm not sure if there is any benefit to doing so, so it might not be worth writing out and providing.

I'll still continue to play around with a combined `RecordUnionType` and `RecordUnionInputType` where `<*>`/`>*<` correspond to combining records, and `<|>`/`>|<` correspond to combining unions in the meantime, but for now I think this version's API isn't too bad already, with the partiality handled :)

Gabriel439:
@mstksg: Yeah, a monomorphic `(>|<)` is fine

Also, this branch still needs to import `Semigroup` for older versions of GHC before CI will pass

mstksg:
New changes should appease GHC 8.0 and above!  However, I couldn't get the dependencies to work with GHC 7.10, but I'm not sure if these versions were still officially supported.

Gabriel439:
@mstksg: Are you sure your changes don't work with GHC 7.10?  CI passed and that requires GHC 7.10 to work

The reason Dhall supports GHC 7.10 is for Eta since Dhall is a supported configuration format for Eta packages and Eta is built using GHC 7.10

Gabriel439:
I will go ahead and merge this since CI thinks this is GHC 7.10-ready

mstksg:
Ah, nice!  I couldn't get to the build part, I got stuck in the dependency resolution part.  But if it works, then I'll take your word for it :)  Thanks!

PR #780, opened 2019-01-15
Fix non-exhaustive pattern match in `dhall lint`

Gabriel439:
`dhall lint` would fail on the following expression:

```
    let replicate = https://prelude.dhall-lang.org/List/replicate

in  let Config = { name : Text, age : Natural }

in  let Configs = List Config

in  replicate 10 Text "!"
```

... because the code (incorrectly) assumed that simplifying an inner
`let` binding would preserve at least one `let` binding.  However, when the
outer `let` (beginning with `let replicate`) is simplified the inner `let`
(beginning with `let Config`) simplifies down to just `replicate 10 Text "!"`
which has no `let` binding at all, leading to a pattern match failure.

This change fixes that by extending the code to correctly handle that case
with an exhaustive pattern match.

### DIFF
--- a/dhall/dhall.cabal
+++ b/dhall/dhall.cabal
@@ -333,7 +333,7 @@ Library
     if flag(with-http)
       Build-Depends:
         http-types                  >= 0.7.0    && < 0.13,
-        http-client                 >= 0.4.30   && < 0.6 ,
+        http-client                 >= 0.4.30   && < 0.7 ,
         http-client-tls             >= 0.2.0    && < 0.4
     if !impl(ghc >= 8.0) && !impl(eta >= 0.8.4)
       Build-Depends: semigroups == 0.18.*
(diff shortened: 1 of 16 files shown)


Gabriel439:
@ocharles: It is being built with `-Wall -Werror`.  I was actually surprised that this sort of non-exhaustive pattern match wasn't a warning by default.  I will test to see if `-fwarn-incomplete-uni-patterns` would have caught this

joneshf:
I almost opened an issue to report this bug, but stumbled onto it by accident.

In case it helps future searches of this bug, the error is: `src/Dhall/Lint.hs:67:9-51: Irrefutable pattern failed for pattern Let (l' :| ls') d'`. Hopefully that will turn up in a GitHub search/suggestion for Related Issues and prevent duplicate reports.

PR #781, opened 2019-01-16
Fix Nix build to reinstate `-Werror`

Gabriel439:
A refactor of the Nix build accidentally removed the `-Werror` flag, which
caused some warnings to get past CI.  This change fixes that and removes the
warnings.


ocharles:
You may still want to look at adding `-fwarn-incomplete-uni-patterns` as that isn't in `-Wall`, but I really wish it was!

Gabriel439:
@ocharles: Yes, my plan was to merge this first and then add `-fwarn-incomplete-uni-patterns` in #780 since the changes in the other branch are necessary to avoid that warning failing the build.

Issue #627, opened 2018-10-09
Include windows precompiled binaries in releases


jneira:
For the tortured souls who use it :stuck_out_tongue_closed_eyes:
I could compile it in my windows 10 laptop

Gabriel439:
@jneira: Yeah, feel free to do so.  You're a contributor so you have permission to attach binaries that you build to each release if you want.

In the long run what I would really like to do is to have CI build windows binaries using cross-compilation.  I'm just waiting for this work to get upstreamed so that we can have Dhall's CI build Windows binaries for every branch and every release:

https://github.com/input-output-hk/nix-hs-hello-windows

jneira:
> @jneira: Yeah, feel free to do so. You're a contributor so you have permission to attach binaries that you build to each release if you want.

Hi, i have no experience making binary distributions or github releases so to confirm the steps:
* create a binary distribution: i guess that i only have to zip the win executable in a file named `dhall-1.18.0-x86_64-windows`
* attach the new binary to an existing release: it seems i would need a tool to use the github api (like https://github.com/aktau/github-release or https://github.com/tfausak/github-release) and push permission to the repo (but i don't have it although i am a contributor, right?)

basile-henry:
`jneira` I don't think you need `github-release`, since you are a contributor here you should be able to go to https://github.com/dhall-lang/dhall-haskell/releases and click `Edit` next to `dhall-1.18.0`, you can then drop your binary to be distributed as part of the release. :smile:

basile-henry:
@jneira Ah alright, I thought you would have the permissions already. I have uploaded the windows binary for you from https://github.com/eta-lang/dhall-haskell/releases/tag/1.18.0

jneira:
@basile-henry thanks a lot! i think we can close this then (unless we want to track the future CI build with win binaries).

jneira:
Uploaded binaries for `dhall-1.19.1` and friends! `dhall-1.19.0` is coming

Gabriel439:
@jneira: Thank you! 🙂 

Just as an update, I'm waiting for this to be upstreamed into Nixpkgs so that we can build Windows binaries automatically in CI using a cross compiler:

https://github.com/input-output-hk/nix-hs-hello-windows

f-f:
I was thinking if it would be fine to add Travis or AppVeyor configs to build macOS and Windows binaries while we wait for the cross-compiler?

Gabriel439:
[...]

PR #785, opened 2019-01-18
Use appveyor to build, test and deploy in windows

jneira:
Hi, to adress #627 i've tested a build script using appveyor ci. Overview:
* Build the project using stack for `stack.yaml` (currently `lts-13`) and `stack-lts-6.yaml`
  * With `lts-6` dhall-json cannot be built cause it uses `ApplicativeDo` so i have to skip it
* Test the project for default `stack.yaml` 
  * Not including `dhall:doctests` cause throws spurious errors due to encoding. I've not found a way to avoid them like in local (setting encoding of console to utf-8)
  * No tests for `stack-lts-6.yaml` cause stack is not able to build test dependencies for dhall (at least i've not found a way to make it work, although `cabal` is able to do it)
* The build is trigger only for master, any branch containing `appveyor` or tags (and for pull requests)
* It creates artifacts for each build (downloadable in appveyor web app)
* It deploys them on tags for the build using the default `stack.yaml`, adding them to the release if it already exists.
* Thanks to caching the build takes 7 minutes
* The appveyor github app should be added to this repo to start the builds
  * I've requested its installation from the webapp
* The github token used to deploy is the mine one, it should work in this repo too, i think


appveyor.yml
```diff
@@ -0,0 +1,70 @@
...
+    name: dhall-json
+  - path: bin\dhall-text-*.zip
+    name: dhall-text
+  - path: bin\dhall-bash-*.zip
+    name: dhall-bash
+    
+deploy:
+  - provider: GitHub
+    auth_token:
+      secure: Gs/tiz6Jtg/Zjd4zGHDhuSv4Y9a+zJ574LWwzOe2bvZNGX/M8XupkxdGmBpDlnOf
```
f-f: You might not want to leave your token exposed here, how about putting it in an envvar?

jneira: Thanks for check it!, that string is my token already encrypted by appveyor, using a utility in its web app. Putting that value in the `secure:` subfield makes appveyor decrypt it before use.

appveyor.yml
```diff
@@ -0,0 +1,70 @@
...
+      - STACK_YAML: stack-lts-6.yaml
+
+  test_script:
+    - stack test dhall:tasty
+    - stack test dhall-json
+    - stack test dhall-text
+    - stack test dhall-bash
+
+artifacts:
+  - path: bin\main\dhall-*.zip
```
Gabriel439: Does this need to be inside of a `main` directory?  It seems like it would be more consistent to put it directly underneath `bin\` like the other projects

jneira: Mmm, it is only a quick workaround to distinguish in file globs dhall itself from dhall-other-ones
But the zips generated for the deploy have the .exe at top level, without any dir. So the dir layout does not survive the build.

Gabriel439: Why not do an exact match (i.e. `"bin\main\dhall-%DEPLOY_SUFFIX%"`) instead of a wildcard match?

jneira: Oh yeah that would be the right one, let me try just in case appveyor doesn't allow env vars in that position

jneira: ci is happy, thanks for the suggestion!

Gabriel439:
Would it be possible to run the Appveyor build on all pull requests?

jneira:
It can only be activate in the web ui for the project: 

>Use Save build cache in Pull Requests checkbox on the General tab of project settings if you need to save cache during PR builds. This setting is not exposed in YAML to prevent unauthorized cache modifications.

https://www.appveyor.com/docs/build-cache/

Gabriel439:
@jneira: Could we verify that by having it build this pull request?

jneira:
@Gabriel439 yeah! we need to install the appveyor app in this repo to est it, but you can view it in this pr in my fork: https://github.com/jneira/dhall-haskell/pull/3

jneira:
Have you received the request to install the appveyor github app? The builds will be triggered only for master, any branch containing `appveyor` and tags. To change that we should add more filters to the config file [here](https://github.com/dhall-lang/dhall-haskell/pull/785/files#diff-180360612c6b8c4ed830919bbb4dd459R4)
The builds without `appveyor.yml` uses the default web ui settings which by default does nothing and report a succesful build so install it before merging wouldn't do any harm.

Gabriel439:
@jneira: I just approved the request.  However, is there any reason we we would not want to build all pull requests?

The way that the Hydra CI works is that it builds the `master` branch and all pull requests

jneira:
@Gabriel439 I have not been precise: in fact it will build master, branch with appveyor, tags *and* all pull requests. Moreover, it will read the cache for pull request, only write cacge from pull requests is optional.

jneira:
Appveyor is building this! Thanks for the setup, @Gabriel439

Gabriel439:
You're welcome!  This is good to merge now

Issue #783, opened 2019-01-17
[Question] Static executables for Nix/NixOS


PierreR:
Is there a way to install the `dhall` static executable (and all other extensions like dhall-bash) in nix/nixos using something like:
```
nix-env -iA dhall -f https://github.com/dhall-lang/archive/v1.21.1.tar.gz
```
I am currently using a [custom derivation](https://github.com/PierreR/devbox-config/blob/master/default.nix) for this but maybe this is not necessary ?

Gabriel439:
@PierreR: This should work:

```bash
$ nix-env --install --attr dhall --file https://github.com/dhall-lang/dhall-haskell/archive/1.20.1.tar.gz
```

PierreR:
@Gabriel439 does it ? It seems to download ghc, ... I really want the static executables only.

ocharles:
I think you need Dhalls Hydra as a binary cache

PierreR:
@ocharles Well I just want to get the provided static executable. So that's what I do locally:
```
with import <nixpkgs> {};

let

# Instead of building from nixpkgs, use the provided static executable.
# By relying solely on the static executable, we don't care about the nix machinery that would build from source.
# (chances are that the version we want is not yet in the binary cache).
# For instance, building with nix using `nix-env -iA haskellPackages.dhall_1_20_1 -f https://github.com/NixOS/nixpkgs/archive/088cb13aee074fb48d104222cb80fb552ec17529.tar.gz `
# would currently pull 250M (unpacked) and then compile dhall ...
dhall =
  stdenv.mkDerivation rec {
    name = "dhall-${version}";
    version = "1.20.1";
    src = fetchurl {
        url = "https://github.com/dhall-lang/dhall-haskell/releases/download/1.20.1/dhall-${version}-x86_64-linux.tar.bz2";
        sha256 = "1ycb0pglwi9k1pn4qqjyvd4bi54000fb9aqxnn550a515rm6cb03";
    };

    installPhase = ''
      mkdir -p $out/bin
      cp dhall $out/bin/
    '';
  };
in
{
inherit dhall;
}
```
But maybe the repo could offer such a derivation if it doesn't already

PierreR:
@ocharles do you mean that if I set up the dhall binary cache I will get it with 
`$ nix-env --install --attr dhall --file https://github.com/dhall-lang/dhall-haskell/archive/1.20.1.tar.gz`

Sorry I read you comment too quickly ;-)

[...]

PR #786, opened 2019-01-19
Change `default.nix` to use static linking when possible

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/783

... as suggested by @PierreR in https://github.com/dhall-lang/dhall-haskell/issues/783#issuecomment-455502506


Gabriel439:
@PierreR: You can verify this works by running:

```
$ nix build --file https://github.com/dhall-lang/dhall-haskell/tarball/gabriel/possibly_static dhall
```

Note that static linking is currently only supported for Linux, so if you use OS X you will get a result that is not statically linkedd

PR #795, opened 2019-01-24
Add --list flag to dhall resolve

aleator:
This pull request adds a flag for producing a machine readable listing of
dependencies of a dhall expression.

This flag can be used to integrate dhall more easily to file watchers or
build systems which require such lists of dependencies.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-13.2
+resolver: lts-13.3
 packages:
   - dhall
   - dhall-bash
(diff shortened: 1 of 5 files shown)


stack.yaml
```diff
@@ -1,4 +1,4 @@
-resolver: lts-13.2
+resolver: lts-13.3
```
aleator: Oops. Didn't mean to change this..

Gabriel439: That's fine.  There's nothing wrong with supporting a newer Stack version

PR #800, opened 2019-01-29
Add support for Unicode path components

Gabriel439:
... as standardized in https://github.com/dhall-lang/dhall-lang/pull/353


fosskers:
It works!
```
let entries : List (List Entry) = [ ./families/"禺.dhall", ./families/"扁.dhall" ]
```

PR #803, opened 2019-01-30
Add --immediate flag to resolve mode

aleator:
This patch adds means for listing only the immediate dependencies of
given dhall expression.

This is useful for integrating dhall to build systems where dhall
expressions (e.g. of type `Text`) are build products.

I would expect there to be other, more sensible, uses for this also.

### DIFF
--- a/dhall/src/Dhall/Main.hs
+++ b/dhall/src/Dhall/Main.hs
@@ -97,10 +97,11 @@ data Mode
     | Encode { json :: Bool }
     | Decode { json :: Bool }
 
-data ResolveMode 
+data ResolveMode
     = Dot
     | ListDependencies
-    
+    | ListImmediateDependencies
+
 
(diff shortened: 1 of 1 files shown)


dhall/src/Dhall/Main.hs
```diff
@@ -198,6 +199,12 @@ parseMode =
               <>  Options.Applicative.help
                     "Output import dependency graph in dot format"
               )
+        <|>
+          Options.Applicative.flag' (Just ListImmediateDependencies)
+              (   Options.Applicative.long "immediate"
```
Gabriel439: How about calling this flag `--immediate-dependencies` and then renaming `--list` to `--transitive-dependencies`?

aleator: ... I was wondering what to call these.. :)

PR #805, opened 2019-01-31
Remove version tag from encoded expressions

Gabriel439:
... as standardized in https://github.com/dhall-lang/dhall-lang/pull/362

### DIFF
--- a/dhall/src/Dhall/Binary.hs
+++ b/dhall/src/Dhall/Binary.hs
@@ -100,6 +100,7 @@ unApply e = (baseFunction₀, diffArguments₀ [])
         ~(baseFunction, diffArguments) = go f
     go baseFunction = (baseFunction, id)
 
+-- | Encode a Dhall expression to a CBOR `Term`
 encode :: Expr s Import -> Term
 encode (Var (V "_" n)) =
     TInteger n
@@ -817,6 +818,7 @@ decodeMaybe (TList [ TInt 26, t₁, _T₁ ]) = do
 decodeMaybe _ =
     empty
 
+-- | Decode a Dhall expression from a CBOR `Term`
 decode :: Term -> Either DecodingFailure (Expr s Import)
 decode term =
     case decodeWithoutVersion <|> decodeWithVersion of
--- a/dhall/src/Dhall/Binary.hs
+++ b/dhall/src/Dhall/Binary.hs
@@ -22,7 +22,7 @@ module Dhall.Binary
     ) where
 
 import Codec.CBOR.Term (Term(..))
-import Control.Applicative (empty)
+import Control.Applicative (empty, (<|>))
 import Control.Exception (Exception)
 import Dhall.Core
     ( Binding(..)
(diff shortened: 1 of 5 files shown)


dhall/src/Dhall/Binary.hs
```diff
@@ -810,47 +810,19 @@ decode (TList (TInt 25 : xs)) = do
...
     return (Annot t₀ _T₀)
-decode _ =
+decodeMaybe _ =
     empty
 
-{-| Decode a Dhall expression
+decode :: Term -> Either DecodingFailure (Expr s Import)
+decode term = case decodeMaybe term of
+    Nothing         -> Left (CBORIsNotDhall term)
+    Just expression -> Right expression
```
f-f: I wonder if we could we provide backwards-compatibility in some way here?

E.g. something like
```haskell
decode :: Term -> Either DecodingFailure (Expr s Import)
decode term = case decodeMaybe term of
    Just expression -> Right expression
    Nothing -> case term of
        -- Pre-6.0.0 expressions
        TList [ TString version, subTerm ] -> case decodeMaybe subTerm of
          Just expression -> Right expression
          Nothing -> Left (CBORIsNotDhall subTerm)
        _ -> Left (CBORIsNotDhall term)
```

The reason is that otherwise every user will have to reimplement some kind of compatibility layer (and this means "forking" `Dhall.Import.exprFromImport` to have it call this code basically)

f-f:
@Gabriel439 I felt there was a more beautiful way to write that :) 

Now the CBOR decoding from cache goes fine, but in case you're importing a hash-protected expression, there's a `HashMismatch` happening, I think [thrown here](https://github.com/dhall-lang/dhall-haskell/blob/gabriel/no_version_tag/dhall/src/Dhall/Import.hs#L830) because the hash of the expression with the current standard doesn't match the hash in the file (as the CBOR term given by the new `encode` is different).  

The last bit of backwards-compat could be another alternative [here](https://github.com/dhall-lang/dhall-haskell/blob/gabriel/no_version_tag/dhall/src/Dhall/Import.hs#L830) where instead of only one `actualHash`, we have an alternative in which we try to `hashExpression` (that calls the new `Dhall.Binary.encode`), and to `hashExpressionLegacy` (that would call a `Dhall.Binary.encodeLegacy`, which would wrap the term with the version in the same way we do currently on master) for every past standard version (or the last one, or the last two, etc), so that there would be at least one hash that matches.

dhall/src/Dhall/Import.hs
```diff
@@ -527,15 +527,21 @@ exprToImport here expression = do
...
+
+                liftIO (Data.ByteString.writeFile cacheFile bytes)
+
+        let fallback = do
+                let actualHash = hashExpression NoVersion normalizedExpression
 
-        liftIO (Data.ByteString.writeFile cacheFile bytes)
+                liftIO (Control.Exception.throwIO (HashMismatch {..}))
+
+        Data.Foldable.asum (map check [ NoVersion , V_5_0_0 ]) <|> fallback
```
f-f:
```suggestion
        Data.Foldable.asum (map check [ NoVersion , V_5_0_0, V_4_0_0 ]) <|> fallback
```

dhall/src/Dhall/Import.hs
```diff
@@ -827,12 +833,19 @@ loadWith expr₀ = case expr₀ of
...
-            let actualHash =
-                    hashExpression _standardVersion (Dhall.Core.alphaNormalize expr)
+            let matches version =
+                    let actualHash =
+                            hashExpression version (Dhall.Core.alphaNormalize expr)
+
+                    in  expectedHash == actualHash
 
-            if expectedHash == actualHash
+            if any matches [ NoVersion, V_5_0_0 ]
```
f-f:
```suggestion
            if any matches [ NoVersion, V_5_0_0, V_4_0_0 ]
```

Spago is on `1.19.1` :)

jneira: And dhall-to-etlas!

Gabriel439: Yeah, I meant to include all versions back to 1.0.0, but I just forgot to go back and add them all here 🙂

f-f:
@Gabriel439 I tested this branch locally with `spago` and it works very nicely. Very many thanks! 😊

PR #806, opened 2019-01-31
Add :hash command to the repl

basile-henry:
This PR makes it possible to get the dhall formatted hash of any dhall expression from the repl.


Gabriel439:
Great idea!  Thank you 🙂

PR #807, opened 2019-01-31
Save and load REPL context via :save and :load commands

### LINKED ISSUE #706: `:save` with no arguments saves REPL state
Currently the `:save` command in the `dhall repl` lets you save individual expressions to files like this:

```haskell
⊢ :save x = 1  -- Creates a file named `x` with the contents `1`
```

it would be neat if the `:save` command with no arguments saved your current REPL state so that if you closed the REPL then you could pick up where you left off by running `:load`.  In other words:

```haskell
$ dhall repl
⊢ :let x = 1

x : Natural

⊢ :let y = 2

y : Natural

⊢ :save
⊢ <Ctrl-D>
Goodbye

$ dhall repl
⊢ :load
⊢ x + y

3

```

basile-henry:
This PR closes #706

The format for the save file is a plain Dhall file where we expect the normal form to be a record literal. This will hopefully mean that the save file can be edited manually if need be without too much effort.

The save files are always kept by default (their index is incrementing) and only the last one is loaded by default.

It is possible to manually specify where to save and load the context to.

### DIFF
--- a/dhall/src/Dhall/Repl.hs
+++ b/dhall/src/Dhall/Repl.hs
@@ -26,7 +26,7 @@ import Dhall.Pretty (CharacterSet(..))
 import Lens.Family (set)
 import System.Console.Haskeline (Interrupt(..))
 import System.Console.Haskeline.Completion ( Completion, simpleCompletion )
-import System.Directory (listDirectory)
+import System.Directory ( getDirectoryContents )
 import System.Environment ( getEnvironment )
 
 import qualified Control.Monad.Trans.State.Strict as State
@@ -270,7 +270,7 @@ saveFilePrefix = ".dhall-repl"
 -- | Find the index for the current _active_ dhall save file
 currentSaveFileIndex :: MonadIO m => m (Maybe Int)
 currentSaveFileIndex = do
-  files <- liftIO $ listDirectory "."
+  files <- liftIO $ getDirectoryContents "."
 
   let parseIndex file
         | saveFilePrefix `isPrefixOf` file
--- a/dhall/src/Dhall/Repl.hs
+++ b/dhall/src/Dhall/Repl.hs
@@ -11,11 +11,13 @@ module Dhall.Repl
     ) where
 
 import Control.Exception ( SomeException(SomeException), displayException, throwIO )
+import Control.Monad ( forM )
 import Control.Monad.IO.Class ( MonadIO, liftIO )
 import Control.Monad.State.Class ( MonadState, get, modify )
 import Control.Monad.State.Strict ( evalStateT )
 import Data.List ( isPrefixOf, nub )
 import Data.List.NonEmpty (NonEmpty(..))
+import Data.Maybe ( mapMaybe )
 import Data.Semigroup ((<>))
 import Dhall.Binary (StandardVersion(..))
 import Dhall.Context (Context)
@@ -24,10 +26,12 @@ import Dhall.Pretty (CharacterSet(..))
 import Lens.Family (set)
 import System.Console.Haskeline (Interrupt(..))
 import System.Console.Haskeline.Completion ( Completion, simpleCompletion )
+import System.Directory (listDirectory)
 import System.Environment ( getEnvironment )
 
 import qualified Control.Monad.Trans.State.Strict as State
 import qualified Data.HashSet
+import qualified Data.Map
 import qualified Data.Text as Text
 import qualified Data.Text.IO as Text
 import qualified Data.Text.Prettyprint.Doc as Pretty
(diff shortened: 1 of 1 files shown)


Gabriel439:
@basile-henry: It's because CI builds `dhall` against GHC 7.10.3 which requires a specific (old) version of the `directory` library

dhall/src/Dhall/Repl.hs
```diff
@@ -260,7 +264,108 @@ hashBinding tokens = do
...
+loadBinding _ = fail ":load should be of the form `:load` or `:load file`"
+
 saveBinding :: ( MonadIO m, MonadState Env m ) => [String] -> m ()
+-- Save all the bindings into a big record
+saveBinding [] = do
+  file <- nextSaveFile
+
+  saveBinding [file]
+
+-- Save all the bindings into a big record into `file`
```
Gabriel439: We probably should use something other than a record to store the bindings because we need to preserve the original order.  The context permits duplicate entries and if their order is switched then you change which one shadows which

basile-henry: I saw that the context keeps the shadowed bindings, but I'm not too sure why.

Maybe if we weren't normalizing the bindings before adding them to the context then we would need to have access to the shadowed bindings. But I believe that when we normalise while adding a new binding, it gets rid of any dependencies on previous bindings, or is that not the case?

Or maybe we need access to the shadowed bindings for something else?

The current solutions I have is to only serialise the only accessible version of a binding accessible to the user when they type `:save`.

Gabriel439: @basile-henry: It's still possible to refer to shadowed bindings using the `@n` suffix, like this:

```
⊢ :let x = 1

x : Natural

⊢ :let x = True

x : Bool

⊢ x@1

1
```

jneira:
Strangely the appveyor build with ghc-7.10.3 succeded 🤔

dhall/src/Dhall/Repl.hs
basile-henry: Ah, I didn't know about this feature.

I liked the idea of using dhall to serialise the REPL state. What would you recommend as a format?

I could use `[ { name : Text, expr : Text } ]` but then the save file can't really be used for anything else.

dhall/src/Dhall/Repl.hs
```diff
@@ -255,26 +254,138 @@ hashBinding tokens = do
...
+        ":load expects `" <> file <> "` to contain one command per line"
+
+      -- Run all the commands
+      handler nullHandle =
+          State.execStateT
+            (forM_ replLines runCommand)
+            (env { outputHandle = nullHandle })
+
+  -- Use a null handle (to discard output) to run all the commands
+  newEnv <- liftIO $ System.IO.withFile "/dev/null" System.IO.WriteMode handler
```
basile-henry: I have no idea how platform independent this solution is. Any pointers to improve this? Or do you think this works on Windows somehow? :crossed_fingers:

jneira: Appveyor is happy so it works in windows 👍

basile-henry: Unfortunately I don't think Appveyor runs the REPL, and this will likely be a runtime error 😅

jneira: oh right, i sometimes forgot that sometimes it compiles but fails at runtime. It seems that it will fail in windows, yes.

jneira: I've tested the pr in my windows env and i've not been able to trigger any error. The steps were:
* run the dhall repl
* type some binding: `:let a = 1`
* `:save` the session
* `:load` the session
* quit and rerun repl and load the session again

basile-henry: Thanks for testing, I'm still surprised it doesn't crash. Could you try to exit in between `save` and `load`? And the check if `a` is in scope.

basile-henry: Because of the `withFile` pattern, if there's an error in the handle. It probably just skips the loading bit

jneira: > Thanks for testing, I'm still surprised it doesn't crash. Could you try to exit in between `save` and `load`? And the check if `a` is in scope.

My last test was just that, including the console output to be more precise:
```console
D:\dhall-haskell>stack run dhall -- repl

⊢ :let a = 1

a : Natural

⊢ a

1

⊢ :save

Context saved to `.dhall-repl-0`

⊢ :quit
Goodbye.

D:\dhall-haskell>stack run dhall -- repl

⊢ :load

Loaded `.dhall-repl-0`

⊢ a

1

Comment (jneira) [reply dhall/src/Dhall/Repl.hs]: I am on windows 10, not sure if it would work in older systems.
Maybe the solution of this so answer could be applicable here: https://stackoverflow.com/a/41527205/49554

Comment (basile-henry) [reply dhall/src/Dhall/Repl.hs]: Ah thank you! If `hSilence` is properly cross platform I could use that. Otherwise I would use a `Maybe Handle` as part of the state instead and just not print anything when I don't have a handle.
That might be cleaner! 😄

Comment (Gabriel439) [reply dhall/src/Dhall/Repl.hs]: Yeah, I would configure the REPL so it can optionally not print output when replaying commands

File: dhall/src/Dhall/Repl.hs [hunk 3]
```diff
@@ -75,6 +79,7 @@ data Env = Env
   , explain          :: Bool
   , characterSet     :: CharacterSet
   , _standardVersion :: StandardVersion
+  , outputHandle     :: Maybe System.IO.Handle
```
Gabriel439: Perhaps instead of storing a `Maybe System.IO.Handle` you can store a `Text -> IO ()`.  Then when you want to write something you set it to `Data.Text.IO.putStrLn` and when you want to output nothing you set it to `mempty`

basile-henry: We also use `renderIO` to pretty print Dhall, which uses the handle.
I could add an extra rendering function similar to what you suggest, as part of the state, but the `Maybe Handle` seems lighter. What do you prefer?

Gabriel439:
I'll go ahead and merge this.  If we need any additional changes I can make them in a subsequent pull request

PR #808, opened 2019-01-31
Change `dhall freeze` to only freeze `Remote` imports

Gabriel439:
The motivation behind this change is so that users can freeze remote imports
(like the Prelude) but ignore local imports so that subsequent runs of the
interpreter reflect changes to local files and environment variables.

The reasoning behind this is that there are two primary benefits of integrity
checks:

* Improved security
* Caching

... and one downside which is that updates to those imports are not pulled in
until the integrity check is updated or removed.

However, environment variables and local file paths do not benefit from
improved security or caching, so there is only a downside to freezing them.

Specifically:

* Environment variables and local file paths are both cheap to resolve

  ... so they don't really benefit from caching.  d

  To be precise, they *could* benefit from caching if the cache expression is
  cheaper to parse and normalize compared to the original file, but with recent
  improvements to the parser this difference should be minor.  Also, even if
  the performance difference were large I believe the right solution would be
  to improve the performance of the interpreter rather than rely on integrity
  checks as an accelerant to type-checking/normalization.

* Environment variables and local file paths are trusted

  For example, when a user runs the `dhall` executable they are implicitly
  trusting their filesystem which provides that executable.  Similarly, when
  they run `dhall` without an absolute path they are implicitly trusting that
  their `PATH` environment variable has not been compromised to point to a
  malicious executable.

  Up until now, Dhall's threat model has always been that local imports are
  trusted but remote imports are not, so this is consistent with that threat
  model.

... so as far as environment variables and local files are concerned there are
only downsides to freezing them and no up-side.  This is why this change
no longer freezes them.

This also renames `hashImport` to `freezeImport` for more terminology
consistency.

### DIFF
--- a/dhall/src/Dhall/Freeze.hs
+++ b/dhall/src/Dhall/Freeze.hs
@@ -124,9 +124,11 @@ freeze
     :: Maybe FilePath
     -- ^ Modify file in-place if present, otherwise read from @stdin@ and write
     --   to @stdout@
+    -> Bool
+    -- ^ If `True` then freeze all imports, otherwise freeze only remote imports
     -> StandardVersion
     -> IO ()
-freeze inplace _standardVersion = do
+freeze inplace everything _standardVersion = do
     (text, directory) <- case inplace of
         Nothing -> do
             text <- Data.Text.IO.getContents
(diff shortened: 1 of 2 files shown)


jneira:
@Gabriel439 i agree, but although parsing and normalizing performance is way better, reading dhall-to-cabal files can take some seconds. I usually was using freeze over a file with only the import of another one to put the later in cache and save those seconds in subsequent reads. However, for me is not a big deal do it manually.

It would be great to make dhall use the caching version of a file automatically, without importing them in another one. It is somewhat odd knowing that a dhall file is already in cache and dhall continue takins seconds to read it. Maybe using another dhall file with a mapping between the canonical path of the file and the entries of the cache?

jneira:
Otoh although you can trust your file system and env vars are not compromised externally, the can be modified by mistake (imagine envs with lot of "trusted" users -humans or programs- making changes) and maybe you can ensure the content of the env var or file had not been changed.

Gabriel439:
@joneshf: I added a commit for an `--everything` flag that will freeze all imports

joneshf:
Awesome! That works for me.

PR #812, opened 2019-02-07
Add source position information for missing imports

Gabriel439:
Related to #561

This adds source position information to missing imports

Before:

```
$ dhall <<< './foo'

↳ ./foo

Error: Missing file …/foo
```

After:

```
$ dhall <<< './foo'

↳ ./foo

Error: Missing file …/foo

(stdin):1:1
```

### DIFF
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -167,7 +167,7 @@ import Dhall.Import.HTTP
 import Dhall.Import.Types
 import Text.Dot ((.->.), userNodeId)
 
-import Dhall.Parser (Parser(..), ParseError(..), Src(..))
+import Dhall.Parser (Parser(..), ParseError(..), Src(..), SourcedException(..))
 import Dhall.TypeCheck (X(..))
 import Lens.Family.State.Strict (zoom)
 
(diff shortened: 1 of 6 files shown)


Profpatsch:
Awesome, I think I can work with this. Now my next request would be to show a `NonEmpty` of all missing imports. I’d like to see you do that with exceptions. :P

Gabriel439:
@Profpatsch: Yeah, that will follow in another pull request (possibly replacing exceptions with something like `ExceptT` along the way)

Profpatsch:
> possibly replacing exceptions with something like `ExceptT` along the way

I approve of this. It’s so darn hard to follow the exception codepaths.

Profpatsch:
Ah, the position of the imports in the source files are not yet available though, I get `1:1:(source)` when running this patch:

`foo.dhall`
```
{ aetiglnetglineeglseglvcnvfglcnlvfgneqlenlvqgenlvqgenleqgvnle =
	{ xyc = ./abc }
}
```

```
$ dhall-flycheck foo.dhall
[{"filename":"foo.dhall","line":1,"column":1,"message":"\n\n↳ ./abc\n\nError: Missing file /home/philip/kot/dhall/flycheck/abc\n\n"}]
```

I would expect the error position to point to the file name on line 2 of `foo.dhall`. That way I can show the problematic file.

Oh, and if I have

`bar.dhall`
```
./foo.dhall
```

```
$ dhall-flycheck bar.dhall
[{"filename":"bar.dhall","line":1,"column":1,"message":"\n\n↳ ./foo.dhall\n  ↳ ./abc\n\nError: Missing file /home/philip/kot/dhall/flycheck/abc\n\n"}]
```

I need some way to point to the import of `./foo.dhall` in `bar.dhall` to highlight the transitive import problem (so that the user can follow the import chain though until the import of the missing file itself is highlighted).

Profpatsch:
Something like `{ foo = ./foo.dhall, abc = ./abc }` should highlight both `./foo.dhall` (because of the transitive import error of `abc`) as well as `./abc` (because of the direct import error) of course.

PR #817, opened 2019-02-13
Remove explicit recursion from Dhall lint in favour of micropasses

ocharles:
This change drastically reduces the amount of code required to write the linter. Instead of explicitly recursing over the expression tree, we can instead break linting apart into small micropasses and compose these as a bottom-up traversal over the `subExpressions` lens. This is done by [`rewriteOf`](http://hackage.haskell.org/package/lens-4.17/docs/Control-Lens-Plated.html#v:rewriteOf) in the `lens` framework - but rather than depending on all of that I've just included the relevant function (using `lens-family-core`). Would appreciate licensing advice here, @phadej / @ekmett!

### DIFF
--- a/dhall/src/Dhall/Lint.hs
+++ b/dhall/src/Dhall/Lint.hs
@@ -3,12 +3,17 @@
 module Dhall.Lint
     ( -- * Lint
       lint
+    , removeLetInLet
+    , removeUnusedBindings
+    , optionalLitToSomeNone
+    , dropConstructorsKeyword
     ) where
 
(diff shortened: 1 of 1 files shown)


Gabriel439:
Very nice! 🙂

ekmett:
You're hereby more than welcome to use the code of `rewriteOf` under the permissive license of your choice. Whatever is suitable to this project is fine.

joneshf:
This has gotta be one of coolest changes I've seen in a while!

PR #821, opened 2019-02-15
Add `linux-dhall*` jobs to Hydra

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/783

These can be used to conveniently install Dhall using `nix-env` on Linux systems

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,20 +26,32 @@ For OS X, use `brew` to install the desired package.  For example:
 $ brew install dhall-json
 ```
 
-You can also install pre-built Linux binaries for `master` using Nix by visiting
-one of the following links:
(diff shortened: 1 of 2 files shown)


ocharles:
Btw, Hydra has an API that exposes the store path for jobs. We do deployments at work using this. To install the latest Dhall, for example, you would do:

```sh
nix-env -i $(https://hydra.dhall-lang.org/job/dhall-haskell/master/linux-dhall/latest/api/get-info | jq -r .outPath)
```

ocharles:
Another option is to suggest people use the channel Hydra gives you:

https://hydra.dhall-lang.org/jobset/dhall-haskell/449/channel/latest

This would also let people easily update.

PR #822, opened 2019-02-16
CBOR-encode only some special values as half-floats

f-f:
As standardized in https://github.com/dhall-lang/dhall-lang/pull/376

### DIFF
--- a/dhall/src/Dhall/Binary.hs
+++ b/dhall/src/Dhall/Binary.hs
@@ -49,7 +49,6 @@ import Data.Text (Text)
 import Options.Applicative (Parser)
 import Prelude hiding (exponent)
 import GHC.Float (double2Float, float2Double)
-import Codec.CBOR.Magic (floatToWord16, wordToFloat16)
 
 import qualified Crypto.Hash
 import qualified Data.ByteArray.Encoding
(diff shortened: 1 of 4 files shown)


dhall/src/Dhall/Binary.hs
```diff
@@ -377,9 +376,10 @@ encode (DoubleLit n64)
     | otherwise = TDouble n64
   where
     n32      = double2Float n64
-    n16      = floatToWord16 n32
     useFloat = n64 == float2Double n32
-    useHalf  = n64 == (float2Double . wordToFloat16 . fromIntegral) n16
+    -- the other three cases for Half-floats are 0.0 and the infinities
+    useHalf  = any id $ fmap (n64 ==) [0.0, infinity, -infinity]
```
basile-henry:
```suggestion
    useHalf  = or $ fmap (n64 ==) [0.0, infinity, -infinity]
```

dhall/src/Dhall/Binary.hs
```diff
@@ -377,9 +376,10 @@ encode (DoubleLit n64)
...
+    infinity = (read "Infinity") :: Double
```
basile-henry: I would personally prefer defining `infinity` via the `Num` interface. But your version via `Read` is fairly understandable too. :smile: 

```suggestion
    infinity = 1/0 :: Double
```

f-f:
Note: I'll also setup the "binary" tests in this PR once the change to the standard is merged

PR #824, opened 2019-02-22
Show some details in InvalidType error output 

jneira:
* Maybe it has not the full context but i think it can help to high level Dhall api users

### DIFF
--- a/dhall/src/Dhall.hs
+++ b/dhall/src/Dhall.hs
@@ -105,6 +105,7 @@ import Data.Monoid ((<>))
 import Data.Scientific (Scientific)
 import Data.Sequence (Seq)
 import Data.Text (Text)
+import Data.Text.Prettyprint.Doc (Pretty)
 import Data.Typeable (Typeable)
 import Data.Vector (Vector)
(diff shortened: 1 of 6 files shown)


dhall/src/Dhall.hs
```diff
@@ -156,19 +158,34 @@ throws (Right r) = return r
...
+instance (Pretty s, Pretty a, Typeable s, Typeable a) => Show (InvalidType s a) where
+    show InvalidType { .. } =
         _ERROR <> ": Invalid Dhall.Type                                                  \n\
         \                                                                                \n\
         \Every Type must provide an extract function that succeeds if an expression      \n\
-        \matches the expected type.  You provided a Type that disobeys this contract     \n"
+        \matches the expected type.  You provided a Type that disobeys this contract     \n\
+        \                                                                                \n\
+        \The Type provided has the expected dhall type:                                  \n\
+        \" <> show txt0 <> "\n\
```
Gabriel439: Could you add one blank line preceding each insert?  This is mainly for consistent output style

jneira: oh, you are right

PR #826, opened 2019-02-22
Correct Dhall installation one-liner in README

ocharles:
The README missed the call to `curl -L`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ If you have the `jq` command-line tool installed then you can do this in one
 command by running:
 
 ```haskell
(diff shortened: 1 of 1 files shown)


ocharles:
You also need to have `hydra.dhall-lang.org` as a binary substitute, I'm not sure how we want to document that.

Gabriel439:
@ocharles: I can take care of documenting how to set up the cache in a separate pull request

Gabriel439:
Thank you for fixing that! :slightly_smiling_face:

PR #835, opened 2019-02-28
Sort record and union fields when CBOR-encoding

f-f:
As standardized in https://github.com/dhall-lang/dhall-lang/pull/392


Gabriel439:
@f-f: It looks like there is a binary serialization roundtrip failure because serialization now sorts the fields

f-f:
@Gabriel439 it's so lovely that QuickCheck can catch these things :)

The only way I could think of to fix this was to change the `Arbitrary` instance for `Dhall.Map` so that the fields are sorted.
With this fix the serialization roundtrips once again for the seed that failed on CI.
Do you have any other idea about it or is this ok?

Gabriel439:
@f-f: Yeah, I think that's a perfectly fine solution

PR #836, opened 2019-02-28
Use binary encoding in parser tests

f-f:
With this PR the tests suite will start to compare its encoding with the "expectation" side of the "Parser" tests from the `dhall-lang` repo, using the binary-encoded tests introduced in https://github.com/dhall-lang/dhall-lang/pull/393


f-f:
CI is not passing because it will need the fixes in #835, but the test results here demonstrate well the issue :)

Gabriel439:
Thank you for doing this! 🙂

f-f:
You're welcome 🙂

Issue #853, opened 2019-03-15
dhall cli: alpha normalization


singpolyma:
Would be great if there were a `dhall normalize --including-alpha` or similar that would completely normalize an expression -- trying to debug semantic hashes, but the hash is of encoded alpha-normalized expression and the only output I can get it encoded non-alpha-normalized.

Gabriel439:
I agree!  I've also been meaning to add this for another reason: we could enable Unicode identifiers and optionally α-normalize expressions when auditing them to bypass Unicode obfuscation attempts

Gabriel439:
Pull request up here: https://github.com/dhall-lang/dhall-haskell/pull/855

PR #855, opened 2019-03-15
Add `--alpha` flag to α-normalize output

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/853

... as suggested by @singpolyma

### DIFF
--- a/dhall/benchmark/dhall-command/Main.hs
+++ b/dhall/benchmark/dhall-command/Main.hs
@@ -8,7 +8,7 @@ import           Dhall.Binary (defaultStandardVersion)
 
 options :: Main.Options
 options = Main.Options
-  { Main.mode = Main.Default False
+  { Main.mode = Main.Default False False
   , Main.explain = False
   , Main.plain = False
   , Main.ascii = False
--- a/dhall/src/Dhall/Main.hs
+++ b/dhall/src/Dhall/Main.hs
@@ -83,11 +83,11 @@ data Options = Options
 
 -- | The subcommands for the @dhall@ executable
 data Mode
-    = Default { annotate :: Bool }
+    = Default { annotate :: Bool, alpha :: Bool }
     | Version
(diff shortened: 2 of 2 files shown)


singpolyma:
Hmm, so maybe I'm not understanding but I'm seeing this:

```
$ dhall normalize --alpha < dhall-lang/tests/normalization/success/simple/optionalBuildA.dhall | dhall encode | sha256sum
f0852efdbd6263099a8feb981c1b245b0109bf930bece3099c69ef1f71ade582  -

$ dhall normalize --alpha < dhall-lang/tests/normalization/success/simple/optionalBuildA.dhall | dhall hash
sha256:57937e066a166baf3d712e744d9aec91894f14d7b25138a6f60c7b7577cc986b
```

Am I missing something?  Should not `normalize --alpha | encode` be exactly the bytes to hash to make the semantic integrity check?

singpolyma:
Ok, so, using some `Debug.Trace` I've found the discrepancy.  The encoding call used by `dhall hash` produces a binary encoding that contains this:

```
(_ (Optional Bool) (Some true))
```

That is, the binary encoding is a function `_` applied to *two* arguments, vs the `dhall encode` output where the same expression is set up like this:

```
((_ (Optional Bool)) (Some true))
```

The former seems like a bug to me, since nothing in the spec indicates a normalization to a function application with multiple arguments.

Gabriel439:
@singpolyma: The binary encoding does specify that a function applied to multiple arguments should be uncurried and encoded as a single array:

https://github.com/dhall-lang/dhall-lang/blob/master/standard/binary.md#function-application

However, you were right that the two different commands should have matched and I have another pull request up with the fix:

https://github.com/dhall-lang/dhall-haskell/pull/859

The short answer is that the output from `dhall hash` was the correct one

Gabriel439:
I'll also go ahead and merge this since the `--alpha` flag is working correctly

PR #857, opened 2019-03-16
Fix for UnionType

mstksg:
I noticed that I used the wrong type of equality for union subtypes for `union`.  This causes the `Type` to falsely reject situations where the "other branches" of a union lit are what are expected, but are not in a normalized form.  Using `judgmentallyEqual` gives the proper equality and gets rid of false negatives when testing for equality.


Gabriel439:
Good catch! 🙂

Issue #878, opened 2019-04-02
json-to-dhall


antislava:
Implemented a simple `json-to-dhall` tool: https://github.com/dhall-lang/dhall-haskell/compare/master...antislava:antislava/json-to-dhall.
Haddock documentation following the structure of `dhall-to-json` is a bit rough but reasonably complete in [Main.hs](https://github.com/antislava/dhall-haskell/blob/antislava/json-to-dhall/dhall-json/json-to-dhall/Main.hs) (`cabal new-haddock exe:json-to-dhall` works in `dhall-json`'s `nix-shell`).

The conversion function is relatively straightforward but I had to put some thought into what the default behavior should be (in particular with regards to unions and records).

So far tested it only on simple cases (listed in the documentation). For a more "real-life" example, check out [bower.dhall](https://github.com/antislava/dhall-haskell/blob/antislava/json-to-dhall/dhall-json/examples/bower.dhall).

Related discussions:
* [Discussion: transpile dhall to other languages in both directions? · Issue #346 · dhall-lang/dhall-lang](https://github.com/dhall-lang/dhall-lang/issues/346)
* [Discussion: Support for importing from JSON · Issue #121 · dhall-lang/dhall-lang](https://github.com/dhall-lang/dhall-lang/issues/121)
* [Interest in json-to-dhall and yaml-to-dhall? · Issue #23 · dhall-lang/dhall-json](https://github.com/dhall-lang/dhall-json/issues/23)

Profpatsch:
Does `dhall-to-json(json-to-dhall(x)) == x` hold?

antislava:
@Profpatsch `json-to-dhall(t, dhall-to-json(x : t)) == x`, where `t` is the Dhall type (schema) which is the first argument to `json-to-dhall`, the second being the actual JSON data, should hold in most cases. Proper proof/tests will be the next step, but it does already hold for the provided `bower.dhall/bower.json` example (up to the key re-ordering, obviously) and in basic cases in the  [Main.hs](https://github.com/antislava/dhall-haskell/blob/antislava/json-to-dhall/dhall-json/json-to-dhall/Main.hs) documentation section:

```sh
json-to-dhall Bool <<< 'true' | dhall-to-json
json-to-dhall Bool <<< 'false' | dhall-to-json
json-to-dhall Integer <<< 2 | dhall-to-json
json-to-dhall Natural <<< 2 | dhall-to-json
[...]

PR #884, opened 2019-04-09
json-to-dhall initial implementation

antislava:
Description: `json-to-dhall(s, j) -> d`, where `j` is a valid JSON being converted, `s` is a Dhall type (schema), and `d` is the resulting Dhall value (term). 

Background: [json-to-dhall · Issue #878 · dhall-lang/dhall-haskell](https://github.com/dhall-lang/dhall-haskell/issues/878)

Note: All the conversion logic is currently in `Main.hs` and will  have to be factored out to `dhall-json` library in the future, to make it usable in other Haskell projects. This should be a separate next step as it will require some coordination with the `dhall-to-json` logic of the library, following initial practical tests and find-tuning of the `json-to-dhall` tool, and most likely after some efforts to standardize/formalize JSON <=> Dhall conversion rules (which may also result in the upstreaming of the core logic to dhall-lang).


dhall-json/json-to-dhall/Main.hs
```diff
@@ -0,0 +1,586 @@
...
+
+== Optional values and unions
+
+    Dhall @Optional@ Dhall type allows null or missing JSON values:
+
+> $ json-to-dhall "Optional Integer" <<< '1'
+> Some +1
+
+> $ json-to-dhall "Optional Integer" <<< null
+> None
```
Gabriel439: I believe this should be `None Integer`

dhall-json/json-to-dhall/Main.hs
```diff
@@ -0,0 +1,586 @@
...
+
+> $ json-to-dhall '{ a : Integer, b : Optional Text }' <<< '{ "a": 1 }'
+{ a = +1, b = None Text }
+
+
+
+    For Dhall __union types__ the correct value will be based on matching the type of JSON expression:
+
+> $ json-to-dhall 'List < Left : Text | Right : Integer >' <<< '[1, "bar"]'
+> [ < Right = +1 | Left : Text >, < Left = "bar" | Right : Integer > ]
```
Gabriel439: The translation should probably use the new union selector syntax because the union literal syntax will be deprecated in the next release of the standard:

```
$ json-to-dhall 'List < Left : Text | Right : Integer >' <<< '[1, "bar"]'
[ < Left : Text | Right : Integer >.Right +1, < Left : Text | Right : Integer >.Left "bar" > ]
```

Similar comment for the following union-related examples

albertov:
This is very useful, thanks! Currently using an adapted version while it is in Main

dhall-json/json-to-dhall/Main.hs
antislava: Thank you. I fixed this.

dhall-json/json-to-dhall/Main.hs
antislava: @Gabriel439 I'm a little confused here: Aren't the two different representations supposed to have the same type constructor  (as one would expect, at least after the normalization) and differ only at the pretty-printing stage? If not, isn't/shouldn't there be an isomorphic function translating between these two?

I notice that in shell:
```sh
$ dhall <<< '[ < Right = +1 | Left : Text >, < Left = "bar" | Right : Integer > ]'
[ < Right = +1 | Left : Text >, < Left = "bar" | Right : Integer > ]

$ dhall <<< '[ < Left : Text | Right : Integer >.Right +1, < Left : Text | Right : Integer >.Left "bar"
 ]'
[ < Left : Text | Right : Integer >.Right +1
, < Left : Text | Right : Integer >.Left "bar"
]
```

while in ghci (now I do see that they actually have different constructors):
```haskell

>>> import qualified Dhall.Parser as D
>>> import qualified Dhall.Core as D

>>> Right uNew <- traverse Dhall.Import.load $ D.normalize <$> D.exprFromText "" "[ < Left : Text | Right : Integer >.Right +1, < Left : Text | Right : Integer >.Left \"bar\" ]"                                                                                                                                       
>>> Right uOld <- traverse Dhall.Import.load $ D.normalize <$> D.exprFromText "" "[ < Right = +1 | Left : Text >, < Left = \"bar\" | Right : Integer > ]"

>>> uNew                                                                                                                                                  
ListLit Nothing (fromList [App (Field (Union (fromList [("Left",Just Text),("Right",Just Integer)])) "Right") (IntegerLit 1),App (Field (Union (fromList [("Left",Just Text),("Right",Just Integer)])) "Left") (TextLit (Chunks [] "bar"))])                                                                                  

>>> uOld                                                                                                                                                  
ListLit Nothing (fromList [UnionLit "Right" (IntegerLit 1) (fromList [("Left",Just Text)]),UnionLit "Left" (TextLit (Chunks [] "bar")) (fromList [("Right",Just Integer)])])
```

Do you think there should be a switch for opting out of the new notation provided by the command-line tool?
I guess that we will also have to factor out the Union type definition in lists of union values for the sake of readability, perhaps not this time around though. 
(BTW, for lists  `[ typ.v1, typ.v2,... ]` longer than one the type prefix is strictly redundant since it must be the same. On the other hand for empty lists it is necessary. Spontaneously, it would seem logical to somehow move the type outside of the list itself by default?)

PR #891, opened 2019-04-12
Fix merging of sort-level record types.

### LINKED ISSUE #890: Can’t merge two record sorts.
```dhall
{ a : Kind } ⩓ { b : Kind }
```
currently fails with “Error: Record type mismatch”, which it should be a valid merge.

sellout:
Fixes #890.


sellout:
The dhall-lang submodule will need to be updated once dhall-lang/dhall-lang#486 is merged.

Gabriel439:
https://github.com/dhall-lang/dhall-lang/pull/486 is merged so if you update this branch then tests should now pass

sellout:
Looks like the `largeExpression` parser test now fails.

/cc @singpolyma, since they made the latest change to that test.

f-f:
@sellout yeah it looks like there's something with that test: now both dhall-haskell master and dhall-clj master produce the same parse for it, and the result differs from the one in dhall-lang. 
However https://github.com/dhall-lang/dhall-lang/pull/488 will change it again, so we could wait for that (though it will require changes in dhall-haskell)

singpolyma:
I've merged https://github.com/dhall-lang/dhall-lang/pull/488 now -- not sure how the failure happened before since I produce that `.dhallb` file previously with `dhall encode` -- now the version in master is produced with `dhall-ruby` and should be correct for the changes in 488

sellout:
Should this wait on #903, then?

Gabriel439:
@sellout: Yeah, if you update to the latest `master` then you should pick up all relevant changes

Gabriel439:
Thanks for fixing this! 🙂

PR #894, opened 2019-04-16
Add json-to-dhall to appveyor artifacts

jneira:
I had to [reduce genericity (from Semigroup to Monoid)](https://github.com/dhall-lang/dhall-haskell/pull/894/files#diff-f78f59aefb8f350f8c05c79481959a80L498) to make it buildable with ghc-7.10


antislava:
@jneira Thank you very much for looking into this! (It was on my own to-do list as well).
(I see hydra complaining about the redundant Semigroup import.)
I don't have a strong opinion on this but I see Gabriel wraps relevant parts of the code in pragma conditions, e.g. in `Dhall.Map`
```
#if !(MIN_VERSION_base(4,11,0))
    mappend = (<>)
    {-# INLINABLE mappend #-}
#endif
```
I guess, we should ideally stick to stick to his style (unless you are convinced otherwise?)

jneira:
@antislava oh, i did a quick search but i didnt find it, thank for pointed it out. 
Well, if we want keep (or recover) the `Semigroup` constraint it would be the way to go.

antislava:
Well, we can return to that at the next iteration, e.g. when moving the conversion logic to the library, which should be the next major step the way I see it (unless you have time to do this alignment before that, of course).

Issue #898, opened 2019-04-19
shell.nix for testing/using core Dhall command-line tools


antislava:
I was missing a user-friendly nix shell (with all the core Dhall executables including shell completions etc) in the root directory, along with `default.nix`: https://github.com/dhall-lang/dhall-haskell/compare/master...antislava:antislava/nix-shell.
I thought that maybe others would find it useful as well? 
At the very least, it would nice for `nix/shared.nix` to export `nixpkgs` (with the respective overlays etc) to facilitate building such shells by end users on an ad-hoc basis and doing development in `nix repl`, or, alternatively export the set of overlays (in which case it would also be nice to factor out the pinned base `nixpkgs` into a separate `nixpkgs.nix` as well to skip copy-pasting the `fetchNixpkgs` argument set or doing something as ugly as `import <nixpkgs> {}` ;-)).

Gabriel439:
@antislava: Yeah, you can open a pull request if you want to add a top-level `shell.nix`

PR #899, opened 2019-04-19
shell.nix with the core Dhall tools

antislava:
Issue #898

### DIFF
--- a/nix/shared.nix
+++ b/nix/shared.nix
@@ -563,6 +563,8 @@ in
   rec {
     inherit trivial;
 
+    nixpkgs = pkgs;
+
     possibly-static = {
       dhall            = makeStaticIfPossible "dhall"           ;
       dhall-bash       = makeStaticIfPossible "dhall-bash"      ;
--- a/nix/shared.nix
+++ b/nix/shared.nix
@@ -563,7 +563,7 @@ in
   rec {
     inherit trivial;
 
-    nixpkgs = pkgs;
+    inherit pkgs;
 
     possibly-static = {
       dhall            = makeStaticIfPossible "dhall"           ;
--- a/shell.nix
+++ b/shell.nix
@@ -3,7 +3,7 @@ let
   shared = import ./nix/shared.nix args;
   static = shared.possibly-static;
 in
-shared.pkgs.runCommand "dhall-shell-bash" {
+shared.pkgs.runCommand "dhall-shell-${shell}" {
   buildInputs = [
     static.dhall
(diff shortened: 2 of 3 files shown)


nix/shared.nix
```diff
@@ -563,6 +563,8 @@ in
   rec {
     inherit trivial;
 
+    nixpkgs = pkgs;
```
Gabriel439: I suggest `inherit pkgs;`, mainly because I try to reserve `nixpkgs` to refer to the Nixpkgs source code, rather than the package set

shell.nix
```diff
@@ -0,0 +1,24 @@
...
+shared.nixpkgs.runCommand "dhall-shell-bash" {
+  buildInputs = [
+    static.dhall
+    static.dhall-json
+    static.dhall-text
+    static.dhall-bash
+    static.dhall-nix
+  ];
+  shellHook = ''
+    echo -e "\e[0;32m""\nA Dhall shell""\e[0m"
```
Gabriel439: The user probably already knows that this is a Dhall shell of some form so this message does not add new information.  Perhaps remove it or replace with a message providing instructions for how to use the shell

antislava: Initially just deleted it but then decided to add some argument examples. Not quite sure if it is necessary. (I do normally keep such examples even for own usage in the comments in `source.nix`). Pls let me know what you think.

shell.nix
```diff
@@ -0,0 +1,32 @@
...
+    static.dhall-bash
+    static.dhall-nix
+  ];
+  shellHook = ''
+    echo "Dhall core tools shell"
+    echo "USAGE EXAMPLES"
+    echo "    Default (GHC version ${(import ./nix/shared.nix {}).pkgs.haskellPackages.ghc.version}, option completions shell - bash):"
+    echo "        $ nix-shell"
+    echo "    Overriding default GHC compiler (see ./nix/shared.nix for available options):"
+    echo "        $ nix-shell --arg args '{ compiler = \"ghc7103\"; }'"
```
Gabriel439:
```suggestion
    echo "        $ nix-shell --arg args '{ compiler = "ghc7103"; }'"
```

shell.nix
Gabriel439: Yeah, this looks good

antislava: Added a short paragraph about the top-level `nix-shell` usage in README.md if it is ok.

Gabriel439: Yeah, good idea!

PR #907, opened 2019-04-26
dhall-nix: Fix field-based union access

### LINKED ISSUE #906: dhall-to-nix generate incorrect Nix for sum types
```
$ /nix/store/hi06lr4hfq52krhkr99vdfxda836mbzz-dhall-nix-1.1.6/bin/dhall-to-nix <<< '<Foo>.Foo'
{}.Foo

$ nix-instantiate --eval-only -E $( /nix/store/hi06lr4hfq52krhkr99vdfxda836mbzz-dhall-nix-1.1.6/bin/dhall-to-nix <<< '<Foo>.Foo' )
error: attribute 'Foo' missing, at (string):1:1
```

`/nix/store/hi06lr4hfq52krhkr99vdfxda836mbzz-dhall-nix-1.1.6` is the latest `linux-dhall-nix` build: https://hydra.dhall-lang.org/build/10712#tabs-details

Note this also happens with sum types with arguments:

```
$ /nix/store/hi06lr4hfq52krhkr99vdfxda836mbzz-dhall-nix-1.1.6/bin/dhall-to-nix <<< '<Foo : Bool >.Foo True'
{}.Foo true
```

ocharles:
Fixes #906.

I copied the approach used in `dhall-json`.

### DIFF
--- a/dhall-nix/src/Dhall/Nix.hs
+++ b/dhall-nix/src/Dhall/Nix.hs
@@ -208,6 +208,13 @@ dhallToNix e = loop (Dhall.Core.normalize e)
     -- None needs a type to convert to an Optional
     loop (App None _) = do
       return (Fix (NConstant NNull))
+    loop (App (Field (Union kts) k) v) = do
+        v' <- loop v
+        let e0 = do
+                k' <- Dhall.Map.keys kts
+                return (k', Nothing)
+        let e2 = Fix (NBinary NApp (Fix (NSym k)) v')
+        return (Fix (NAbs (ParamSet e0 False Nothing) e2))
     loop (App a b) = do
         a' <- loop a
         b' <- loop b
@@ -488,6 +495,12 @@ dhallToNix e = loop (Dhall.Core.normalize e)
         a' <- loop a
         b' <- loop b
         return (Fix (NBinary NUpdate a' b'))
+    loop (Field (Union kts) k) = do
+        let e0 = do
+                k' <- Dhall.Map.keys kts
+                return (k', Nothing)
+        let e2 = Fix (NSym k)
+        return (Fix (NAbs (ParamSet e0 False Nothing) e2))
     loop (Field a b) = do
         a' <- loop a
         return (Fix (NSelect a' [StaticKey b] Nothing))
--- a/dhall-nix/src/Dhall/Nix.hs
+++ b/dhall-nix/src/Dhall/Nix.hs
@@ -496,25 +496,25 @@ dhallToNix e = loop (Dhall.Core.normalize e)
         b' <- loop b
         return (Fix (NBinary NUpdate a' b'))
     loop (Field (Union kts) k) =
-        case join ( Dhall.Map.lookup k kts ) of
-            Nothing -> do
-                let e0 = do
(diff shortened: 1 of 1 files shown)


dhall-nix/src/Dhall/Nix.hs
```diff
@@ -488,6 +495,12 @@ dhallToNix e = loop (Dhall.Core.normalize e)
         a' <- loop a
         b' <- loop b
         return (Fix (NBinary NUpdate a' b'))
+    loop (Field (Union kts) k) = do
+        let e0 = do
+                k' <- Dhall.Map.keys kts
+                return (k', Nothing)
+        let e2 = Fix (NSym k)
+        return (Fix (NAbs (ParamSet e0 False Nothing) e2))
```
Gabriel439: If I understand correctly, this compiles `< Foo >.Foo` to `{ Foo }: Foo`, but I believe the result should instead be: `x: { Foo }: Foo x`

ocharles: `<Foo>.Foo` compiling to `{Foo}: Foo` seems correct to me - what is that `x` about?

Gabriel439: Oh, you're right.  Never mind, my thoughts were confused

Gabriel439: Now I remember what led to my confusion.  The issue here is that you need to check if the alternative is empty when compiling the union to Nix.

For example:

```nix
dhallToNix "< Foo >.Foo" = { Foo }: Foo

dhallToNix "< Foo : {} >.Foo {=}" = { Foo }: Foo {}

dhallToNix "< Foo : {} >.Foo" = x: { Foo }: Foo x
```

Carefully note the difference between the 1st and 3rd examples.  If the type of the alternative is present then you need to turn it into an anonymous function.

ocharles: Ah, so it's a case of `< .. >.x` being partially applied when the alternative does take data

ocharles: Ok, I believe my latest commit resolves this.

ocharles:
Here's the latest behaviour:

```
[nix-shell:~/work/dhall-haskell/dhall-nix]$ ./dist/build/dhall-to-nix/dhall-to-nix <<< '<Foo : {} | Bar>.Foo'
x:
  { Bar, Foo }:
    Foo x

[nix-shell:~/work/dhall-haskell/dhall-nix]$ ./dist/build/dhall-to-nix/dhall-to-nix <<< '<Foo : {} | Bar>.Foo {=}'
{ Bar, Foo }:
  Foo {}

[nix-shell:~/work/dhall-haskell/dhall-nix]$ ./dist/build/dhall-to-nix/dhall-to-nix <<< '<Foo : {} | Bar>.Bar'
{ Bar, Foo }:
  Bar
```

Gabriel439:
@ocharles: I think you need to import `join`: https://hydra.dhall-lang.org/build/10741/nixlog/2/tail

ocharles:
Ah no, I just need to push another commit

PR #911, opened 2019-04-27
Add `dhall-to-nix` to Appveyor build

Gabriel439:


jneira:
Not sure if `dhall-to-nix` can be built in windows without changes: hnix dependency depends on `unix` package

Gabriel439:
@jneira: Oh yeah, good point 🙂

PR #912, opened 2019-04-27
Use MonadFail.fail, not Monad.fail.

quasicomputational:
This is a breaking change in API.

Monad.fail is no more with GHC 8.8, so this is a forced change.
However, it will also work on older GHCs.

This looks like most of the GHC 8.8 fallout for Dhall, so not much at all; the only other change I had to make to get it to build is to ifdef out a `Monad.fail` definition, and then presumably a load of bounds will need to be adjusted, as well as waiting on dependencies.

This isn't actually ready to merge yet because it needs sdiehl/repline#13 to be merged and then to make it into a release.


quasicomputational:
Oh, never mind about it being a breaking change - all of the things I touched aren't exported, so consumers won't notice the difference!

Gabriel439:
@quasicomputational: It looks like `repline` needs to add a `MonadFail` instance for `HaskelineT`:

https://hydra.dhall-lang.org/build/10819/nixlog/1

quasicomputational:
Ah, here's trouble: there aren't `MonadFail` instances for the version of `transformers` that shipped with GHC 7.10. But, as noted, using `MonadFail` is mandatory on 8.8.

Options here:
* CPP to switch between `Monad.fail` and `MonadFail.fail`.
* Use a local definition of `StateT` with the instances we need.
* Drop 7.10 support.

Is the support window documented anywhere? Either of the first two options are basically hacks to support 7.10, and eventually they would go away when that's no longer relevant - but we'd need to know when that happens.

quasicomputational:
Aha - `transformers-compat` has exactly the orphan instances we need. Crisis averted.

quasicomputational:
Now it's getting caught on `transformers-compat`'s complicated configuration - it needs a different flag setting for each `transformers` version it builds against, and hence for each GHC. I'm not sure my Nix skills are up to this...

Gabriel439:
@quasicomputational: Don't worry, you did the right thing!  It's okay to unconditionally add the `fail` build dependency to `transformers-compat`.  Older versions will just ignore it and not use it

Issue #915, opened 2019-04-28
1.22: union construction seems not to be normalised


quasicomputational:
Running this program against current HEAD (44a2750ca0c602b9c088937d938e22ebe32f23ba)

```haskell
{-# LANGUAGE OverloadedStrings #-}

import qualified Dhall

license :: Dhall.Type ()
license = Dhall.union $ mconcat
  [ Dhall.constructor "AllRightsReserved" Dhall.unit
  ]

main :: IO ()
main = do
  let
    settings = Dhall.defaultInputSettings
  Dhall.input
    license
    "< AllRightsReserved : {} >.AllRightsReserved {=}"

```

results in this output

```
foobar: Error: Invalid Dhall.Type                                                  
                                                                                
Every Type must provide an extract function that succeeds if an expression      
matches the expected type.  You provided a Type that disobeys this contract     
                                                                                
The Type provided has the expected dhall type:                                  
                                                                                
↳ < AllRightsReserved : {} >
                                                                                
and it couldn't extract a value from the well-typed expression:                 
                                                                                
↳ < AllRightsReserved : {} >.AllRightsReserved {=}
```

By copying in the definition of `union`, `unionType`, etc, and sticking in some judicious debugging prints, the source of the failure's narrowed to this statement:

https://github.com/dhall-lang/dhall-haskell/blob/44a2750ca0c602b9c088937d938e22ebe32f23ba/dhall/src/Dhall.hs#L1538

The expression isn't a `UnionLit`, but rather an `App` of the field of the union and then the empty record. It looks very much like something's not being normalised properly - or perhaps this is expected in 1.22, and `union` needs to learn how to massage the expression into the form that it needs?

quasicomputational:
Also, I should note that this worked in 1.21 (or, well, something with very much the same shape did), so either it's a regression or needs a heads-up in the changelog.

AndrasKovacs:
You're right, it appears that `Dhall.union` only accepts old-style union literals. I am not sure what the fix should be, of the following two alternatives: a) parsing both union styles in `Dhall.union` b) parsing only new-style unions, but normalizing old-style to the new style.

Gabriel439:
Yeah, the bug here is that `Dhall.union` doesn't accept new-style union literals.  In version 7.0.0 of the standard union literals have two valid normal forms (which will go away once we remove the old-style union literals).  See:

[...]

PR #918, opened 2019-04-30
Fix marshaling new-style union literals

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/915


quasicomputational:
LGTM!

PR #920, opened 2019-04-30
Converted sample code to doctest for `RecordType`, `UnionInputType`, `RecordInputType`

klntsky:
`<>` is `infixr 6`, `<$>` is `infixr 4`, therefore existing example is incorrect.

It might be a good idea to reflect every piece of sample code in the tests, so that the docs will never become outdated.

I could dedicate some time to this. Should I add a test to `Dhall.Test.Tutorial` or is there a more suitable module?

### DIFF
--- a/dhall/src/Dhall.hs
+++ b/dhall/src/Dhall.hs
@@ -1506,12 +1506,10 @@ field key valueType =
     However, we can use a 'UnionType' to build a 'Type' for @Status@:
 
 > status :: Type Status
-> status =
->   union
->     ( Queued  <$> constructor "Queued"  natural
->    <> Result  <$> constructor "Result"  string
->    <> Errored <$> constructor "Errored" string
->     )
+> status = union
+>    $  ( Queued  <$> constructor "Queued"  natural )
+>    <> ( Result  <$> constructor "Result"  strictText )
+>    <> ( Errored <$> constructor "Errored" strictText )
 
 -}
 newtype UnionType a =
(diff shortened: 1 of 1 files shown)


f-f:
I thought we had doctests running? 🤔

quasicomputational:
`doctest` only recognises the `>>>` prefix.

Gabriel439:
I believe this can be turned into a doctest using `ghci`'s support for multi-line expressions, like this:

```haskell
>>> data Status = Queued Natural | Result | Text | Errored Text
>>> :{
let status :: Type Status
    status = union
        $   (Queued <$> constructor "Queued" natural )
        <> ( Result  <$> constructor "Result"  strictText )
        <> ( Errored <$> constructor "Errored" strictText )
:}
```

klntsky:
Both `injectStatus` examples were broken too (`inputUnion` call was missing).

klntsky:
By the way, `dhall:doctest` suite [is skipped](https://github.com/dhall-lang/dhall-haskell/blob/684cbac7be0c28ce60dbe089e1f9ae6c95f6f13f/appveyor.yml#L60) in `appveyor.yml`. Is it intentional?

Gabriel439:
@klntsky: The `doctest` omission for the `dhall` package appears to be intentional.  See:

https://github.com/dhall-lang/dhall-haskell/pull/785

> Not including dhall:doctests cause throws spurious errors due to encoding. I've not found a way to avoid them like in local (setting encoding of console to utf-8)

Gabriel439:
@klntsky: Thank you! 🙂

PR #923, opened 2019-05-01
Nix: Upgrade from 8.6.1 to 8.6.5.

quasicomputational:
8.6.1 had a couple of really awful bugs in it - including two that
were being worked around in the derivations. Upgrading will both
simplify the configuration and mean that there's less risk of a crazy
bug biting in CI.

### DIFF
--- a/dhall-bash/dhall-bash.cabal
+++ b/dhall-bash/dhall-bash.cabal
@@ -2,7 +2,7 @@ Name: dhall-bash
 Version: 1.0.20
 Cabal-Version: >=1.8.0.2
 Build-Type: Simple
-Tested-With: GHC == 7.10.3, GHC == 8.4.3, GHC == 8.6.1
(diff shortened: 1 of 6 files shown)


Gabriel439:
@quasicomputational: The reason I haven't upgraded to GHC 8.6.5 yet in CI is because we're pinned to a revision of Nixpkgs that precedes it because we depend on it for static linking of executables (See: https://github.com/NixOS/nixpkgs/issues/43795).  Once full static linking support is available in Nixpkgs `master then we can upgrade.

As a result, you get an evaluation error on this branch: https://hydra.dhall-lang.org/jobset/dhall-haskell/923#tabs-errors

Alternatively, we could backport the GHC 8.6.5 derivation to the revision of Nixpkgs that we currently use

quasicomputational:
Makes sense. I'll open an issue for this, since I think it's something we really ought to do (those bugs are scary!), but there are evidently complications.

Issue #926, opened 2019-05-01
genericAuto is broken


quasicomputational:
Running against 4f9defec2516edab4707f85c1bfec7e1eac09f72, this program

```haskell
{-# LANGUAGE DeriveGeneric #-}
{-# LANGUAGE OverloadedStrings #-}
import GHC.Generics (Generic)
import qualified Dhall

data CompilerFlavor =
  GHC | GHCJS
  deriving (Generic, Show)

main =
  print =<< Dhall.input
    (Dhall.genericAuto :: Dhall.Type CompilerFlavor)
    "< GHC : {} | GHCJS : {} >.GHC {=}"
```

produces this output

```
foobar: Error: Invalid Dhall.Type                                                  
                                                                                
Every Type must provide an extract function that succeeds if an expression      
matches the expected type.  You provided a Type that disobeys this contract     
                                                                                
The Type provided has the expected dhall type:                                  
                                                                                
↳ < GHC : {} | GHCJS : {} >
                                                                                
and it couldn't extract a value from the well-typed expression:                 
                                                                                
↳ < GHC : {} | GHCJS : {} >.GHC {=}
```

The unary case also seems to be broken - if you delete the `GHCJS` constructor in both Haskell and the Dhall input, the output is instead

```
Error: Expression doesn't match annotation

- { … : … }
+ < … : … >

1: < GHC : {} >.GHC {=} : {}

(input):1:1
```

but I'd also expect it to have worked.

This code seems to have worked in 1.21, so this is a regression that's probably affecting 1.22, though I haven't actually tested this specific example on the as-released versions.

quasicomputational:
Here's the bug - it's quite like #915:

https://github.com/dhall-lang/dhall-haskell/blob/66df01b26c0985363eba0af1edd9d9cd481b7357/dhall/src/Dhall.hs#L956-L960

That needs to learn about the new style of union. I'll see about a PR.

quasicomputational:
Ah - the single constructor case isn't a bug at all. It just wants to be a record, not a union, which does actually make perfect sense.

Gabriel439:
@quasicomputational: Also, keep in mind that it should expect a type of `< GHC | GHCJS >` now

quasicomputational:
@Gabriel439: I think that's a separable change. This issue's about just getting it to work at all again for sum types; making it use the snazzy new syntax for nullary constructors doesn't overlap (that) much, I don't think.

Gabriel439:
[...]

PR #927, opened 2019-05-01
Teach genericAuto about new-style union construction.

quasicomputational:
I discovered it wasn't working right in #926. The underlying bug is
very similar to #915, and in fact this builds on the fix for that in #918.

It looks like the single-constructor case is a different underlying
bug that I don't understand just yet, so let's tackle that separately.

### DIFF
--- a/dhall/src/Dhall.hs
+++ b/dhall/src/Dhall.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiWayIf                 #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE RankNTypes                 #-}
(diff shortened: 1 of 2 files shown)


dhall/src/Dhall.hs
```diff
@@ -976,10 +988,11 @@ instance (Constructor c, GenericInterpret (f :+: g), GenericInterpret h) => Gene
...
 
-        extract u@(UnionLit name' e _)
-            | name == name' = fmap (R1 . M1) (extractR e)
-            | otherwise     = fmap  L1       (extractL u)
-        extract _ = Nothing
+        extract u = do
+          (name', e, _) <- extractUnionConstructor u
+          if
+            | name == name' -> fmap (R1 . M1) (extractR e)
+            | otherwise     -> fmap  L1       (extractL u)
```
quasicomputational: This instance's implementation is actually not covered by the new tests - GHC seems to prefer to recurse to the right and hence the following instance gets used, though I didn't try very hard to convince it not to.

quasicomputational:
Thanks!

Gabriel439:
You're welcome! 🙂

PR #928, opened 2019-05-01
Add 'cross' flag to simplify cross-compilation

vmchale:
This adds a 'cross' flag to the .cabal file for Dhall, which disables the `Dhall.TH` module.

This makes cross-compilation much simpler. 

I've also added to the `other-extensions` field of the `.cabal` file so that the new solver will be able to use this information.


dhall/dhall.cabal
```diff
@@ -376,6 +376,11 @@ Flag with-http
   Default:     True
   Manual:      True
 
+Flag cross
+  Description: Disable TemplateHaskell to make cross-compiling eaiser
```
Gabriel439:
```suggestion
  Description: Disable TemplateHaskell to make cross-compiling easier
```

Gabriel439:
Thank you! 🙂

Issue #929, opened 2019-05-02
Bug: current Prelude cannot be cached as it fails to deserialize


f-f:
If I have a `Prelude.dhall` containing the following:
```hs
https://raw.githubusercontent.com/dhall-lang/dhall-lang/v7.0.0/Prelude/package.dhall
```

Then `dhall freeze` errors out:

```bash
$ dhall freeze --inplace Prelude.dhall

Use "dhall --explain" for detailed errors

_ : Bool
_ : Type
_ : _@2
_ : _@2

Error: Invalid predicate for ❰if❱

```

This happens with the latest release `1.22.0` and on `master`. Also the error with `--explain` is not much more helpful than this one

f-f:
Actually the issue seems to be that the deserialization doesn't roundtrip: putting a hash on the above import works the first time you evaluate the expression, but not the second (because it deserializes from the cache)

Gabriel439:
@f-f: It would still help me to see the error message with `--explain`

f-f:
@Gabriel439 sorry, here it is:

```bash
$ dhall --explain <<< "./Prelude.dhall"


_ : Bool
_ : Type
_ : _@2
_ : _@2

Error: Invalid predicate for ❰if❱

Explanation: Every ❰if❱ expression begins with a predicate which must have type
❰Bool❱

For example, these are valid ❰if❱ expressions:


    ┌──────────────────────────────┐
    │ if True then "Yes" else "No" │
    └──────────────────────────────┘
         ⇧
         Predicate


    ┌─────────────────────────────────────────┐
    │ λ(x : Bool) → if x then False else True │
    └─────────────────────────────────────────┘
                       ⇧
                       Predicate


... but these are not valid ❰if❱ expressions:


    ┌───────────────────────────┐
    │ if 0 then "Yes" else "No" │  ❰0❱ does not have type ❰Bool❱
    └───────────────────────────┘


    ┌────────────────────────────┐
    │ if "" then False else True │  ❰""❱ does not have type ❰Bool❱
    └────────────────────────────┘


Some common reasons why you might get this error:

● You might be used to other programming languages that accept predicates other
  than ❰Bool❱

  For example, some languages permit ❰0❱ or ❰""❱ as valid predicates and treat
  them as equivalent to ❰False❱.  However, the Dhall language does not permit
  this

────────────────────────────────────────────────────────────────────────────────

Your ❰if❱ expression begins with the following predicate:

↳ _@1

... that has type:

[...]

PR #931, opened 2019-05-02
Fix alpha normalization (#929)

AndrasKovacs:
Trivial typo fix.

### DIFF
--- a/dhall/src/Dhall/Eval.hs
+++ b/dhall/src/Dhall/Eval.hs
@@ -878,7 +878,7 @@ alphaNormalize = goEnv NEmpty where
       BoolOr t u       -> BoolOr  (go t) (go u)
       BoolEQ t u       -> BoolEQ  (go t) (go u)
       BoolNE t u       -> BoolNE  (go t) (go u)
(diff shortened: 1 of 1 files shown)


Gabriel439:
I'll also cut a new release once this and all of the union-related issues are fixed

ocharles:
May I suggest turning off -fno-warn-unused?

jneira:
Also, have you considered setup a little bit slower release process with beta or/and alpha releases (for one week max?)? Not sure if it could help to avoid quick fix minor releases after major ones

AndrasKovacs:
@ocharles agree, I intended but forgot to remove that flag

Gabriel439:
@jneira: I would prefer to increase the test coverage or improve our use of the compiler/type-checker rather than slow down the release process.  For example: as @ocharles  suggested, this would have been caught by removing the `-fno-warn-unused` flag

The reason I prefer not to slow down the release process or to have stabilization periods is because in my experience it leads to less frequent releases and higher coordination costs.  It also outsources the burden of quality to our users.

Issue #930, opened 2019-05-02
Can't load a partial `merge` application as a Haskell function


blamario:
I thought I came up with a clever way to reduce my configuration size and make it safer at the same time like so:

```
let    schemaPath = "core/configurations/generic_DITA_1.2/converter_server/schema"
-- in let Format = <DITA : {} | Ditamap : {} | Bookmap : {}>
in let Format = <DITA | Ditamap | Bookmap>
in let FormatConfig = { dtdSystemPathPrefix : Text
                      , originalSchema : Text
                      , xmlNamespace : Text}
in let Configuration = { preloadFormats : List < DITA | bookmap | ditamap >
                       , logFilePath : Text
                       , timeout : Natural
                       , port : Natural
                       , publicIdentifierMap : List { mapKey : Text, mapValue : Format }
                       , rootElementMap : List { mapKey : Text, mapValue : Format }
                       , formats : Format -> FormatConfig}

in {preloadFormats = [Format.DITA, Format.Ditamap, Format.Bookmap],
    logFilePath = "${env:CONVERTER_SERVER_LOGFILE as Text}" ? "converter-server.log",
    timeout = env:CONVERTER_SERVER_TIMEOUT ? 300,
    port = env:CONVERTER_SERVER_PORT ? 8001,
    publicIdentifierMap = [{mapKey = "-//OASIS//DTD DITA Concept//EN",   mapValue= Format.DITA},
                           {mapKey = "-//OASIS//DTD DITA Reference//EN", mapValue= Format.DITA},
                           {mapKey = "-//OASIS//DTD DITA Task//EN",      mapValue= Format.DITA},
                           {mapKey = "-//OASIS//DTD DITA Topic//EN",     mapValue= Format.DITA},
                           {mapKey = "-//OASIS//DTD DITA Composite//EN", mapValue= Format.DITA},
                           {mapKey = "-//OASIS//DTD DITA Map//EN",       mapValue= Format.Ditamap},
                           {mapKey = "-//OASIS//DTD DITA BookMap//EN",   mapValue= Format.Bookmap}],
    rootElementMap = [{mapKey = "concept",   mapValue= Format.DITA},
                      {mapKey = "reference", mapValue= Format.DITA},
                      {mapKey = "task",      mapValue= Format.DITA},
                      {mapKey = "topic",     mapValue= Format.DITA},
                      {mapKey = "dita",      mapValue= Format.DITA},
[...]

PR #936, opened 2019-05-04
Fix marshalling of unions (including enums) in and out of Dhall

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/930

### DIFF
--- a/dhall/src/Dhall.hs
+++ b/dhall/src/Dhall.hs
@@ -1768,7 +1768,9 @@ inputUnion ( UnionInputType ( Data.Functor.Product.Pair ( Control.Applicative.Co
     InputType
       { embed = \x ->
           let (name, y) = embedF x
-          in  UnionLit name y (Dhall.Map.delete name fields')
+          in  case notEmptyRecordLit y of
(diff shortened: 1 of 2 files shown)


quasicomputational:
Some cases that I think we'll want tests for:

* Haskell: `data Foo = A () | B`, using `genericAuto` to extract. Dhall: `<A : {} | B>.A` - shouldn't work as a union constructor, I think? And it shouldn't typecheck against `<A | B>`.

* The above, but testing as many branches of the generic machinery as we can.

* `union $ constructor "A" unit` likewise shouldn't typecheck against `<A>`, and `<A : {}>.A` should have a function type, not a union constructor type.

Gabriel439:
@quasicomputational: Note that `< A : {} | B>.A` is a valid union constructor, which has type `{} → < A : {} | B >`.  However, I can still add the tests you suggested

Gabriel439:
@quasicomputational: Also, note that I've decided to make `union`/`inputUnion` behave the same as the generic-deriving machinery for now (i.e. treating an empty record as an empty alternative)

PR #939, opened 2019-05-05
Reduce dependency requirements to what is in sid

singpolyma:
This make it easier to build the package on debian unstable, which is
the first step towards an official package.


Gabriel439:
Thank you! 🙂

PR #940, opened 2019-05-05
Add and expose genericInject.

quasicomputational:
This is a counterpart to `genericAuto`, which is useful for dealing
with types coming from a library without orphan instances.


Gabriel439:
Good idea!  Thanks 🙂

Issue #933, opened 2019-05-03
Double quotes with a text value


erthalion:
Most likely it's a strange question, but I'm struggling with one aspect of dhall.

I want to generate some yaml, that should be processed by another service. The problem is with one value, which happened to be a date in a text format (do not ask me why). Due to quirks of a parsing library, that this service uses, the value should be quoted (like `value: "2000-00-00"`), otherwise it would be parsed as a date time. Of course dhall generates it as `value: 2000-00-00` and all my attempts to put single quotes or load value as a text didn't succeed (I've got `value: ! '2000-00-00'` or `value:|\n2000-00-00` etc, but it didn't help to trick this third party library). So I wonder if it's possible to generate `value: "2000-00-00"` via dhall?

f-f:
@erthalion would this work?

```bash
dhall-to-yaml <<< 'Text/show "2000-00-00"'
'"2000-00-00"'
```

erthalion:
Thanks @f-f - unfortunately no, I've tried this too.

Gabriel439:
Yeah, this can't be fixed within the language.  The place to fix this is in our use of the Haskell `yaml` package.  My guess is that by default it's opinionated about whether or not to quote strings, but this behavior might be configurable.

erthalion:
> Yeah, this can't be fixed within the language. The place to fix this is in our use of the Haskell yaml package. My guess is that by default it's opinionated about whether or not to quote strings, but this behavior might be configurable.

After a short investigation I found out that there are EncodingOptions in the yaml package, so something like that would work for me:

```haskell
isDate s = case parsed :: Maybe UniversalTime of
  Just v  -> True
  Nothing -> False
  where
    parsed = parseTimeM False defaultTimeLocale "%Y-%m-%d" (T.unpack s)

customStyle = \s ->
     case () of
      ()
        | isDate s  -> (NoTag, SingleQuoted)
        | otherwise -> (NoTag, PlainNoTag)

main = print $ encodeWith opts (Object {value = "2000-01-01"})
  where
    opts = setStringStyle customStyle defaultEncodeOptions

-- Produces "version: '2017-09-20'\n"
```

Gabriel439:
@erthalion: Do you need command-line support for that or does the solution with the Haskell API work for you?

erthalion:
I'm actually not sure how to incorporate it easily via just API, so I guess command-line support would be great (as an example of what I'm talking about I've created #941 )

PR #941, opened 2019-05-07
[POC] Add 'quoted' option

erthalion:
For #933 

Allow to generare quoted scalars if needed via providing a custom encode
options to Data.Yaml.encodeWith. So far two corner cases from yaml
itself (an empty scalar, and special strings) are omitted in the
implementation.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,7 @@ environment:
   # Override the temp directory to avoid sed escaping issues
   # See https://github.com/haskell/cabal/issues/5386
   TMP: c:\tmp
-  
+
   matrix:
     # Commenting out for now default stack (lts-13) cause
     # compilations times are reaching appveyor default build timeout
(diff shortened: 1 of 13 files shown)


Gabriel439:
@erthalion: If you merge https://github.com/erthalion/dhall-haskell/pull/1 into your branch then I think it will fix the build failures in CI

Gabriel439:
@erthalion: Hmmm.  It looks like you will need to upgrade `stack-lts-6.yaml` and `stack-lts-12.yaml` to use `yaml-0.10.4.0` in order to pass the AppVeyor build because older `yaml` versions don't support customizing the rendering

Gabriel439:
@erthalion: One solution would be to disable the `dhall-json` build for `stack-lts-6.yaml`.  The only reason we test against LTS 6 is for Eta's Etlas package manager (which supports Dhall as a configuration language and is built using LTS 6), but that only needs `dhall` to build and not `dhall-json`. (cc: @jneira)

Another solution would be to use CPP to conditionally compile out the `--quoted` option feature (i.e. `#if MIN_VERSION_yaml(0,10,4)`) so that it can build against older versions of `yaml`

You can also try `allow-newer: true`, but my guess is that `yaml`'s lower bound on `base` exists for a reason and all that will happen if you do that is that the `yaml` package build will fail.

My personal preference is the first solution (discontinue `dhall-json` support for LTS 6)

jneira:
Well, really the component needed for etlas is only dhall-haskell, so we could remove dhall-json from lts-6.
Otoh eta is somewhat between ghc-7.10 and ghc-8.x, so packages buildables with 7.10 usually can be built with eta (but not always the other way around). 
In this case the change would not impact eta cause it uses base-4.11.1.0 and cant use yaml anyway cause it is based in c ffi, so the package has to be patched.
(Btw i am working in a dhall-json version that could be built with ghc, eta and hopefully ghcjs)

erthalion:
> My personal preference is the first solution (discontinue dhall-json support for LTS 6)
> so we could remove dhall-json from lts-6

Stupid question, how to do this? I thought it would be enough just to remove `dhall-json` from packages in `stack-lts-6.yaml`, but looks like it's not.

Gabriel439:
@erthalion: The `appveyor.yml` script needs to be changed to not attempt to package `dhall-json` into a TAR archive for `stack-lts6.yaml`.  I'm still new to appveyor myself, but I believe this entails separating out the package step into a block that doesn't run for `stack-lts6.yaml`, similar to what is done for the tests here:

https://github.com/dhall-lang/dhall-haskell/blob/2530207abbce71bc36c55e33be548e5a7f23db6e/appveyor.yml#L52-L64

erthalion:
Indeed, @Gabriel439 it worked, thanks for the explanation. I'm going to write a few real tests now.

erthalion:
To make it testable I've introduced `jsonToYaml` function in the `dhall-json` library itself (which brought yaml dependencies there), is it acceptable?

Gabriel439:
@erthalion: Yeah, that's totally fine.  I usually treat the executable and library dependencies as essentially synonymous.  The only dependencies I treat differently are test and benchmark dependencies (since they are not always used).

Also, for this package we can be quite liberal with dependencies since it's most commonly use as an executable rather than as a package.

Gabriel439:
Excellent job!  Thank you for doing this :slightly_smiling_face:

Gabriel439:
You're welcome!

PR #942, opened 2019-05-07
Set syntax highlighting for Dhall files

robbiemcmichael:
Syntax highlighting for Dhall files, refer to https://github.com/dhall-lang/dhall-lang/pull/533 for details.


Gabriel439:
@robbiemcmichael: Thank you! 🙂

PR #943, opened 2019-05-08
Version 1.22.0 → 1.22.1

Gabriel439:


Gabriel439:
The main motivation for cutting this release is to get out the fixes for α-normalization and union marshaling

dhall/CHANGELOG.md
```diff
@@ -1,3 +1,49 @@
...
+* BUG FIX: Fix merging of sort-level record types
+    * The language standard requires that `{ a : Kind } ⩓ { b : Kind }` is
+      valid, which this change fixes
+    * See: https://github.com/dhall-lang/dhall-haskell/pull/891
+* BUG FIX: `dhall freeze` respects the `--ascii` flag
+    * See: https://github.com/dhall-lang/dhall-haskell/pull/934
+* BUG FIX: Don't autocomplete fields for record types
+    * This prevents the REPL from expanding `{ x : T }.<TAB>` to `{ x : T }.x`
+    * See: https://github.com/dhall-lang/dhall-haskell/pull/937
+* Support GHC 8.8
```
quasicomputational: Still one or two things to do before 8.8 support is fully baked and ready to roll.

dhall/CHANGELOG.md
```diff
@@ -1,3 +1,49 @@
+1.22.1
+
+* BUG FIX: Fix marshaling union literals
+    * 1.22.0 introduced two separate bugs in marshaling union literals between
+      Dhall and Haskell, which this release fixes:
+        * Dhall enums did not correctly map onto Haskell enums
+        * New-style union literals (i.e. `< A : T >.A x`) were not correctly
+          supported
+    * See: https://github.com/dhall-lang/dhall-haskell/pull/918
+    * See: https://github.com/dhall-lang/dhall-haskell/pull/927
```
quasicomputational: This one was actually a breaking change, wasn't it? The Dhall types expected by `genericAuto` are different now in that they use the nullary syntax.

Gabriel439: @quasicomputational: Actually, I meant to include #936 in this list

The reason I didn't classify it as a breaking change is that it is fixing a bug.  The Dhall types expected by `genericAuto` in 1.22.0 were not correct.

Or in other words, the breaking change included in this fix should have been included as part of the bundle of breaking changes for 1.22.0 had there not been a bug.

quasicomputational: Unfortunately, `genericAuto` wasn't so broken as to be totally unusable.

Consider, for example, a program that accepts Dhall input, and only uses `genericAuto` for an `Optional` value that's very rarely actually `Some`. This would have worked when built against 1.22.0 except in the rare case where it's `Some`, but the same input will cause type errors when the program's built against 1.22.1.

Users will have to adapt code that was previously correct; there's also neither forwards nor backwards compatibility between 1.22.0 and 1.22.1, so the maintainer will probably have to retroactively revise the bound to `dhall >= 1.22.0 && < 1.22.1`. Those are signal features of a breaking change that should have a major bump.

Gabriel439: Alright, then I'll bump the version number

PR #946, opened 2019-05-11
Change YAML/JSON encoding for `NaN`/`Infinity`/`-Infinity`

Gabriel439:
Before, `dhall-to-json`/`dhall-to-yaml` would use approximate
representations for special `Double` values.  Specifically, `NaN`
would encode as `null` and `±Infinity` would encode as the minimum
and maximum `Double` values.

After this change, YAML will now use `nan`/`inf`/`-inf` to encode
these values (since special `Double` values are natively supported
by YAML) and the JSON encoding will reject them by default.  The
user can restore the old behavior for the JSON encoding by enabling
the `--approximate-special-doubles` flag.

### DIFF
--- a/dhall-json/dhall-json.cabal
+++ b/dhall-json/dhall-json.cabal
@@ -36,6 +36,7 @@ Library
         base                      >= 4.8.0.0  && < 5   ,
         aeson                     >= 1.0.0.0  && < 1.5 ,
         dhall                     >= 1.22.0   && < 1.24,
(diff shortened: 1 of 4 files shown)


Gabriel439:
Note that https://github.com/dhall-lang/dhall-haskell/pull/941 removes `dhall-json` from the `stack-lts-6.yaml` build, which should fix the Appveyor build failure for this pull request once that is merged

PR #949, opened 2019-05-13
Add `--file` flag

Gabriel439:
This allows users to supply the Dhall expression by path instead of via
standard input.  This also ensures that transitive imports are resolved
appropriately relative to the file's path.

In other words, instead of this:

```
$ dhall <<< './foo/bar.dhall'
```

Users can now do:

```
$ dhall --file foo/bar.dhall
```

### DIFF
--- a/dhall/src/Dhall/Main.hs
+++ b/dhall/src/Dhall/Main.hs
@@ -262,6 +262,7 @@ getExpression :: Maybe FilePath -> IO (Expr Src Import)
 getExpression maybeFile = do
     inText <- do
         case maybeFile of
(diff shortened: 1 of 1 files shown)


ocharles:
Have we thought about just using a positional argument for the file, and
interpreting - as 'read from STDIN'? This is quite a common Unix idiom

Gabriel439:
@ocharles: The main reason I wasn't sure about using a positional argument is because I thought we might want to leave the door open for a parallel `--expression` flag for supplying the expression via the command line instead of standard input.

However, I can definitely add support for the idiom of `-` meaning `STDIN`

PR #952, opened 2019-05-14
Fix diff output to not double-print key-value separators.

sellout:
E.g.,
```
{ unit :   …
```
was being printed as
```
{ unit : :   …
```


Gabriel439:
Thank you! 🙂

PR #953, opened 2019-05-15
Fix broken link

Gabriel439:
Fixes https://discourse.dhall-lang.org/t/broken-link-on-dhall-lang-org/87

... as caught by @philandstuff


Gabriel439:
@philandstuff: You're welcome! 🙂

PR #954, opened 2019-05-15
Add chunkExprs.

quasicomputational:
I found myself needing this combinator, and I think it makes sense to upstream it - it's even got an immediate use site in `subExpressions`.


Gabriel439:
Great idea!  Thank you 🙂

PR #956, opened 2019-05-17
Add dependent types.

sellout:
Wondering what your feelings are on this or something like it (obviously it requires a dhall-lang/dhall-lang PR first, but I like to experiment). I added one large awkward test case here, but I also have other things implemented in sellout/dada, like `Vect`, `Decidable` relations, etc.

I have other branches with universes and universe polymorphism, but the lowest-implementation-cost way to present universe polymorphism is to add `Level : Type`, which requires dependent types.

### DIFF
--- a/dhall/src/Dhall/TypeCheck.hs
+++ b/dhall/src/Dhall/TypeCheck.hs
@@ -62,16 +62,12 @@ axiom Type = return Kind
 axiom Kind = return Sort
 axiom Sort = Left (TypeError Dhall.Context.empty (Const Sort) Untyped)
 
-rule :: Const -> Const -> Either () Const
-rule Type Type = return Type
-rule Kind Type = return Type
-rule Sort Type = return Type
-rule Kind Kind = return Kind
-rule Sort Kind = return Sort
-rule Sort Sort = return Sort
--- This forbids dependent types. If this ever changes, then the fast
--- path in the Let case of typeWithA will become unsound.
-rule _    _    = Left ()
+rule :: Const -> Const -> Const
+rule _ Type = Type
+rule Sort _ = Sort
+rule _ Sort = Sort
+rule Kind _ = Kind
+rule _ Kind = Kind
 
 {-| Type-check an expression and return the expression's type if type-checking
     succeeds or an error if type-checking fails
@@ -131,10 +127,7 @@ typeWithA tpa = loop
         kB <- case tB of
             Const k -> return k
             _       -> Left (TypeError ctx' e (InvalidOutputType _B))
-
-        case rule kA kB of
-            Left () -> Left (TypeError ctx e (NoDependentTypes _A _B))
-            Right k -> Right (Const k)
+        return (Const (rule kA kB))
     loop ctx e@(App f a         ) = do
         tf <- fmap Dhall.Core.normalize (loop ctx f)
         (x, _A, _B) <- case tf of
@@ -163,8 +156,6 @@ typeWithA tpa = loop
                     else Left (TypeError ctx e (AnnotMismatch a0 nf_A0 nf_A1))
             Nothing -> return ()
 
-        t <- loop ctx _A1
-
         let a1 = Dhall.Core.normalize a0
         let a2 = Dhall.Core.shift 1 (V x 0) a1
 
@@ -172,27 +163,9 @@ typeWithA tpa = loop
                 []       -> b0
                 l' : ls' -> Let (l' :| ls') b0
 
-        -- The catch-all branch directly implements the Dhall
-        -- specification as written; it is necessary to substitute in
-        -- types in order to get 'dependent let' behaviour and to
-        -- allow type synonyms (see #69). However, doing a full
-        -- substitution is slow if the value is large and used many
-        -- times. If the value being substitued in is a term (i.e.,
-        -- its type is a Type), then we can get a very significant
-        -- speed-up by doing the type-checking once at binding-time,
-        -- as opposed to doing it at every use site (see #412).
-        case Dhall.Core.normalize t of
-          Const Type -> do
-            let ctx' = fmap (Dhall.Core.shift 1 (V x 0)) (Dhall.Context.insert x (Dhall.Core.normalize _A1) ctx)
-            _B0 <- loop ctx' rest
-            let _B1 = Dhall.Core.subst (V x 0) a2 _B0
-            let _B2 = Dhall.Core.shift (-1) (V x 0) _B1
-            return _B2
-
-          _ -> do
-            let b1 = Dhall.Core.subst (V x 0) a2 rest
-            let b2 = Dhall.Core.shift (-1) (V x 0) b1
-            loop ctx b2
+        let b1 = Dhall.Core.subst (V x 0) a2 rest
+        let b2 = Dhall.Core.shift (-1) (V x 0) b1
+        loop ctx b2
 
     loop ctx e@(Annot x t       ) = do
         _ <- loop ctx t
@@ -262,18 +235,8 @@ typeWithA tpa = loop
         case tx of
             Bool -> return ()
             _    -> Left (TypeError ctx e (InvalidPredicate x tx))
-        ty  <- fmap Dhall.Core.normalize (loop ctx y )
-        tty <- fmap Dhall.Core.normalize (loop ctx ty)
-        case tty of
-            Const Type -> return ()
-            _          -> Left (TypeError ctx e (IfBranchMustBeTerm True y ty tty))
-
+        ty <- fmap Dhall.Core.normalize (loop ctx y)
         tz <- fmap Dhall.Core.normalize (loop ctx z)
-        ttz <- fmap Dhall.Core.normalize (loop ctx tz)
-        case ttz of
-            Const Type -> return ()
-            _          -> Left (TypeError ctx e (IfBranchMustBeTerm False z tz ttz))
-
         if Dhall.Core.judgmentallyEqual ty tz
             then return ()
             else Left (TypeError ctx e (IfBranchMismatch y z ty tz))
@@ -864,7 +827,6 @@ data TypeMessage s a
     | InvalidSome (Expr s a) (Expr s a) (Expr s a)
     | InvalidPredicate (Expr s a) (Expr s a)
     | IfBranchMismatch (Expr s a) (Expr s a) (Expr s a) (Expr s a)
-    | IfBranchMustBeTerm Bool (Expr s a) (Expr s a) (Expr s a)
     | InvalidField Text (Expr s a)
     | InvalidFieldType Text (Expr s a)
     | FieldAnnotationMismatch Text (Expr s a) Const Text (Expr s a) Const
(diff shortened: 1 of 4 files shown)


dhall/src/Dhall/TypeCheck.hs
```diff
@@ -163,36 +156,16 @@ typeWithA tpa = loop
...
-        -- specification as written; it is necessary to substitute in
-        -- types in order to get 'dependent let' behaviour and to
-        -- allow type synonyms (see #69). However, doing a full
-        -- substitution is slow if the value is large and used many
-        -- times. If the value being substitued in is a term (i.e.,
-        -- its type is a Type), then we can get a very significant
-        -- speed-up by doing the type-checking once at binding-time,
-        -- as opposed to doing it at every use site (see #412).
-        case Dhall.Core.normalize t of
-          Const Type -> do
```
sellout: I removed this “fast path”, but I’d love to have a test case that behaves incorrectly when this code exists.

Gabriel439:
I think the main blockers to dependent type support are:

* Ergonomic `Refl`
* Some clear standard for which judgmental equalities the language standard commits to supporting
* Ergonomic idioms for implementing tests using dependent types in Dhall (possibly using `Refl`)

If we had those then I think we could enable dependent types in the standard

dhall/src/Dhall/TypeCheck.hs
MonoidMusician: Let's assume we are in a context as follows: `Vec : Natural -> Type, Vec/nil : Vec 0`. Consider the (dependently-typed) expression `let n : Natural = 0 in Vec/nil : Vec n`. The fast path is equivalent to adding `n : Natural` to the context (and forgetting about its value). But `Vec : Natural -> Type, Vec/nil : Vec 0, n : Natural |- Vec/nil : Vec n` is not valid, since the value of `n` is not known. However, the slow path, of substitution, is valid (and obviously correct): `Vec/nil : Vec 0` is the result of the substitution, and it is clearly in the context, as desired.

The big picture is that, without dependent types, the only things that influence typechecking are types (and kinds (and sorts …)), not values. Values don't matter to typechecking, only to computation. However, with dependent types, that is obviously not the case: by definition, types are now allowed to depend on values.

I think there are ways to address the concern in the comment by sharing the work more efficiently though, but it requires a different representation. The type of a term has to be associated with the node that triggered it. It starts to look like a lazy graph. I might try that approach in the PS project when I return to it over the summer. Premature optimization though.

AndrasKovacs: @MonoidMusician Those kinds of performance questions are completely solved on the [nbe-elaboration](https://github.com/dhall-lang/dhall-haskell/tree/nbe-elaboration) branch, which I hope can be merged in the summer when I get some time.

AndrasKovacs:
@Gabriel439 `Refl` is easy to implement with Church encoding, I don't think there are ergonomic issues. I would be fine with an error message like "expected x, got y".

```
let Eq = \(A : Type) -> \(x : A) -> \(y : A) -> forall (P : A -> Type) -> P x -> P y
     Refl = \(A : Type) -> \(x : A) -> \(P : A -> Type) -> \(px : P x) -> px
```

My take on dependent types is that type-in-type would be the most appropriate for Dhall. Non-termination is not practically relevant, because we can already write effectively non-terminating programs. Logical consistency does not seem to be much needed now either; but if we want to check it, type-in-type is still the simplest. That's because normalizing an expression becomes a semidecision procedure for checking universe consistency, i.e. an expression with normal form is logically valid. This is extremely convenient compared to any leveling scheme; the reason it is not used elsewhere is the eye-watering overhead, but in Dhall we already normalize protected imports routinely.

Gabriel439:
@AndrasKovacs: I prefer not to compromise on totality because the market of non-total programming languages is already well-served and we inspire confidence in our users by going the extra mile in our commitment to safety, even if only symbolically.

Gabriel439:
@AndrasKovacs: Also, regarding `Refl`, the main ergonomic improvements I was thinking of were:

*   Not having to specify one side of the equality twice.  For example, when implementing `Refl` within the language, you end up having to specify three things that are equal, instead of two:

    ```haskell
    let Eq
        : ∀(a : Type) → ∀(x : a) → ∀(y : a) → Type
        = λ(A : Type) → λ(x : A) → λ(y : A) → ∀(P : A → Type) → P x → P y

    let Refl
        : ∀(a : Type) → ∀(x : a) → Eq a x x
        = λ(A : Type) → λ(x : A) → λ(P : A → Type) → λ(px : P x) → px

    in  Refl Natural 4 : Eq Natural 4 (2 + 2)
    ```

*   Not having to specify the type of the two expressions being compared (i.e. inferring it)

    For example: `Assert : 4 ≡ 2 + 2` instead of `Refl 4 : Eq 4 (2 + 2)`

    Note that I'm not sure if that is sound since I'm still new to dependently-typed programming

*   Possibly making the built-in `Refl` universe-polymorphic, if that is sound

    Mainly so that you can compare types, too, instead of just terms

*   Better error messages (as you noted):

    In other words, instead of this:

    ```

      …
    → …
    → …@…
      - 3
      + 4

    9:     Refl Natural 4 : Eq Natural 4 (2 + 1)
    ```

    Perhaps you get this:

    ```
    Error: Assertion failed

    You asserted that this expression:

    ↳ 4

    ... which has this normal form:

    ↳ 4

    ... is equivalent to this expression:

    ↳ 2 + 1

    ... which has the following different normal form:

    ↳ 3

    9:     Assert : 4 ≡ 2 + 1
    ```

AndrasKovacs:
@Gabriel439 fair enough on Refl.

Profpatsch:
In all this we shouldn’t forget that the main goal of dhall is to replace yaml. Not sure dependent types are a valid step in that direction, especially if the error messages degrade.

Gabriel439:
@Profpatsch: The main feature that this enables for the more "pragmatic" YAML use case is the ability to write tests within Dhall.  In other words, instead of this:

```haskell
{-
Transform a list by applying a function to each element

Examples:

./map Natural Bool Natural/even [ 2, 3, 5 ]
= [ True, False, False ]

./map Natural Bool Natural/even ([] : List Natural)
= [] : List Bool
-}
let map
    : ∀(a : Type) → ∀(b : Type) → (a → b) → List a → List b
    =   λ(a : Type)
      → λ(b : Type)
      → λ(f : a → b)
      → λ(xs : List a)
      → List/build
        b
        (   λ(list : Type)
          → λ(cons : b → list → list)
          → List/fold a xs list (λ(x : a) → cons (f x))
        )

in  map
```

... we would now be able to check the tests:

```haskell
{-
Transform a list by applying a function to each element
-}
let map
    : ∀(a : Type) → ∀(b : Type) → (a → b) → List a → List b
    =   λ(a : Type)
      → λ(b : Type)
      → λ(f : a → b)
      → λ(xs : List a)
      → List/build
        b
        (   λ(list : Type)
          → λ(cons : b → list → list)
          → List/fold a xs list (λ(x : a) → cons (f x))
        )

let _ = Assert : ./map Natural Bool Natural/even [ 2, 3, 5 ] ≡ [ True, False, False ]

let _ = Assert : ./map Natural Bool Natural/even ([] : List Natural) ≡ [] : List Bool

in  map
```

You'd even be able to add "property tests" that check that normal forms of functions are the same.  For example, this would be a valid test:

```haskell
let _ = λ(x : Natural) → Assert : x + 0 ≡ x
```

ocharles:
> You'd even be able to add "property tests" that check that normal forms of functions are the same.

Nit-picking, but these are formal proofs. I don't think it's a good idea to muddy the waters with terminology.

Gabriel439:
@ocharles: Yeah, I agree.  I just edited my comment

Gabriel439:
Also, the desire to write proofs is why we need to be careful, because users will want to write all sorts of proofs that will strain our judgmental equality and create pressure on us to make it more robust

An example proof that a user might want to write is the Dhall equivalent of `map id = id`:

```haskell
Assert : λ(a : Type) → map a a (λ(x : a) → x) ≡ λ(a : Type) → λ(xs : List a) → xs
```

... but that will fail due to our current poor support for build-fold fusion, which we would need to either fix or explicitly choose to abandon (and therefore not support proving that property).

sellout:
> Possibly making the built-in Refl universe-polymorphic

This relates to something that has come up on Slack, which is having void and unit `Kind`s, etc. We talked about having to annotate `{} : Type` and `<> : Kind`, like we do with `[] : List Natural`. But @MonoidMusician suggested having it _default_ to `Type`, so `{}` is a `Type`, but `{} : Kind` is a `Kind`. How do you feel about handling universe polymorphism like that?

Gabriel439:
@sellout: Yeah, that seems reasonable to me

sellout:
> `Refl` is easy to implement with Church encoding (@AndrasKovacs)

That’s the (non-dependent) approach I’ve been using for a while: https://github.com/sellout/caterwaul/blob/master/Leibniz/groupoid, although I don’t think anyone would accuse my definition of being ergonomic.

> Possibly making the built-in Refl universe-polymorphic, if that is sound

With general universe polymorphism, the Church-encoded `Refl` could also be universe-polymorphic (but it _would_ be an extra parameter).

> Not having to specify the type of the two expressions being compared (i.e. inferring it)

> Better error messages

I get wanting to minimize the redundancy and improve the error messages in order to make this more usable, but it seems like being able to implement that _in_ Dhall is evidence that we _can_ build it in. If so, it seems like getting it in incrementally would be good (i.e., let the people like me see which parts are most painful in practice, and then build in the minimum that helps the most). Granted, this proposal still requires a bunch of tests (especially negative ones), examples, and docs in a corresponding dhall-lang/dhall-lang change.

> especially if the error messages degrade (@Profpatsch)

I don’t think this would _degrade_ error messages at all (would it?), just that there are new things that are possible and those wouldn’t have great error messages (until we build in more stuff).

> Some clear standard for which judgmental equalities the language standard commits to supporting

This is what I see as the biggest hurdle. Even before dependent types, I’ve run into judgmental equality issues without an obvious solution. Trying to get to a point where the boundaries are fairly clear and explainable would be really helpful.

Profpatsch:
> I don’t think this would _degrade_ error messages at all (would it?), just that there are new things that are possible and those wouldn’t have great error messages

Yes, that’s the biggest problem for users that don’t want to use advanced features of a system. They accidentally stumble into advanced corner cases, error messages suddenly degrade, and the user is left guessing what they did wrong.

jneira:
Well, i am not a big fan of optionally switch languages features, like ghc and its LANGUAGE pragma, but maybe we could implement "something" to turn on dependent types, and later set it as default if we see is interesting.
Not sure if the added complexity of "something" would worth, though.

Gabriel439:
I don't think this would deteriorate error messages, especially if we added built-in support for `Refl`/`Assert`.  The error messages you would get for dependent type mismatches would be the same as the error messages you get for ordinary non-dependent type mismatches.

If you look at the example I gave with `Refl` above, even when it was encoded within the language the default error message was actually not too bad, since it highlighted that the crux of the problem was that `3` did not equal `4`:

```

  …
→ …
→ …@…
  - 3
  + 4

9:     Refl Natural 4 : Eq Natural 4 (2 + 1)
```

Gabriel439:
I'm going to close this as it is superseded by https://github.com/dhall-lang/dhall-haskell/pull/1164

PR #957, opened 2019-05-17
Implement universe support (non-cumulative, non-polymorphic).

sellout:
Tries to be syntax & binary compatible with previous versions (e.g., Universes
0–2 are serialized as "Type", "Kind", and "Sort"). However, `Sort` is no longer
valid syntax (it could never typecheck before anyway). Instead it must be
followed by a natural number, e.g.: `λ(object : Sort 2)`.

I’m just posting this for illustration. I think adding polymorphism support would require not using naturals for universe indices, so I wouldn’t want to add this then break it again.


Gabriel439:
I'll close this for now since the branch has gone stale

PR #962, opened 2019-05-22
dhall-json: fix yaml dependency.

quasicomputational:
In yaml-0.11, Text.Libyaml was split into the libyaml package. Hence,
as dhall-json imports that module, a libyaml dependency is needed.
Further, the lower bound on yaml needs to be bumped to 0.11, or else
that module can be provided by both a pre-0.11 yaml or libyaml.

### DIFF
--- a/dhall-json/dhall-json.cabal
+++ b/dhall-json/dhall-json.cabal
@@ -30,6 +30,10 @@ Source-Repository head
     Type: git
     Location: https://github.com/dhall-lang/dhall-haskell/tree/master/dhall-json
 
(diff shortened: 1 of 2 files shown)


quasicomputational:
If keeping a wider bound on `yaml` is preferred, the way to do it would be an automatic flag that switches between `yaml >= 0.5.0 && < 0.11` and `yaml >= 0.11 && < 0.12, libyaml`

quasicomputational:
Makes sense. Weird that Stack figures it out for the other configurations yet not that one, but code don't lie.

PR #964, opened 2019-05-22
dhall: fix the dhall-command benchmark.

quasicomputational:
Seems that Main.Default changed in #949, but this use-site wasn't
changed at the same time.


quasicomputational:
Maybe building benchmarks in CI would be an idea? This sort of bit-rot is liable to happen again if it's not machine-checked, IMO.

PR #970, opened 2019-05-25
Map-like Lambda converts to Map

singpolyma:
This allows translating Dhall expressions such as:

```haskell
\(env: < production | development >) -> merge {
	production = "thing1",
	development = "thing2"
} env
```

into the equivalent JSON or YAML map.

These kinds of expressions are a very natural way to encode in Dhall
config files which contain configurations for multiple environments, of
which only one is ever active.  YAML versions of these files tend to be
such a map, since YAML lacks the ability to express it as a function.

### DIFF
--- a/dhall-json/src/Dhall/JSON.hs
+++ b/dhall-json/src/Dhall/JSON.hs
@@ -178,6 +178,7 @@ import Control.Applicative (empty, (<|>))
 import Control.Monad (guard)
 import Control.Exception (Exception, throwIO)
 import Data.Aeson (Value(..), ToJSON(..))
+import Data.Maybe (isNothing)
 import Data.Monoid ((<>), mempty)
 import Data.Text (Text)
 import Dhall.Core (Expr)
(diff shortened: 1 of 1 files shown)


Gabriel439:
@singpolyma: What if you stored the record of handlers in a separate file:

```haskell
\(env: < production | development >) -> merge ./handlers.dhall env
```

```haskell
-- ./handlers.dhall
{
	production = "thing1",
	development = "thing2"
}
```

... and then translated the record of handlers?  That would generate the same output

singpolyma:
The example record of handlers here is just a simplified example. In practise you likely have shared keys between environments and just some parts using merge on the env lower down.

Profpatsch:
Can you give a full example?

Gabriel439:
@singpolyma: I still think it would be more idiomatic to encode the handlers directly.  They already have the right shape and don't require special treatment.  The main reason I suggest this is that this seems like a surprising feature to convert a function to a record.

A secondary reason for this is that at some point I'd like to be able to implement `dhall-to-json`/`dhall-to-yaml` in pure Dhall (see: https://github.com/dhall-lang/dhall-lang/compare/gabriel/json_pure_dhall) where this sort of trick wouldn't work.

singpolyma:
> Can you give a full example?

\(env : < production | development | test >) -> {
    driver = "sqlite3",
    timeout = 5000,
    database = merge {
        production = "db/prod.sqlite3",
        development = "db/dev.sqlite3",
        test = "db/test.sqlite3"
    }
}‎

Is a more realistic, but still simple, example.

Profpatsch:
No, I mean what this should convert to.

Gabriel439:
I'll close this out for now since I think I still prefer the solution of preserving the original record of handlers directly.

PR #974, opened 2019-05-28
Fix LSP diagnostics to trailing whitespace

EggBaconAndSpam:
Dhall's current parser includes trailing whitespace in source range
annotations. Until now, we simply passed these "loose" ranges on to the
LSP client when generating diagnostic messages.

Previously (assuming funcTion is misspelled):
```
 funcTion argument
 ~~~~~~~~~

```
Now:
```
  funcTion argument
  ~~~~~~~~
```

### DIFF
--- a/dhall-lsp-server/src/Backend/Dhall/Diagnostics.hs
+++ b/dhall-lsp-server/src/Backend/Dhall/Diagnostics.hs
@@ -26,6 +26,8 @@ import Dhall.Import(Imported(..), Cycle(..), ReferentiallyOpaque(..),
 
 import Data.Text (Text)
 import qualified Data.Text as T
+import Data.List.NonEmpty (NonEmpty (..))
+import qualified Data.List.NonEmpty as NonEmpty
 
 
 import qualified Text.Megaparsec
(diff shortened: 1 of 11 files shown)


dhall-lsp-server/src/Backend/Dhall/Diagnostics.hs
```diff
@@ -188,3 +192,40 @@ errorFancyLength :: Text.Megaparsec.ShowErrorComponent e => Text.Megaparsec.Erro
...
   _             -> 1
+
+-- sanitise range to exclude surrounding whitespace
+-- makes sure that the resulting range is well-formed
+sanitiseRange :: Range -> Text -> Range
+sanitiseRange (Range l r) text = Range l (max l r')
+  where r' = trimEndPosition r text
+
+-- Variants of T.lines and T.unlines that are inverses of one another.
+lines' :: Text -> [Text]
```
Gabriel439: Perhaps restrict the type to `Text -> NonEmpty Text`.  My understanding is that `Data.Text.split` always returns a non-empty list even if the type does not convey that information.  You could implement it as:

```haskell
lines' :: Text -> NonEmpty Text
lines' text =
    case Data.Text.split (== '\n') text of
        []     -> "" :| []
        l : ls -> l  :| ls
```

The main reason I suggest this is so that functions like `offsetToPosition` don't need to handle the case where the list is empty since you've upstreamed the logic to handle that into the `lines'` function

EggBaconAndSpam: Thanks for spotting this! I changed the definition to the one you proposed.

Gabriel439:
Thank you!  Great work 🙂

bubba:
Hopefully the first of many PRs this summer!

PR #977, opened 2019-05-29
Use json-to-dhall to implement yaml-to-dhall

robbiemcmichael:
I was really happy to see @antislava implemented `json-to-dhall` recently in https://github.com/dhall-lang/dhall-haskell/pull/884. This refactors that code into the `dhall-json` library so that we can easily implement `yaml-to-dhall` as well.

It could probably be cleaned up a bit more, happy to take suggestions of further refactoring as well.

### DIFF
--- a/dhall-json/src/Dhall/JSONToDhall.hs
+++ b/dhall-json/src/Dhall/JSONToDhall.hs
@@ -6,10 +6,148 @@
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
+{-| Convert JSON data to Dhall given a Dhall /type/ expression necessary to make the translation unambiguous.
+
+    Reasonable requirements for conversion are:
+
+    1. The Dhall type expression @/t/@ passed as an argument to @json-to-dhall@ should be a valid type of the resulting Dhall expression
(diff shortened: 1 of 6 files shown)


dhall-json/src/Dhall/JSONToDhall.hs
```diff
@@ -0,0 +1,392 @@
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE RecordWildCards     #-}
+{-# LANGUAGE PatternGuards       #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE FlexibleInstances   #-}
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Dhall.JSONToDhall where
```
Gabriel439: Since this module is publicly visible it's probably worth adding a module header to document what it is used for and also an explicit export list so that we can safely change internal utility functions in this module without breaking the public API

robbiemcmichael: Good point. I've added the explicit exports and moved the docs over to the public module.

robbiemcmichael:
@Gabriel439 Looks like I don't have the ability to merge as I don't have write access. Could you please merge this PR if you're happy with the changes?

Gabriel439:
@robbiemcmichael: Yeah, no problem!  Also, great job on documentation 🙂 

Double-check your inbox and/or spam folder because I believe you should have an invite e-mail to become a collaborator, which would grant you write access for future contributions

PR #978, opened 2019-05-31
Bump up lower bound of aeson-pretty

jneira:
`dhall-to-json`uses the field `Data.Aeson.Encode.Pretty.confTrailingNewline`, introduced in `aeson-pretty-0.8.5`

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,7 +21,7 @@ environment:
     - STACK_YAML: stack-lts-6.yaml
 
 install:
-  - choco install -dv -y haskell-stack --version %STACK_VERSION%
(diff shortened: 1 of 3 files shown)


jneira:
appveyor failed forlts-6 with a strange error about downloading stack with chocolatey: i am taking a look

jneira:
I've cleaned up the appveyor script, adding traces to `choco install` and it worked again.
I've changed default `stack.yaml` to use the lastest lts.

Now, let's see if removing traces doesnt make it fail again :laughing:

Gabriel439:
@jneira: Thank you! :slightly_smiling_face:

PR #980, opened 2019-06-01
Add new `--cache` flag to `dhall freeze`

Gabriel439:
Part of https://github.com/dhall-lang/dhall-lang/issues/563

This flag freezes imports in the same way as the Prelude by providing a
fallback unprotected import without an integrity check.  The primary use
case for this is caching imports with a graceful fallback, which is why
the flag is named `--cache`


dhall/src/Dhall/Core.hs
```diff
@@ -2141,3 +2152,49 @@ prettyURIComponent text
...
+transformOf :: ASetter a b a b -> (b -> b) -> a -> b
+transformOf l f = go
+  where
+    go = f . over l go
+{-# INLINE transformOf #-}
+
+{-| Convenience utility identical to @"Control.Lens".`Control.Lens.rewriteMOf`@
+    re-exported for convenience in order to minimize dependencies on @lens@
+-}
+rewriteMOf
```
Gabriel439: cc: @ocharles 

I re-exported these utilities along the way, which you might find useful

PR #982, opened 2019-06-03
dhall-lsp-server: Rework diagnostics backend + detailed error messages on hover

EggBaconAndSpam:
Reworks the diagnostics backend completely and adds the "detailed error explanations on hover" functionality. The code for the added feature is very much quick-and-dirty and will be reworked later.

Sorry for bundling the rewrite and new feature into a single pull request. Won't happen again ;)

For the future: should I clean up the commit history before creating a pull request?


f-f:
@EggBaconAndSpam no worries about cleaning up commit history, we squash-merge so a every PR ends up as a single commit in master

EggBaconAndSpam:
Ah, that makes sense. Thanks!

f-f:
You're welcome 🙂

dhall-lsp-server/src/LSP/Dispatcher.hs
```diff
@@ -109,10 +117,37 @@ dispatcher lf inp = do
           formattedDocument <- formatDocument uri tabSize insertSpaces
           publish req RspDocumentFormatting formattedDocument
 
+      ReqHover req -> do
+          liftIO $ LSP.Utility.logm "****** reactor: processing ReqHover"
+          let (J.Position line col) = req ^. (J.params . J.position)
+              doc = req ^. (J.params . J.textDocument . J.uri)
+              Just fileName = J.uriToFilePath doc
```
Gabriel439: I try to avoid non-exhaustive pattern match.  If you need a quick solution, you can do:

```haskell
fileName <- case J.uriToFilePath doc of
    Nothing       -> fail "A terse error message is fine for now"
    Just fileName -> return fileName
```

dhall-lsp-server/src/LSP/Dispatcher.hs
```diff
@@ -109,10 +117,37 @@ dispatcher lf inp = do
...
+              isHovered :: Diagnosis -> Bool
+              isHovered (Diagnosis _ Nothing _) = False
+              isHovered (Diagnosis _ (Just (Range left right)) _)
+                = left <= (line,col) && (line,col) <= right
+              hover = case filter isHovered explanations of
+                [] -> Nothing
+                (diag : _) -> hoverFromDiagnosis diag
+              -- J.encode
+          publish req RspHover hover
+          -- "command:dhallLSPClient.explain?" ++
```
Gabriel439: Did you mean to leave this comment here?

dhall-lsp-server/src/LSP/Handlers/Diagnostics.hs
```diff
@@ -23,34 +22,56 @@ import Control.Monad.Reader (ReaderT)
...
+  
 -- | Analyze the file and send any diagnostics to the client in a
 -- "textDocument/publishDiagnostics" notification
-
-sendDiagnostics :: J.Uri -> Maybe Int ->  ReaderT (LSP.Core.LspFuncs ()) IO ()
+sendDiagnostics :: LSP.Uri -> Maybe Int ->  ReaderT (LSP.Core.LspFuncs ()) IO ()
 sendDiagnostics fileUri version = do
   let
-   filePath = maybe (error "can't convert uri to file path") id $ J.uriToFilePath fileUri -- !FIXME: handle non-file uris
+   filePath = maybe (error "can't convert uri to file path") id $ LSP.uriToFilePath fileUri -- !FIXME: handle non-file uris
```
Gabriel439: I also try to avoid the use of `error` and prefer using `fail` (specialized to `ReaderT … IO` in this case) instead:

```haskell
filePath <- case LSP.uriToFilePath of
    Nothing       -> fail "can't convert uri to fail path"
    Just filePath -> return filePath
```

The main reason why is that `error` is sensitive to evaluation order, whereas `fail` is not (at least, for `IO`).

EggBaconAndSpam:
Thanks for the comments! Turns out dhall-lsp-server was building without -Wall all along... ;)

I'll leave a proper fix to DhallExceptions for later, as that will require some more reading of Dhall's source code. The current version replicates PanAeon's code in functionality, so at least we aren't losing anything ;)

dhall-lsp-server/src/LSP/Handlers/DocumentFormatting.hs
```diff
@@ -6,23 +6,19 @@ import qualified Backend.Dhall.Formatting as Formatting
...
-import qualified Data.SortedList
 import Control.Monad.Trans (lift)
 import Control.Monad.Reader (ReaderT)
 
 -- TODO: implement tabSize and spaces/tabs options
 -- * Note: any formatting errors would be swallowed. I think this is fine in this case, but generally we'd like to send user a notification
 -- (e.g. the error occurred in the formatter itself, and user requests format constantly and nothing happens)
 formatDocument :: J.Uri -> Int -> Bool -> ReaderT (LSP.Core.LspFuncs ()) IO (J.List J.TextEdit)
-formatDocument fileUri tabSize insertSpaces = do
+formatDocument fileUri _ {-tabSize-} _ {-insertSpaces-} = do
```
f-f: If you commented this out to avoid warnings about unused bindings, then you can just prefix names with underscores: 
```suggestion
formatDocument fileUri _tabSize _insertSpaces = do
```

EggBaconAndSpam: Didn't know that! Thanks!

EggBaconAndSpam:
Should I merge this myself?

Gabriel439:
@EggBaconAndSpam: Yes!  Once you have an approval, you should feel free to merge at any time.

The general rule of thumb of contributions is to bias towards merging them in.  If there are any issues or concerns that come up after merging then we can always open a new pull request to address them.

EggBaconAndSpam:
Here wo go, pressing the magic button!

PR #987, opened 2019-06-05
Implement new Unicode braced escape sequence

Gabriel439:
... as standardized in https://github.com/dhall-lang/dhall-lang/pull/580


dhall/src/Dhall/Parser/Expression.hs
```diff
@@ -471,11 +473,20 @@ completeExpression embedded = completeExpression_
 
                 unicode = do
                     _  <- Text.Parser.Char.char 'u';
-                    n0 <- hexNumber
-                    n1 <- hexNumber
-                    n2 <- hexNumber
-                    n3 <- hexNumber
-                    let n = ((n0 * 16 + n1) * 16 + n2) * 16 + n3
+
+                    let toNumber = Data.List.foldl' (\x y -> x * 16 + y) 0
```
EggBaconAndSpam: toNumber returns an unbounded integer, which could be a problem since we blindly cast this into a Char later. Note that the maximum valid unicode codepoint is 10FFFF.

GHC's Char implementation seems to agree: trying to parse the dhall expression "\u{110000}" results in "Prelude.chr: bad argument: 1114112".

PR #989, opened 2019-06-05
[dhall-json] Isolate yaml usage behind `Dhall.Yaml`

jneira:
The goal is prepare the package for:
* The possible replacement of `yaml` with `HsYaml` (see #744)
  * Although likely the functions will have to change too (now the make conversions between aeson values and yaml)
* Make easier adding support for other compilers as `eta` or `ghcjs` (see https://github.com/eta-lang/dhall-haskell/tree/eta-support and https://github.com/eta-lang/dhall-haskell/tree/ghcjs-support


Gabriel439:
@jneira: Great work!  Thank you 🙂

PR #993, opened 2019-06-07
Another step in isolate yaml and prepare for HsYaml replacement

jneira:
Hi, this is a continuation of #989:
* It makes json and yaml modules more symmetric (although not quite):
  * dhall ones includes some io logic and options/error handling extracted from the executables
* It hides the fact we are converting from/to yaml using the json conversions. This likely will change if we finally use HsYaml (that has its own `FromYaml` type class)

However i somewhat dont like the signatures of main functions:

```haskell
dhallFromYaml :: Options -> ByteString -> IO Text

dhallToYaml :: Options -> Text  -> Text -> IO ByteString
```

They are generic but not very safe (basically they are `any -> any` and throw exceptions).
When we will use HsYaml we could replace them or add another ones using `FromYaml`.


Gabriel439:
@jneira: Thanks again! 🙂

PR #995, opened 2019-06-08
Switch to native GitHub highlighting for .dhall

f-f:
See https://github.com/dhall-lang/dhall-lang/pull/583


Gabriel439:
Thank you! 🙂

PR #1001, opened 2019-06-10
Fix linting of unused let bindings

EggBaconAndSpam:
- ### Find unused bindings inside nested lets
  Fixes cases like `let a = 0 let b = 0 in a`, that didn't get correctly linted (whereas `let a = 
0 in let b = 0 in a` would).
- ### Fix `freeIn` to correctly handle de Briujn indices 
  Fixes Core.Dhall.freeIn -- previously `freeIn "x@0" "x@1"`; as a result, the linter did not notice cases like `let a = 0 in let a = 0 in a@1`.
- ### Fix `removeUnusedBindings` to update de Bruijn indices
  Fixes `removeUnusedBindings` to correctly update any affected de Bruijn indices.

### DIFF
--- a/dhall/src/Dhall/Lint.hs
+++ b/dhall/src/Dhall/Lint.hs
@@ -25,20 +25,25 @@ import qualified Dhall.Optics
     * switches legacy @List@-like @Optional@ literals to use @Some@ / @None@ instead with 'optionalLitToSomeNone'
 -}
 lint :: Expr s Import -> Expr t Import
-lint =
-  Dhall.Optics.rewriteOf
-    subExpressions
-    ( \e -> removeLetInLet e
-        `mplus` optionalLitToSomeNone e
-    )
-    . Dhall.Optics.rewriteOf subExpressions removeUnusedBindings
-    . Dhall.Optics.rewriteOf subExpressions unfoldNestedLets
-    . Dhall.Core.denote
+lint = postproc . linting . preproc
+  where
+    -- pre-processing: remove Note constructors and unfold Let blocks
+    preproc = Dhall.Optics.rewriteOf subExpressions unfoldNestedLets . Dhall.Core.denote
+    -- main linting step: remove unused let bindings and update optional syntax
+    linting = Dhall.Optics.rewriteOf subExpressions
+      (\e -> removeUnusedBindings e `mplus` optionalLitToSomeNone e)
+    -- post-processing: fold nested Lets into Let blocks
+    postproc = Dhall.Optics.rewriteOf subExpressions removeLetInLet
 
+-- Merge nested Let blocks. Only finds immediately nested blocks -- use
+-- Dhall.Core.denote to make sure there aren't any Note constructors in the way!
 removeLetInLet :: Eq a => Expr s a -> Maybe (Expr s a)
 removeLetInLet (Let a (Let b c)) = Just (Let (a <> b) c)
 removeLetInLet _ = Nothing
 
+-- Remove unused Let bindings. Only considers the first variable in each Let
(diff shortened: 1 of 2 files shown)


dhall/src/Dhall/Lint.hs
```diff
@@ -28,17 +28,18 @@ lint :: Expr s Import -> Expr t Import
 lint =
   Dhall.Optics.rewriteOf
     subExpressions
-    ( \e ->
-                removeLetInLet e
-        `mplus` removeUnusedBindings e
+    ( \e -> removeLetInLet e
         `mplus` optionalLitToSomeNone e
     )
+    . Dhall.Optics.rewriteOf subExpressions removeUnusedBindings
```
Gabriel439: Are you sure that you want to compose `rewriteOf`s instead of using `mplus`?  The reason I ask is that I believe using `mplus` ensures that we don't miss newly-exposed opportunities for rewrites.

For example, consider the generic case of:

```haskell
rewriteOf optic (\e -> f e `mplus` g e)
```

... versus:

```haskell
rewriteOf optic f . rewriteOf optic g
```

The latter approach means that if the rewrite named `f` exposes new opportunities for the rewrite named `g`, they will be missed, whereas the former approach using `mplus` would still work because it will keep trying either `f` or `g`.

Or, to use regular expression notation, the former approach using `mplus` is sort of like:

```
(f + g)*
```

... whereas the latter approach using function composition of `rewriteOf` would be like:

```
f*g*
```

... which is not equivalent.

ocharles: Yea, this is not quite the intention I had when I rewrote the module like this

Gabriel439: Oh wait, I see why they are done sequentially.  It's because `unfoldNestedLets` is the opposite of `removeLetInLet`

Is there any way to do this that doesn't require the `unfoldNestedLets` preprocessing pass?

EggBaconAndSpam: Yes. In fact I tried a direct fix first, i.e. modifying `removeUnusedBindings` to walk through all binders in a let block. But that turned out to get really quite complicated... 

I think this is by far the cleanest solution (and time complexity wise the unfolding/folding is no more than a constant factor). One thing that I might try is rewriting `lint` in a more explicit fashion, with comments at each step &ndash; the point-free style is a bit hard to read.

dhall/src/Dhall/Lint.hs
```diff
@@ -3,9 +3,6 @@
 module Dhall.Lint
     ( -- * Lint
       lint
-    , removeLetInLet
-    , removeUnusedBindings
-    , optionalLitToSomeNone
```
EggBaconAndSpam: Given as the semantics of the remove* rules are a bit tricky currently I don't feel too comfortable exporting them... Shouldn't lint be enough?

ocharles: Right, originally I intended them to be exported so we could test independent passes in unit tests. E.g., does `optionalLitToSomeNone` do the right thing? We can test that by just using `lint`, but it's harder to isolate what's gone wrong. However, I didn't really anticipate the interaction between the phases!

EggBaconAndSpam: Good point about testing. In my opinion that shouldn't be a problem at the moment, but we might want to revisit that if the linter ever gets more complicated.

dhall/src/Dhall/Lint.hs
Gabriel439: Alright, then we'll keep it the way you have it

PR #1003, opened 2019-06-11
dhall-lsp-server: Implement linter diagnostics and "Lint and Format" command

EggBaconAndSpam:
- The diagnostics now include linter hints. For example:
  ```
  let a = 0 in let b = a in b
               ~
  ```
  Here we mark the `let` after the superfluous `in` (that's the best we can do). 
- Does not implement linting of old-style optionals since those are due to be removed, see #1002.
- Implements a "Lint and format" command (needs a correspondingly patched `vscode-dhall-lsp-server`)

### DIFF
--- a/dhall-lsp-server/src/Dhall/LSP/Backend/Linting.hs
+++ b/dhall-lsp-server/src/Dhall/LSP/Backend/Linting.hs
@@ -23,8 +23,8 @@ data Suggestion = Suggestion {
 
 -- Diagnose nested Let blocks.
 diagLetInLet :: Expr Src a -> [Suggestion]
-diagLetInLet (Note src (Let _ (Note _ (Let _ _)))) =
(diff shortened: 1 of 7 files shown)


dhall-lsp-server/src/Dhall/LSP/Backend/Linting.hs
```diff
@@ -0,0 +1,60 @@
...
+    Suggestion (rangeFromDhall src) ("Unused let binding '" <> var <> "'"))
+  (unusedBindings e)
+diagUnusedBinding _ = []
+
+-- | Given an dhall expression suggest all the possible improvements that would
+--   be made by the linter.
+suggest :: Text -> [Suggestion]
+suggest txt = case exprFromText "" txt of
+  Right expr -> concat [ diagLetInLet e ++ diagUnusedBinding e
+                       | e <- universeOf subExpressions expr ]
```
Gabriel439: This is very clever! 🙂

PR #1005, opened 2019-06-12
dhall-lsp-server: Towards implementing a "convert to JSON" command

EggBaconAndSpam:
At the moment the VSCode plugin contains a prototype of
a dhall-to-json (and -to-yaml) preview feature. We should ultimately
move that kind of functionality to dhall-lsp-server to 1) minimise the
amount of VSCode specific code and 2) avoid having to work with and fix
bugs in typescript code ;)

This takes us most of the way there; what is missing is the
ability to create files (via a WorkspaceEditRequest), which is not yet
implemented by haskell-lsp-types. I will wait with exposing the
corresponding command (and removing the existing preview code) in
vscode-dhall-lsp-server until this is fixed upstream.

Also contains some amount of refactoring of the Handlers code, for good measure.

### DIFF
--- a/dhall-lsp-server/src/Dhall/LSP/Handlers/Command.hs
+++ b/dhall-lsp-server/src/Dhall/LSP/Handlers/Command.hs
@@ -20,10 +20,12 @@ import qualified Data.Text as Text
 
 executeCommandHandler :: LSP.LspFuncs () -> J.ExecuteCommandRequest -> IO ()
 executeCommandHandler lsp request
-  | command == "dhall.server.lint" =
-    executeLintAndFormat lsp (parseUriArgument request)
-  | command == "dhall.server.toJSON" =
-    executeDhallToJSON lsp (parseUriArgument request)
+  | command == "dhall.server.lint" = case parseUriArgument request of
+      Right uri -> executeLintAndFormat lsp uri
+      Left msg -> LSP.logs msg
+  | command == "dhall.server.toJSON" = case parseUriArgument request of
+      Right uri -> executeDhallToJSON lsp uri
+      Left msg -> LSP.logs msg
   | otherwise = LSP.logs
     ("LSP Handler: asked to execute unknown command: " ++ show command)
   where command = request ^. J.params . J.command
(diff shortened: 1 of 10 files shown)


dhall-lsp-server/src/Dhall/LSP/Handlers/Command.hs
```diff
@@ -0,0 +1,88 @@
...
+                   $ J.ApplyWorkspaceEditParams
+                   $ J.WorkspaceEdit (Just (singleton uri edit)) Nothing
+    _ -> LSP.logs "LSP Handler: linting failed"
+
+-- Helper that appends a suffix to a uri. Fails if the uri does not represent a
+-- file path.
+appendSuffixToUri :: J.Uri -> Text -> J.Uri
+appendSuffixToUri uri suffix = case J.uriToFilePath uri of
+  Just path -> J.filePathToUri $ replaceExtension path (show suffix)
+  Nothing -> error $ "failed to append suffix to uri " ++ show uri
```
Gabriel439: I recommend returning `Either String J.Uri` instead of `error`

dhall-lsp-server/src/Dhall/LSP/Handlers/Command.hs
```diff
@@ -0,0 +1,88 @@
...
+                     ++ " because it's not a valid file path"
+
+parseUriArgument :: J.ExecuteCommandRequest -> J.Uri
+parseUriArgument request = case request ^. J.params . J.arguments of
+  Just (J.List (x : _)) -> case J.fromJSON x of
+    J.Success uri -> uri
+    _             -> error $ "unable to parse uri argument to " <> show
```
Gabriel439: Same thing here: I try to avoid using `error` when possible

dhall-lsp-server/src/Dhall/LSP/Handlers/Command.hs
EggBaconAndSpam: That's much better! I was being a bit sloppy there.

Gabriel439:
@EggBaconAndSpam: Thank you! 🙂

PR #1007, opened 2019-06-13
Add `dhall-json` support for weakly-typed JSON values

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/409

This matches the standard representation for arbitrary JSON proposed in
https://github.com/dhall-lang/dhall-lang/pull/586

This change supports conversions in both directions (from `JSON/Type` to
JSON and back)


dhall-json/src/Dhall/JSON.hs
```diff
@@ -149,6 +150,24 @@
...
 > }
 
+    You can also translate Dhall expressions encoding weakly-typed JSON
+    (see: <https://prelude.dhall-lang.org/JSON/Type>):
+
+> $ cat ./example.dhall
+> let JSON = https://prelude.dhall-lang.org/JSON/package.dhall
+> 
+> in  JSON.object
+>     [ { mapKey = "foo", mapValue = JSON.null }
```
ocharles: This ship has probably sailed now, but it's a little strange to me that we call it an `object` in the constructor, but the elements talk about a `map`. `object` with `propertyKey`/`propertyValue` might have been better, as objects have properties, rather than maps (they *are* maps).

Gabriel439: @ocharles: The names for the constructors are based on the official grammar for JSON: https://www.json.org/

Gabriel439: Sorry, I misunderstood your comment.  The reason for still using `mapKey`/`mapValue` is so that once we support the `toMap` keyword, we can do this:

```dhall
let JSON = https://prelude.dhall-lang.org/JSON/package.dhall

in  JSON.object (toMap { foo = JSON.number 1, bar = JSON.number 2 })
```

Issue #387, opened 2018-05-14
Make type of `extract` richer.


sellout:
Currently, if we have an error anywhere in our many `Interpret` instances, we get the same message. I propose that
```haskell
extract :: Expr Src X -> Maybe a
```
be changed to something like
```haskell
extract :: Expr Src X -> Validation (NonEmpty InterpretError) a
```
in order to get more precision in the messages. I’m not yet sure exactly what `InterpretError` looks like, but I imagine it would have cases like `UnexpectedExpression` and `UnsupportedExpression` for handling things like
```haskell
nonEmptyTy :: Type a -> Type (NonEmpty a)
nonEmptyTy Type{..} =
  Type
  (\case
      e@(D.ListLit _ elems) ->
        maybe (Failure . pure $ UnsupportedExpression e D.List "list must not be empty")
              Success
              (nonEmpty . toList =<< traverse extract elems)
      e -> Failure . pure $ UnexpectedExpression e D.List)
  D.List
```

If this seems acceptable, I could take a swipe at it.

ocharles:
I've also been thinking that we should some how maintain the 'path' we're validating - a bit like how aeson can see exactly where in a JSON document the parse failed

sellout:
@ocharles Agreed. I’d be happy with either my suggestion or yours (or both 😄).

Gabriel439:
Yeah, if it's a simple change.  Keep in mind that if a `Type` is implemented correctly then `extract` should never fail, so I wouldn't go overboard in improving the error message here

sellout:
@Gabriel439 Yeah, the problem is finding programming bugs.

joneshf:
An example of hitting this issue: I had a type that used to be a record:
```Dhall
{ executables :
    List { executableName : Text, executableDirectory : Text }
, manifest :
    < Cabal : {} | Hpack : {} >
, name :
    Text
, sourceDirectory :
    Text
, tests :
    List { suite : Text, testDirectory : Text }
, version :
    Text
}
```
Using `auto` and the following Haskell module:
```Haskell
module Shake.Package where

data Executable
  = Executable
    { executableName      :: String
    , executableDirectory :: FilePath
    }
  deriving (Generic)

instance Interpret Executable

data Manifest
  = Cabal
  | Hpack
  deriving (Generic)

instance Interpret Manifest

data Package
  = Package
    { executables     :: [Executable]
    , manifest        :: Manifest
    , name            :: String
[...]

PR #1011, opened 2019-06-17
[WIP] make the type of extract richer

divarvel:
See #387 for the whole context. The main idea is to make `extract` return a more detailed errors (currently, it returns a `Maybe`).

Following discussion with @ocharles, and based on @sellout's comments in the issue, I went with a `Validation (NonEmpty ExtractError)`, where `ExtractError` is either `InvalidType` or a freeform text for user-supplied errors. Reading the tests, the example provided by @sellout seems at odds with the contact that `extract` should only fail on type errors.

For now it does not track line number or anything else, but thanks to the applicative instance, it accumulates errors for lists and records. Instead of providing a monad instance through a newtype, I added a `validate` function which is essentially `>>=`.

I've added dependencies to `either` (for `Validation`) and to `semigroupoids` (for `Data.Functor.Alt`, as `NonEmptyList a` only forms a semigroup).

# ToDo

- [x] add a proper error message for the errors that are not type mismatches
- [x] make tests pass (there is an error message mismatch)


dhall/src/Dhall.hs
```diff
@@ -96,13 +99,16 @@ module Dhall
...
     ) where
 
-import Control.Applicative (empty, liftA2, (<|>), Alternative)
+import Control.Applicative (empty, liftA2, Alternative)
 import Control.Exception (Exception)
 import Control.Monad.Trans.State.Strict
 import Control.Monad (guard)
 import Data.Coerce (coerce)
+import Data.Either.Validation (Validation(..))
+import Data.Functor.Alt ((<!>))
```
Gabriel439: I believe you can use `<|>` instead of `<!>`, that way you don't need to directly depend on `semigroupoids`

dhall/src/Dhall.hs
```diff
@@ -157,12 +165,47 @@ import qualified Dhall.Util
...
 -}
+type Extractor s a = Validation (ExtractErrors s a)
+
+newtype ExtractErrors s a =
+    ExtractErrors
+  { getErrors :: NonEmpty (ExtractError s a)
+  } deriving Semigroup
+
+typeError :: Expr s a -> Expr s a -> Extractor s a b
+typeError expected actual = Failure . ExtractErrors . (:| []) . TypeMismatch $ InvalidType expected actual
```
Gabriel439:
```suggestion
typeError expected actual = Failure . ExtractErrors . pure . TypeMismatch $ InvalidType expected actual
```

dhall/src/Dhall.hs
```diff
@@ -157,12 +165,47 @@ import qualified Dhall.Util
...
+
+extractError :: Text -> Extractor s a b
+extractError = Failure . ExtractErrors . (:| []) . ExtractError
```
Gabriel439:
```suggestion
extractError = Failure . ExtractErrors . pure . ExtractError
```

dhall/src/Dhall.hs
```diff
@@ -157,12 +165,47 @@ import qualified Dhall.Util
...
+
+-- | opt-in monadic composition for `Extractor`. This is done this way to
+-- keep applicative behaviour on @ExtractErrors@
+validate :: Extractor s a b -> (b -> Extractor s a c) -> Extractor s a c
```
Gabriel439: If we don't use `validate` internally I would omit it from the API, especially given the fact that we expose the `Extractor` type's internals

dhall/src/Dhall.hs
```diff
@@ -157,12 +165,47 @@ import qualified Dhall.Util
...
+validate (Success a) f = f a
+validate (Failure es) _ = Failure es
+
+instance (Pretty s, Pretty a, Typeable s, Typeable a) => Show (ExtractErrors s a) where
+    show (ExtractErrors es)  = unlines . Data.List.NonEmpty.toList . fmap show $ es
```
Gabriel439: Two suggestions for formatting the error output:

* If there is more than one error, prefix all of the errors with a brief message saying that there multiple errors
* Separate the errors visually (i.e. perhaps add a blank line between them if there's not one already)

dhall/src/Dhall.hs
```diff
@@ -362,9 +405,9 @@ inputWithSettings settings (Type {..}) txt = do
     let normExpr = Dhall.Core.normalizeWith (view normalizer settings) expr'
 
     case extract normExpr  of
-        Just x  -> return x
-        Nothing -> Control.Exception.throwIO
-                     (InvalidType expected expr')
+        Success x  -> return x
+        Failure e -> Control.Exception.throwIO
+                     e
```
Gabriel439: The `e` no longer needs to be on its own line

dhall/src/Dhall.hs
divarvel: I've replaced `<|>` with `<!>` because it requires a `Monoid`, whereas `NonEmpty a` (and `Validation (NonEmpty a) b`) only have a `Semigroup`. If the dependency on semigroupoids is a problem, I can just reimplement `<!>` instead of pulling the library.

dhall/src/Dhall.hs
divarvel: :+1: , i'll do that

dhall/src/Dhall.hs
divarvel: Indeed, that's fixed

dhall/src/Dhall.hs
Gabriel439: Oh, alright.  Then you can leave the `(<!>)`.  Minimizing the constraint is a good reason to use it

dhall/src/Dhall.hs
Gabriel439: I think it's fine if people use this for non-type-errors.

I think my main issue with this function is that it's basically a `Monad` instance but without the instance.  I think if we firmly believe that the user will need `Monad` operations then we should go all the way and provide a `newtype` that exposes a `Monad` instance rather than a named bind function, because the latter will not work with `do` notation, monad transformers, or `Control.Monad` utilities.

divarvel: Ok, I'll make a monad instance with a newtype then.

divarvel: Now that I think of it, `Extractor s a` is just a type alias for `Validation (NonEmpty (ExtractError s a)) b`. I can add a new type alias for `Either (NonEmpty (ExtractError s a)) b` (named eg `MonadicExtractError`), and maybe provide a pair of  `from/to` functions (aliases of `eitherToValidation` and `validationToEither`). Or we just let implementors use `Data.Either.Validation` functions themselves

divarvel:
As suggested, i've special-cased the error message for a single error (so that it doesn't change current behaviour). I've added a small message when there are multiple errors, and added an error message for extraction errors.

Once we've found a satisfying solution with `validate`, I can squash my commits.

divarvel:
I can drop semigroupoids and use `ealt` instead of `<!>`.

divarvel:
I've downgraded deps in stack-lts-6.yaml, but there are still issues with transformers-compat.
I've tried removing semigroupoids as well, but it still fails. I'm not sure what I can try next.

Gabriel439:
@divarvel: I'll take a look at it tonight to see if I can get it building with `stack`'s LTS6 resolver

divarvel:
Thanks!

Gabriel439:
Alright, I have a pull request up against your branch that I believe should fix the LTS 6 build failures: https://github.com/divarvel/dhall-haskell/pull/1

divarvel:
Thanks a lot! It looks like it's fixed on LTS-6. I've rebased on latest master, so unless something else is missing, I can shash the commits.

Gabriel439:
@divarvel: You're welcome!  There's no need to rebase or squash ahead of time because the repository settings enforce that we "Squash and merge" anyway when merging pull requests

Issue #1009, opened 2019-06-14
Record projection by expression doesn't allow the expression to be a variable


quasicomputational:
```
⊢  let e = { a = 10, b = "Text" } let s = { a : Natural } in e.({ a : Natural }) 

{ a = 10 }

⊢  let e = { a = 10, b = "Text" } let s = { a : Natural } in e.(s) 

it : { a : Natural }
e : { a : Natural, b : Text }

Error: Unbound variable: s

1:                                                               s

(stdin):1:63
```

This exact example has type inference tested by `dhall/dhall-lang/tests/type-inference/success/unit/RecordProjectionByTypeA.dhall`, but evidently something's wrong with the actual evaluation.

quasicomputational:
Huh. Having added the example to the test suite under `normalization/success/unit/`, it works when the test suite runs and fails if I wibble its output, as you would expect if it was working, but it still breaks when I pass the test input to `dhall normalize`. I thought this would be a simple bug with a straightforward fix, but now I'm quite baffled at why the behaviour under test and the actual behaviour are differing.

Gabriel439:
I found the bug: the `subst` function does not do the right thing for the `Project` constructor:

https://github.com/dhall-lang/dhall-haskell/blob/b001a61a02b691977525f8833798ea3bf1128bac/dhall/src/Dhall/Core.hs#L1138-L1140

Specifically, it does not substitute into `s` of `e.(s)`.  The above code should be:

```haskell
 subst x e (Project a b) = Project a' b'
   where 
    a' = subst x e a
    b' = subst x e b 
```

Gabriel439:
Fix is up here: https://github.com/dhall-lang/dhall-haskell/pull/1012

PR #1012, opened 2019-06-18
Fix substitution into record projection by type

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/1009

The `subst`/`shift`/`subExpressions` functions did not correctly
reflect the recent change to support projection by type

### DIFF
--- a/dhall/src/Dhall/Core.hs
+++ b/dhall/src/Dhall/Core.hs
@@ -961,9 +961,10 @@ shift d v (Merge a b c) = Merge a' b' c'
 shift d v (Field a b) = Field a' b
   where
     a' = shift d v a
-shift d v (Project a b) = Project a' b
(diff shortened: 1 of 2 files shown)


quasicomputational:
Change looks good to fix the proximal problem, but I'm still concerned by how the testsuite wasn't able to catch it.

I guess the test code isn't using the same code-path as the production code somehow.

Gabriel439:
@quasicomputational: It's because the normalization test suite wasn't type-checking the code.  The (slower) type-checking code uses the `subst`/`shift` functions whereas the faster normalization code does not.

The absence of type-checking is the difference between test code path and the production code path.  The reason the test code is not type-checking is because of the unit tests (which are not well-typed since they are full of unbound variables).

PR #1014, opened 2019-06-18
dhall-lsp-server: (new feature) annotate lets

EggBaconAndSpam:
The "annotate Let bindings" feature is here!

An earlier version modified the AST and pretty-printed afterwards, which was rather annoying. This version instead inserts the annotation directly into the dhall source code.

In order to parse the `let` expressions I needed to access Dhall.Token and Dhall.Expression, which I exposed in dhall.cabal. Is that a good idea?

### DIFF
--- a/dhall-lsp-server/src/Dhall/LSP/Backend/Parsing.hs
+++ b/dhall-lsp-server/src/Dhall/LSP/Backend/Parsing.hs
@@ -11,8 +11,14 @@ import qualified Data.Text as Text
 import qualified Text.Megaparsec as Megaparsec
 import Text.Megaparsec (SourcePos(..))
 
-parseLetInnerOffset :: Parser (Int, SourcePos)
-parseLetInnerOffset = do
+
+-- | Parse the outermost binding in a Src descriptor of a let-block and return
+--   the rest. Ex. on input `let a = 0 let b = a in b` parses `let a = 0 ` and
+--   returns the Src descriptor containing `let b = a in b`.
+getLetInner :: Src -> Maybe Src
+getLetInner (Src left _ text) = Megaparsec.parseMaybe (unParser parseLetInnerOffset) text
+ where parseLetInnerOffset = do
+          setSourcePos left
   _let
   _ <- label
   _ <- optional (do
@@ -21,22 +27,18 @@ parseLetInnerOffset = do
   _equal
   _ <- expr
   _ <- optional _in
-  off <- getOffset
-  pos <- getSourcePos
-  _ <- Megaparsec.takeRest
-  return (off, pos)
-
--- | Parse the outermost binding in a Src descriptor of a let-block and return
---   the rest. Ex. on input `let a = 0 let b = a in b` parses `let a = 0 ` and
---   returns the Src descriptor containing `let b = a in b`.
-getLetInner :: Src -> Maybe Src
-getLetInner (Src left right text) =
-  case Megaparsec.parseMaybe (unParser parseLetInnerOffset) text of
-    Just (n, dpos) -> Just $ Src (addPosDelta left dpos) right (Text.drop n text)
-    Nothing -> Nothing
+          begin <- getSourcePos
+          tokens <- Megaparsec.takeRest
+          end <- getSourcePos
+          return (Src begin end tokens)
 
 addPosDelta :: SourcePos -> SourcePos -> SourcePos
 addPosDelta (SourcePos name line col) (SourcePos _ dline dcol) =
   (SourcePos name
     (Megaparsec.mkPos $ Megaparsec.unPos line + Megaparsec.unPos dline - 1)
     (Megaparsec.mkPos $ Megaparsec.unPos col + Megaparsec.unPos dcol - 1))
+
+setSourcePos :: SourcePos -> Parser ()
+setSourcePos src = Megaparsec.updateParserState
+                     (\(Megaparsec.State s o (Megaparsec.PosState i o' _ t l)) ->
+                       Megaparsec.State s o (Megaparsec.PosState i o' src t l))
--- a/dhall-lsp-server/src/Dhall/LSP/Backend/Parsing.hs
+++ b/dhall-lsp-server/src/Dhall/LSP/Backend/Parsing.hs
@@ -2,7 +2,6 @@ module Dhall.LSP.Backend.Parsing (getLetInner, getLetAnnot) where
 
 import Dhall.Src (Src(..))
 import Dhall.Parser
--- import Dhall.Parser.Combinators
 import Dhall.Parser.Token
 import Dhall.Parser.Expression
 
--- a/dhall-lsp-server/src/Dhall/LSP/Backend/Parsing.hs
+++ b/dhall-lsp-server/src/Dhall/LSP/Backend/Parsing.hs
@@ -1,0 +1,42 @@
+module Dhall.LSP.Backend.Parsing (getLetInner) where
+
+import Dhall.Src (Src(..))
+import Dhall.Parser
+-- import Dhall.Parser.Combinators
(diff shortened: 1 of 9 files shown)


dhall-lsp-server/src/Dhall/LSP/Backend/Typing.hs
```diff
@@ -52,14 +55,80 @@ typeAt' pos ctx expr = do
...
+printExpr expr = Pretty.renderStrict $ Pretty.layoutCompact (Pretty.unAnnotate (prettyCharacterSet Unicode expr))
+
+
+-- Split all multilets into single lets in an expression
+splitLets :: Expr Src a -> Expr Src a
+splitLets (Note src (Let (b :| (b' : bs)) e)) =
+  splitLets (Note src (Let (b :| []) (Note src' (Let (b' :| bs) e))))
+  where src' = case getLetInner src of
+                 Just x -> x
+                 Nothing -> error "The impossible happened: failed\
```
Gabriel439: I suggest returning a `Maybe`/`Either` instead of an `error`.  Each of the sites where you use `splitLets` supports gracefully handling the failure case gracefully

Gabriel439:
Also, for the changes to `dhall.cabal`, I think a simpler thing to do would be to just re-export the parser you need from `Dhall.Parser`, rather than exposing those modules.  In particular, if you create a parser for just the prefix of a `let` binding then you can narrow the re-export to that.

dhall-lsp-server/src/Dhall/LSP/Backend/Typing.hs
EggBaconAndSpam: Hmm that's an interesting design question. Wouldn't we want the entire thing to go up in flames if the "impossible" happens, rather than pretending that nothing happened?

dhall-lsp-server/src/Dhall/LSP/Backend/Typing.hs
```diff
@@ -52,14 +55,80 @@ typeAt' pos ctx expr = do
...
   let subExprs = toListOf subExpressions expr
   in case [ (src, e) | (Note src e) <- subExprs, pos `inside` src ] of
     [] -> Nothing
-    ((src, e) : _) -> srcAt pos e <|> Just src
+    ((src,e) : _) -> exprAt pos e <|> Just (Note src e)
+
+
+-- | Find the smallest Src annotation containing the given position.
+srcAt :: Position -> Expr Src a -> Maybe Src
+srcAt pos expr = do e <- exprAt pos expr
```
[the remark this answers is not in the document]
EggBaconAndSpam: Thanks, that's better! I was afraid of incomplete pattern matches, but turns out those are acceptable in MonadFail monads.

dhall-lsp-server/src/Dhall/LSP/Backend/Typing.hs
ocharles: The suggestion is not to pretend nothing happens, but to handle the `Nothing` at `splitLets` call site, where you likely have more information available to construct a good error message.

EggBaconAndSpam: Hmm... both call sites return a `Maybe`, where `Nothing` is not necessarily supposed to mean an error has occurred. Another example of this is in
```
annotateLet' pos ctx (Note src e@(Let (Binding _ _ a :| []) _))
  | not $ any (pos `inside`) [ src' | Note src' _ <- toListOf subExpressions e ]
  = do _A <- rightToMaybe $ typeWithA absurd ctx a
       let srcAnnot = case getLetAnnot src of
                        Just x -> x
                        Nothing -> error "The impossible happened: failed\
                                         \ to re-parse a Let expression."
       return (srcAnnot, ": " <> printExpr _A <> " ")
```
where again I think crashing is a better idea than just returning `Nothing`. Basically, failing to parse an expression that successfully parsed earlier can never happen: the branch is in a sense equivalent to `Nothing -> undefined`.

dhall-lsp-server/src/Dhall/LSP/Backend/Parsing.hs
```diff
@@ -0,0 +1,54 @@
...
+
+
+-- | Parse the outermost binding in a Src descriptor of a let-block and return
+--   the rest. Ex. on input `let a = 0 let b = a in b` parses `let a = 0 ` and
+--   returns the Src descriptor containing `let b = a in b`.
+getLetInner :: Src -> Maybe Src
+getLetInner (Src left _ text) = Megaparsec.parseMaybe (unParser parseLetInnerOffset) text
+ where parseLetInnerOffset = do
+          setSourcePos left
+          _let
```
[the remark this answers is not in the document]
EggBaconAndSpam: Hmm. Currently the parser for let bindings is somewhere inside a where block in the definition of `completeExpression :: Parser a -> Parser (Expr Src a)`, and references `expression`, which is also only defined in that where block. It would be possible to expose a `binding :: Parser (Binding Src a)` if we unfold the whole `where` monster.

dhall-lsp-server/src/Dhall/LSP/Backend/Typing.hs
EggBaconAndSpam: Ah I see what you mean now, but I can't think of a more informative error message in the context of `typeAt` or `annotateLet`. Furthermore, I think the type `splitLets :: Expr Src a -> Expr Src a` should be enough to guarantee that the function never fails (assuming the Src annotations come directly from the parser).

EggBaconAndSpam: Looking at it again, I now agree with you after all :) Have a look at the latest commit and tell me what you think.

ocharles: Possibly better, but you're unlikely to convince me that calling `error` is a good idea I'm afraid. If it's a case that `typeAt` and friends are called with data types that are "too big" then that is the exact problem - we need a more accurate data type! Ultimately, this is a data type identical to `Expr`, but with `Let` changed to not accept multiple bindings. Or maybe an `Expr` where the container for `Let` bindings is a parameter, and we can change `NonEmpty` to `Identity`

 However, I'm more of a drive by reviewer, so I hope my comments don't change whether or not this gets merged. I'm a bit of a stickler for correctness.

Gabriel439: Yeah, I agree with @ocharles that we should strive to make things total from end-to-end when feasible.  That doesn't necessarily mean throwing away the error as a `Nothing`: the `Maybe` type could be changed to an `Either` to preserve the error message.

Or, as @ocharles suggested, we could parametrize the `Expr` type on the number of bindings (i.e. `NonEmpty` vs. `Identity`) so that we could statically guarantee that the expression with split lets has exactly one binding per `Let`

EggBaconAndSpam: Ok, thinking about it some more I think I agree with the majority here ;)

Regarding `Expr`, wouldn't it make much more sense to simply drop lets with multiple binders completely from the AST representation? This would mean we would have to change the pretty printer to omit superfluous `in`s of course...

EggBaconAndSpam:
I don't really see how to factor the two parsers I need in such a way that I can avoid exporting too much from Dhall.Token... Is it ok if I leave it how it is at the moment?

The fundamental problem is that inserting annotations directly into the source code is always going to be a messy endeavour... For this (and future additions like freezing of imports) it would be quite useful to be able to access the internals of Dhall's parsers from dhall-lsp-server.

Gabriel439:
@EggBaconAndSpam: Yeah, that's fine.  It was just a minor suggestion

PR #1018, opened 2019-06-21
dhall-lsp-server: Correctly handle dependent lets

EggBaconAndSpam:
Previously `typeAt` and `annotateLet` failed in the presence of dependent lets. For example the following example failed (i.e. upon hovering over the body of the let no type was displayed):
```
let two = < zero : {} | one : {} >
in two.zero
```


Gabriel439:
@EggBaconAndSpam: Thank you! :slightly_smiling_face:

PR #1040, opened 2019-06-28
dhall-lsp-server: Implement caching

EggBaconAndSpam:
Implements caching of Dhall expressions: we only need to fetch, typecheck and normalise each import once per session, unless they change! This means that `dhall-lsp-server` is now viable for non-trivial
Dhall projects, for example probing around in `dhall-nethack` everything feels near-instantaneous once the imports have been resolved.

This implementation currently has a bug: we don't invalidate imports transitively, i.e. if `A.dhall` loads `B.dhall` and `B.dhall` changes we do not discard the cached version of `A.dhall`. This should be reasonably easy to fix given some time with Dhall's import graph. Furthermore, there is some cleaning up left to do:
  - Fix warnings
  - Reorganise things in a less ad-hoc way
  - Make the code a bit prettier

In adding the caching functionality I ended up rewriting most of the frontend; as a result, we got one other feature for free: better feedback. If for example linting fails because the file does parse as Dhall code then we display a corresponding warning rather than silently failing as before.

Sorry in advance for the massive PR...


Gabriel439:
@EggBaconAndSpam: Don't worry.  I've dealt with much larger pull requests than this before.  This is actually pretty tame 🙂

dhall-lsp-server/src/Dhall/LSP/Backend/Dhall.hs
```diff
@@ -1,34 +1,156 @@
...
+importFromFileIdentifier :: FileIdentifier -> Dhall.Import
+importFromFileIdentifier (FileIdentifier importType) =
+  Dhall.Import (Dhall.ImportHashed Nothing importType) Dhall.Code
+
+-- Construct the hashed import corresponding to the given file.
+hashedImportFromFileIdentifier :: FileIdentifier -> Digest SHA256 -> Dhall.Import
+hashedImportFromFileIdentifier (FileIdentifier importType) hash =
+  Dhall.Import (Dhall.ImportHashed (Just hash) importType) Dhall.Code
+
+-- | Invalidate all imports of a file (unhashed, as well as any hashed
```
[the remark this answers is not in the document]
EggBaconAndSpam: Suppose `A` imports `B` and `B` changes, then we would want to use the updated version of `B` the next time we are evaluating `A`. The reason whe need to invalidate `B` when it changes is that the new version may fail to typecheck (or even parse), which means we would never overwrite the old version using `cacheExpr`...

dhall-lsp-server/src/Dhall/LSP/Backend/Linting.hs
```diff
@@ -1,15 +1,14 @@
 module Dhall.LSP.Backend.Linting
-  ( suggest
-  , Suggestion(..)
-  , lintAndFormatDocument
+  ( Suggestion(..)
+  , suggest
+  , lint
```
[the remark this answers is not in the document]
EggBaconAndSpam: Didn't know you could do that, thanks!

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -2,59 +2,348 @@ module Dhall.LSP.Handlers where
...
-initializedHandler :: LSP.LspFuncs () -> J.InitializedNotification -> IO ()
+-- Workaround to make our single-threaded LSP fit dhall-lsp's API, which
+-- expects a multi-threaded implementation.
+wrapHandler
+  :: TVar (Maybe ServerState)
+  -> (a -> HandlerM ())
+  -> a
+  -> IO ()
+wrapHandler vstate handle message = do
+  mstate <- readTVarIO vstate
```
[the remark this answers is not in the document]
EggBaconAndSpam: I'm not too concerned about race conditions since I don't intend to introduce multitasking at all. The use of `TVar` over `IORef` dates back to the sample code from `haskell-lsp`...

On the other hand, I do like the idea of using `withMVar` to be on the safe side. Will change that!

dhall-lsp-server/src/Dhall/LSP/Backend/Dhall.hs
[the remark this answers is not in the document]
Gabriel439: Maybe I need to read more closely, but I assumed that you would only be invalidating the unhashed import and its transitive unhashed reverse dependencies.

A hashed import is valid indefinitely, though, even when the underlying import is broken.  That's why `missing sha256:…` works so long as the cache has that import cached (and this behavior is part of the standard).

dhall-lsp-server/src/Dhall/LSP/Backend/Dhall.hs
```diff
@@ -1,34 +1,156 @@
...
+                 $ emptyStatus
+  (do (expr', status') <- runStateT (Dhall.loadWith expr) status
+      let cache' = Cache $ view Dhall.cache status'
+      return . Right $ (cache', expr'))
+    `catch` (\e -> return . Left $ ErrorImportSourced e)
+    `catch` (\e -> return . Left $ ErrorInternal e)
+
+-- | Typecheck a fully resolved expression.
+typecheck :: Expr Src X -> Either DhallError WellTyped
+typecheck = first ErrorTypecheck . second WellTyped . Dhall.typeOf
```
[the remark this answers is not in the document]
EggBaconAndSpam: Thanks!

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -2,59 +2,348 @@ module Dhall.LSP.Handlers where
...
+  case mstate of
+    Just state -> do
+      state' <- (execStateT . runExceptT) (catchE (handle message) lspUserMessage) state
+      atomically . writeTVar vstate $ Just state'
+    Nothing ->
+      fail "A handler was called before the language server was initialized\
```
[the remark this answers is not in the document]
EggBaconAndSpam: By using `MVar`s and in particular `modifyMVar_ :: MVar a -> (a -> IO a) -> IO ()` we can get rid of the error message entirely. Though admittedly this means, were the impossible to happen the server would just lock up, since `modifyMVar` would wait indefinitely.

dhall-lsp-server/src/Dhall/LSP/Backend/Dhall.hs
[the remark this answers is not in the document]
EggBaconAndSpam: I fixed `invalidate` to only invalidate unhashed imports.

EggBaconAndSpam:
Is there a way to stop AppVeyor from building out-dated versions? Each build takes around half an hour...

Gabriel439:
@EggBaconAndSpam: I just sent you an invite to the AppVeyor project so that you (hopefully) should be able to stop/start AppVeyor builds

dhall-lsp-server/src/Dhall/LSP/Backend/Dhall.hs
```diff
@@ -1,34 +1,167 @@
...
+import Control.Monad.Trans.State.Strict (runStateT)
+import Network.URI (URI)
+import Data.Bifunctor (first)
+
+-- | A @FileIdentifier@ represents either a local file or a remote url.
+newtype FileIdentifier = FileIdentifier Dhall.ImportType
+
+-- | Construct a FileIdentifier from a given URI. Supports "file:", "http:" and
+--   "https:" URI schemes.
+--   TODO: Split into FilePath -> FileIdentifier and URI -> Maybe FileIdentifier
```
Gabriel439: It looks like this TODO is done

EggBaconAndSpam:
> @EggBaconAndSpam: I just sent you an invite to the AppVeyor project so that you (hopefully) should be able to stop/start AppVeyor builds

Got the invite, thanks!

dhall-lsp-server/src/Dhall/LSP/Backend/Dhall.hs
```diff
@@ -1,34 +1,167 @@
...
+-- Construct the hashed import corresponding to the given file.
+hashedImportFromFileIdentifier :: FileIdentifier -> Digest SHA256 -> Dhall.Import
+hashedImportFromFileIdentifier (FileIdentifier importType) hash =
+  Dhall.Import { importHashed = Dhall.ImportHashed (Just hash) importType,
+                 importMode = Dhall.Code }
+
+-- | Invalidate any _unhashed_ imports of the given file. Hashed imports are
+--   kept around as per
+--   https://github.com/dhall-lang/dhall-lang/blob/master/standard/imports.md.
+--   Note to future self: this doesn't correctly invalidate reverse
```
Gabriel439: You should probably create an issue to track improving this invalidation behavior

Gabriel439:
@EggBaconAndSpam: You're welcome! 🙂

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -2,59 +2,346 @@ module Dhall.LSP.Handlers where
...
+  asd <- liftIO $ getVirtualFileFunc uri
+  case asd of
+    Just (LSP.VirtualFile _ rope) -> return (Rope.toText rope)
+    Nothing -> fail $ "Could not find " <> show uri <> " in VFS."
+
+loadFile :: J.Uri -> HandlerM (Expr Src X)
+loadFile uri = do
+  txt <- readUri uri
+  fileIdentifier <- fileIdentifierFromUri uri
+  cache <- lift $ gets (view importCache)
```
Gabriel439: There is a `use` utility which you can use:

```haskel
  cache <- lift (use importCache)
```

Similarly, for something like:

```haskell
getVirtualFileFunc <- lift $ gets (LSP.getVirtualFileFunc . view lspFuncs)
```

... you can do:

```haskell
getVirtualFileFunc <- lift $ use (lspFuncs . to LSP.getVirtualFileFunc)
```

... or:

```haskell
getVirtualFileFunc <- lift $ uses lspFuncs LSP.getVirtualFileFunc
```

In fact, if you use `lens`/`lens-family` instead of `lens-family-core` then `use` is generalized to work with anything that implements `MonadState`, so then you don't even need the `lift`:

http://hackage.haskell.org/package/lens-family-1.2.3/docs/Lens-Family2-State-Lazy.html#v:use

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -2,59 +2,346 @@ module Dhall.LSP.Handlers where
...
+    Right e -> return e
+    _ -> throwE (Error, "Failed to parse Dhall file.")
+
+  loaded <- liftIO $ load fileIdentifier expr cache
+  (cache', expr') <- case loaded of
+    Right x -> return x
+    _ -> throwE (Error, "Failed to resolve imports.")
+  -- Update cache. Don't cache current expression because it might not have been
+  -- written to disk yet (readUri reads from the VFS).
+  lift $ modify (set importCache cache')
```
Gabriel439: Similarly, here you can use `assign` or `(.=`):

```haskell
assign importCache cache'
```

```haskell
importCache .= cache'
```

See: http://hackage.haskell.org/package/lens-family-1.2.3/docs/Lens-Family2-State-Lazy.html#v:assign

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -2,59 +2,346 @@ module Dhall.LSP.Handlers where
...
+            encodedDiag = URI.encode (Text.unpack diagnosis)
+            command = "[Explain error](dhall-explain:?"
+                        <> Text.pack encodedDiag <> " )"
+            _contents = J.List [J.PlainString command]
+        in Just J.Hover { .. }
+      hoverFromDiagnosis _ = Nothing
+
+      mHover = do err <- mError
+                  explanation <- explain txt err
+                  True <- return $ isHovered explanation
```
Gabriel439: You can do: `guard (isHovered explanation)`

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -2,59 +2,346 @@ module Dhall.LSP.Handlers where
...
+                  hoverFromDiagnosis explanation
+  lspRespond LSP.RspHover request mHover
+
+hoverType :: J.HoverRequest -> HandlerM ()
+hoverType request = do
+  let uri = request ^. J.params . J.textDocument . J.uri
+      (J.Position line col) = request ^. (J.params . J.position)
```
Gabriel439:
```suggestion
      J.Position line col = request ^. J.params . J.position
```

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -2,59 +2,346 @@ module Dhall.LSP.Handlers where
...
+
+      mHover = do err <- mError
+                  explanation <- explain txt err
+                  True <- return $ isHovered explanation
+                  hoverFromDiagnosis explanation
+  lspRespond LSP.RspHover request mHover
+
+hoverType :: J.HoverRequest -> HandlerM ()
+hoverType request = do
+  let uri = request ^. J.params . J.textDocument . J.uri
```
Gabriel439: Since lenses are first class, you can define a reusable composite lens from the request to the URI:

```haskell
requestUri = J.params . J.textDocument . J.uri
```

... and then you can reuse that for each handler:

```haskell
let uri = request ^. requestUri
```

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -2,59 +2,346 @@ module Dhall.LSP.Handlers where
...
+      welltyped <- case typecheck expr' of
+        Right (wt, _typ) -> return wt
+        Left err -> throwE err
+      let normal = normalize welltyped
+      -- cache the new expression
+      lift $ modify (set importCache (cacheExpr fileIdentifier normal cache'))
+      return Nothing
+
+  let suggestions =
+        case parse txt of
```
Gabriel439: It looks like this function parses text twice.  It might be possible to parse the text once and then share the result between the two use sites.

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -2,59 +2,346 @@ module Dhall.LSP.Handlers where
...
+    (J.ApplyWorkspaceEditParams edit)
+
+
+
+
+-- handler that doesn't do anything. Useful for example to make haskell-lsp shut
+-- up about unhandled DidChangeTextDocument notifications (which are already
+-- handled haskell-lsp itself).
+nullHandler :: a -> HandlerM ()
+nullHandler _ = return ()
```
Gabriel439: Nothing to do here, but it's a shame that `ExceptT` doesn't have a `Monoid` instance of the form:

```haskell
instance (Monad m, Monoid a) => Monoid (ExceptT e m a) where
    mempty = pure mempty

    mappend = liftA2 mappend
```

... because if it did then you could just define:

```haskell
nullHandler = mempty
```

dhall-lsp-server/src/Dhall/LSP/Backend/Dhall.hs
EggBaconAndSpam: Ups.

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
EggBaconAndSpam: I believe parsing to be cheap compared to typechecking, so why complicate the code? :)

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
EggBaconAndSpam: Which would save us two characters and thoroughly confuse everybody xD

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
EggBaconAndSpam: I tried that, but it doesn't work due to the way `haskell-lsp-types` is set up: we can't give a type to `requestUri` in general.

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
EggBaconAndSpam: Ups, that line has been copy-pasted around for a while now.

dhall-lsp-server/src/Dhall/LSP/Backend/Dhall.hs
EggBaconAndSpam: Will do!

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
Gabriel439: Ah, I didn't realize that they were using chained multi-parameter type-classes.  Never mind, then

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
EggBaconAndSpam: Since `dhall-lsp-server` is already depending on `lens`: `cache <- use importCache`. Much nicer!

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
EggBaconAndSpam: I guess that's a bit more idiomatic ;)

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
EggBaconAndSpam: I'll leave this for the time being. Might come back to it when the typechecker gets fast enough that this could make a difference.

Issue #1039, opened 2019-06-28
Reviving fast type-checking branch


Gabriel439:
The purpose of this issue is to document my attempt to revive @AndrasKovacs's unfinished [`nbe-elaboration` branch](https://github.com/dhall-lang/dhall-haskell/commits/nbe-elaboration)

Up until now the two main approaches that I've tried were:

* Merge the `nbe-elaboration` branch into `master` one commit at a time
* Merge the `master` into the `nbe-elaboration` branch one commit at a time

However, both of these approaches do not work because the initial commit of the `nbe-elaboration` branch deleted a large amount of code.  That means that the only remaining approach left is to reverse engineer what the `nbe-elaboration` did to improve performance and manually reimplement those changes on top of `master`.

I'm using this issue to document anything on that branch that appears to be related to the performance improvement and track my progress on forwarding porting these changes to `master`:

* [ ] Use normalization-by-evaluation for the type-checking code
* [ ] Interleaving type-checking and normalization
* [x] Change the `Var` constructor to use `Int` instead of `Integer` and `UNPACK` the `Text` field
* [ ] Use pointer-equality fast path for judgmental equality
* [ ] Use `unsafeCoerce` instead of `denote`
* [ ] Use strict fields for abstract syntax tree
* [ ] Switch to a monomorphic `Val` syntax tree
* [ ] Use `INLINE` instead of `INLINABLE` for `Dhall.Map.{toMap,keys}`
* [ ] Speed up `Dhall.Map.insert`, which is currently O(N) in the number of existing keys
* [ ] Use "linear" programming style where every variable is consumed exactly once

  This will make more sense if you study https://github.com/dhall-lang/dhall-haskell/commit/627a6cdea0170336ff08de34851d8bdf5180571d in depth

I'm not sure how much each of the following changes contribute to performance until I attempt to manually forward-port them.

AndrasKovacs:
I intend to do the merging into `master` this summer; I've been busy since my last bout of activity in the repository, and I'm going to have significant spare time again roughly two weeks from now. I indeed deleted and shuffled lot of code on the `nbe-elaboration` branch, but in the next code sprint I plan to be more judicious and ideally retain all functionality in the `dhall-haskell` repository. I also think that the most efficient way to do this would be to reimplement features on the latest master. It doesn't seem too be bad because I'm much more familiar with the sources now than last time. 

[...]

PR #1044, opened 2019-07-01
Switch `Var` to use an `Int`

Gabriel439:
Related to: https://github.com/dhall-lang/dhall-haskell/issues/1039

We'll probably never see indices that exceed the space of an `Int` and
the interpreter would probably not be able to handle 9223372036854775807
nested binders anyway.


dhall/src/Dhall/Core.hs
```diff
@@ -336,7 +336,7 @@ instance Pretty Import where
     Zero indices are omitted when pretty-printing `Var`s and non-zero indices
     appear as a numeric suffix.
 -}
-data Var = V Text !Integer
+data Var = V Text !Int
```
ocharles: Maybe go all the way and `{-# UNPACK #-}`?

ocharles: Actually, I think GHC already unpacks these strict fields now.

Gabriel439: Yeah, I believe GHC will automatically unbox strict fields if they are no larger than a pointer.

Note that I haven't actually benchmarked this; I'm only forward-porting small chunks of the work mentioned in #1039

PR #1060, opened 2019-07-03
dhall-lsp-server: Normalize types before displaying them to the user

EggBaconAndSpam:
Previously, the type displayed when hovering over the arrow in a lambda
(or forall) binder would not always be normal:

This fix normalises types before displaying them to the user (fixes both
the type-on-hover and annotate-let behaviour):

### DIFF
--- a/dhall-lsp-server/src/Dhall/LSP/Backend/Typing.hs
+++ b/dhall-lsp-server/src/Dhall/LSP/Backend/Typing.hs
@@ -9,7 +9,7 @@ import Data.List.NonEmpty (NonEmpty (..))
 import Control.Lens (toListOf)
 import Data.Text (Text)
 import Control.Applicative ((<|>))
-import Data.Bifunctor (first)
+import Data.Bifunctor (first, bimap)
 
 import Dhall.LSP.Backend.Parsing (getLetInner, getLetAnnot)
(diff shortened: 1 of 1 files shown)


ocharles:
Hmm, this is a tricky one. Normal forms are not necessarily what you want to see - they are often significantly larger than the original expression. However, for types, because we don't have full computation in types (dependent types) that's not too likely *yet*, but it may happen in the future.

EggBaconAndSpam:
I think normal forms are more in line with the expected behaviour. I stumbled over this in `dhall-nethack`. For example, with the old behaviour, the following type does not even fit into the tooltip window:


While the normalised type is far more informative (and shorter):


I think the essence here is that we don't have _computation_ in types, only polymorphism, and beta-reduction in the presence of polymorphism "almost always" makes the type smaller.

EggBaconAndSpam:
@ocharles what is your opinion in this? Yay, Nay, undecided?

ocharles:
Oh, please don't let me change what's happening here - it was merely a passing thought. If you think this is good, then stick with it. We'll find out overtime if it's not, anyway!

EggBaconAndSpam:
No worries! You did raise an interesting point that I hadn't considered before, though. I think I'll merge and hope for the best ;)

EggBaconAndSpam:
So how do I tell mergify that this should be merged?

PR #1073, opened 2019-07-05
New Feature: Optimistic caching

EggBaconAndSpam:
Normally, we only cache hashed (i.e. "frozen") imports. This commit adds
a flag `--use-dump` that causes dhall to keep the state of its import
cache around in the file "dhall-cache" in the current directory. The
resulting speedup is the same as would be gained by manually freezing
all imports, with the important difference that whenever a _local_
import changes it and its reverse dependencies are correctly
re-evaluated.

This feature is a prototype! There are a few things left to do:

- Improve the user interface:
  Either separate the flag into `--use-dump` and `--write-dump` to be
  more explicit. In that case using a hidden ".dhall-cache" would be
  reasonable.

  Otherwise use an option `--dump-file = <file>` that takes a file name
  explicitely.

- Pessimistic timestamps:
  Currently we compare the modification times of imports against the
  modification time of the cache dump. Since the dump gets written after
  everything is loaded this creates a race condition. It would be better
  to date back the cache dump to the time before we start loading so
  we never miss any changes.

- Code improvements:
  - cacheFromDump is too complicated
  - Possibly move cacheFromDump from Dhall.Main to Dhall.Import?

Note that this PR depends on #1069 &ndash; that one should probably be merged first.

An example:
```
$ time dhall <<< https://github.com/vmchale/cpkg/raw/master/pkgs/pkg-set.dhall > /dev/null

real	0m33.443s
user	0m28.549s
sys	0m0.915s

$ rm dhall-cache 
$ time dhall --use-dump <<< https://github.com/vmchale/cpkg/raw/master/pkgs/pkg-set.dhall > /dev/null

real	0m33.629s
user	0m29.600s
sys	0m0.931s
$ time dhall --use-dump <<< https://github.com/vmchale/cpkg/raw/master/pkgs/pkg-set.dhall > /dev/null

real	0m13.569s
user	0m12.466s
sys	0m0.989s
```


f-f:
@EggBaconAndSpam great to see this! 🎉 

A few points:
- why not use file content hashes instead of modification times?
- why not use the global Dhall cache to store this info (using absolute paths) instead of a file in the current directory?
- I assume the flag is there because this is experiemental? I'd love this to become the default behaviour

EggBaconAndSpam:
- `make` also uses modification times; hashes are presumably slower (read vs stat?)
- Good question. In my head I had this as an "unsafe" feature, but this depends on how unhashed remote imports are being used in the wild... One argument for using a local cache is that we can conveniently use a single file for everything (while this might become too slow if the cache contains everything you ever looked at?). Otherwise we'd probably store each file individually, negating the first point of timestamps vs hashes.
- I originally intended the flag to be opt-in (like stack's `--fast`). See previous point.

Gabriel439:
@EggBaconAndSpam: I would check to see if hashes are fast enough, because timestamps are not necessarily accurate.  The usual way that timestamps give incorrect results is if you depend upon a file that is located underneath a symlinked directory: if you change the symlink to point to a different directory the file can change without the timestamp being modified.  This scenario is more common than you think (especially in the Nix world, which depends heavily on symlinked directories and timestamps set to the Unix epoch).

I understand that this is an opt-in optimization that is advertised as possibly unsafe, but I think it would add more value for enterprise users if it could be enabled all the time with only a slight performance cost.  My reasoning is that the typical person introducing Dhall within their company usually fits this profile:

* They probably chose Dhall because they place a high premium on correctness

* They are putting their reputation on the line by introducing a new technology within their company

* They will be expected to support the use of Dhall within the company since they were the ones who introduced it

* Users in enterprise environments often blindly copy ("cargo cult") flags/settings without understanding their tradeoffs 

* The original advocate may not have the time nor proficiency to help other users debug or navigate these tradeoffs

However, it might be that hashes end up being so slow that they negate the optimization, in which case we would have to settle for timestamps.  We just need some performance numbers to compare, though, before making a decision.

EggBaconAndSpam:
@f-f Taking a step back, I think what you are proposing makes a lot more sense. I guess the main motivation behind this prototype was making things easy for me ;)

The thing I forgot about is that (at least at the moment) the bottleneck is _typechecking_. If it takes upwards of half a minute to typecheck something we can certainly ignore the time it takes to read anything from disk (or, more likely, ssd).

@Gabriel439 You are arguing your case well; I am convinced :)

I'll try and implement caching based on hashes instead (on a new branch). I'll also leave this branch around for the time being, as inspiration and later for benchmarking comparisons.

EggBaconAndSpam:
Brainstorming:

- We can't use content hashes for indexing, since there might be multiple dhall files with the same content but with different meaning (e.g. "./B.dhall" has a different meaning depending on `B.dhall`).

**Proposal 1&nbsp;** Store the cached expressions by absolute file path / URL (hashed, of course). Keep a separate map from file paths to content hashes for invalidation. Also keep the entire import graph (where import paths have been mapped to absolute paths) around, in order to be able to invalidate transitive dependencies.

  There are a few problems with this approach:
   - renaming a directory of dhall files changes the indices of all files inside, which means we have to re-evaluate everything.
  - There is loads of interdependent state to keep consistent. Not great if we want this feature enabled by default.

**Proposal 2&nbsp;** In a first pass, parse and load all (transitive) dependencies of the input file. Construct a "Merkle tree" (sort of) from the import graph, i.e. starting from the leaves compute the content hashes of each node and hash them together with the resulting hashes of the nodes corresponding to its imports. Use these "Merkle" hashes as addresses into the cache.

The main advantage of this approach is that there isn't any additional state to keep consistent. Of course we need to handle remote imports separately, e.g. by keeping around a map from URLs to content hashes and treating remote imports as "leaf" imports. But this approach is a lot less brittle and easier to get right.

The main downside of this "hash-of-hashes" approach is that we need to parse everything (except for remote imports) even if the files are in the cache. However, I would argue that parsing is a lot less of a performance concern right now, and unlikely to become one any time soon.

Gabriel439:
@EggBaconAndSpam: I'd go with Proposal 2 (the Merkle tree)

EggBaconAndSpam:
I'm still working on this and should have a prototype by tomorrow.

PR #1094, opened 2019-07-11
Add `dhall-json` support for unions as keys

Gabriel439:
This is inspired by this StackOverflow question:

https://stackoverflow.com/questions/56967374/dynamic-records-key-type


Gabriel439:
@sjakobi: I added documentation, but I'm going to split out the test-related changes into a separate pull request

PR #1101, opened 2019-07-13
Build macOS binaries with Travis

### LINKED ISSUE #737: "OSX release binaries" should be available
I get support requests for some projects I have that use Dhall from OSX users who don't realize they can use the linux releases. It would be good to upload the same content as `-macos` or something like that.

f-f:
Fix #737 

This PR changes the Travis build (I'm replacing the current one which I don't understand with the one [from spago](https://github.com/spacchetti/spago/blob/master/.travis.yml) which I mostly wrote) to compile macOS binaries and push them to GitHub releases alongside the Linux and Windows binaries.

Things left to do:
- [x] enable the Travis integration (I don't have access to it)
- [x] add an `$API_KEY` variable [to authenticate to GitHub](https://docs.travis-ci.com/user/deployment/releases/) (though I can reuse the current `$GITHUB_OAUTH_TOKEN` if it's still there)
- [ ] verify that packaging is all fine, and move the `tar` operations to the `before_deploy` section

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ install:
 
 script:
   - export PATH="${PATH}:$(pwd)/artifacts"
-  - stack build --copy-bins --local-bin-path ./bin;
+  - stack build --copy-bins --local-bin-path ./bin
   - get_cabal_version() { cat $1/$1.cabal | grep '^Version: ' | gsed -e 's/^Version: //g'; }
   - mk_release_name() { echo "$1-$(get_cabal_version $1)-x86_64-macos.tar.bz2"; }
   - tar -jcvf $(mk_release_name dhall) bin/dhall
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ before_deploy:
 
 deploy:
   - provider: releases
-    api_key: $API_KEY
+    api_key: $GITHUB_OAUTH_TOKEN
     file_glob: true
     file: uploads/*
     skip_cleanup: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_install:
   # Download and unpack the stack executable
   - export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$HOME/.local/bin:/opt/alex/$ALEXVER/bin:/opt/happy/$HAPPYVER/bin:$HOME/.cabal/bin:$PATH
   - mkdir -p ~/.local/bin
-  -  travis_retry curl --insecure -L https://get.haskellstack.org/stable/osx-x86_64.tar.gz | tar xz --strip-components=1 --include '*/stack' -C ~/.local/bin
+  - travis_retry curl --insecure -L https://get.haskellstack.org/stable/osx-x86_64.tar.gz | tar xz --strip-components=1 --include '*/stack' -C ~/.local/bin
   # Use the more reliable S3 mirror of Hackage
   - mkdir -p $HOME/.cabal
   - echo 'remote-repo: hackage.haskell.org:http://hackage.fpcomplete.com/' > $HOME/.cabal/config
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,100 +1,75 @@
-# This file has been generated -- see https://github.com/hvr/multi-ghc-travis
-language: c
-sudo: false
+sudo: required
+
+language: generic
 
 cache:
   directories:
-    - $HOME/.cabsnap
-    - $HOME/.cabal/packages
-
-before_cache:
-  - rm -fv $HOME/.cabal/packages/hackage.haskell.org/build-reports.log
-  - rm -fv $HOME/.cabal/packages/hackage.haskell.org/00-index.tar
+  - $HOME/.ghc
+  - $HOME/.cabal
+  - $HOME/.stack
+  - $TRAVIS_BUILD_DIR/.stack-work
 
 matrix:
   include:
-    - env: CABALVER=1.24 GHCVER=8.0.1 DEPLOY_GITHUB_RELEASE=true
-      compiler: ": #GHC 8.0.1"
-      addons: {apt: {packages: [cabal-install-1.24,ghc-8.0.1], sources: [hvr-ghc]}}
-    - env: CABALVER=1.24 GHCVER=8.0.1 DEPLOY_GITHUB_RELEASE=true
-      compiler: ": #GHC 8.0.1"
-      os: osx
+  # Build only on macOS
+  - env: BUILD=stack
+    compiler: ": #stack default osx"
+    os: osx
+
+# Build only master and release tags
+branches:
+  only:
+  - master
+  - /^\d+\.\d+\.\d+(\.\d+)?$/
+
 
 before_install:
- - unset CC
- - export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$PATH
- # Work around travis issue by updating Homebrew: https://github.com/travis-ci/travis-ci/issues/8552
- - if [ "$TRAVIS_OS_NAME" = osx ];
-   then
-     brew update;
-     brew install cabal-install;
-     brew install gnu-sed --with-default-names;
-   fi
+  # Using compiler above sets CC to an invalid value, so unset it
+  - unset CC
+  # We want to always allow newer versions of packages when building on GHC HEAD
+  - CABALARGS=""
+  - if [ "x$GHCVER" = "xhead" ]; then CABALARGS=--allow-newer; fi
+  # Download and unpack the stack executable
+  - export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$HOME/.local/bin:/opt/alex/$ALEXVER/bin:/opt/happy/$HAPPYVER/bin:$HOME/.cabal/bin:$PATH
+  - mkdir -p ~/.local/bin
+  -  travis_retry curl --insecure -L https://get.haskellstack.org/stable/osx-x86_64.tar.gz | tar xz --strip-components=1 --include '*/stack' -C ~/.local/bin
+  # Use the more reliable S3 mirror of Hackage
+  - mkdir -p $HOME/.cabal
+  - echo 'remote-repo: hackage.haskell.org:http://hackage.fpcomplete.com/' > $HOME/.cabal/config
+  - echo 'remote-repo-cache: $HOME/.cabal/packages' >> $HOME/.cabal/config
+
 
 install:
- - cabal --version
- - echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]"
- - if [ -f $HOME/.cabal/packages/hackage.haskell.org/00-index.tar.gz ];
-   then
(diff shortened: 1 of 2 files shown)


jneira:
Travis has experimental windows support, so you can build in the three envs, what do you think?

f-f:
@jneira the only purpose of this Travis build is to output macOS binaries, so I wouldn't want to enable anything else on the matrix (since we already build Windows on Appveyor and Linux on Hydra)

jneira:
Yeah, i was thinking in remove appveyor (but keep nix) to somewhat unify the ci, appveyor has one hour of build time limit
And maybe add a linux build to test it without nix

f-f:
@jneira I see. I really don't like Travis, I'd say if we have to setup out new CI builds we could try Azure. Recently both `stack` and [`haskell-ide-engine`](https://github.com/haskell/haskell-ide-engine/pull/1238) switched to it

jneira:
Yep, my idea is to use only one, travis already has cache support and azure has stable windows env and more concurrent jobs (i think).

.travis.yml
```diff
@@ -1,100 +1,75 @@
...
-   fi
+  # Using compiler above sets CC to an invalid value, so unset it
+  - unset CC
+  # We want to always allow newer versions of packages when building on GHC HEAD
+  - CABALARGS=""
+  - if [ "x$GHCVER" = "xhead" ]; then CABALARGS=--allow-newer; fi
+  # Download and unpack the stack executable
+  - export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$HOME/.local/bin:/opt/alex/$ALEXVER/bin:/opt/happy/$HAPPYVER/bin:$HOME/.cabal/bin:$PATH
+  - mkdir -p ~/.local/bin
+  -  travis_retry curl --insecure -L https://get.haskellstack.org/stable/osx-x86_64.tar.gz | tar xz --strip-components=1 --include '*/stack' -C ~/.local/bin
```
joneshf:
```suggestion
  - travis_retry curl --insecure -L https://get.haskellstack.org/stable/osx-x86_64.tar.gz | tar xz --strip-components=1 --include '*/stack' -C ~/.local/bin
```

Looks like an extra space snuck in there.

.travis.yml
```diff
@@ -1,100 +1,75 @@
...
- - cabal check
- - cabal sdist   # tests that a source-distribution can be generated
-
-# Check that the resulting source distribution can be built & installed.
-# If there are no other `.tar.gz` files in `dist`, this can be even simpler:
-# `cabal install --force-reinstalls dist/*-*.tar.gz`
- - SRC_TGZ=$(cabal info . | awk '{print $2;exit}').tar.gz &&
-   (cd dist && cabal install --force-reinstalls "$SRC_TGZ")
+  - export PATH="${PATH}:$(pwd)/artifacts"
+  - stack build --copy-bins --local-bin-path ./bin;
```
joneshf:
```suggestion
  - stack build --copy-bins --local-bin-path ./bin
```

I _think_ this semicolon is unnecessary. If it's needed, feel free to ignore this suggestion.

.travis.yml
```diff
@@ -1,100 +1,75 @@
...
+  - tar -jcvf $(mk_release_name dhall) bin/dhall
+  - tar -jcvf $(mk_release_name dhall-json) bin/dhall-to-json bin/dhall-to-yaml bin/json-to-dhall bin/yaml-to-dhall
+  - tar -jcvf $(mk_release_name dhall-bash) bin/dhall-to-bash
+  - tar -jcvf $(mk_release_name dhall-lsp-server) bin/dhall-lsp-server
 
 before_deploy:
-  - tar --create --file "$TRAVIS_OS_NAME.tar" --files-from /dev/null
-  - tar --append --file "$TRAVIS_OS_NAME.tar" --directory dist/build/dhall dhall
-  - gzip "$TRAVIS_OS_NAME.tar"
+  - stack build --copy-bins --local-bin-path ./bin;
```
joneshf:
```suggestion
  - stack build --copy-bins --local-bin-path ./bin
```

Similar thing here about the semicolon.

.travis.yml
```diff
@@ -1,100 +1,75 @@
...
+  # TODO: move the tar operations here
```
joneshf: The `before_deploy` phase only runs when the `deploy` phase runs, right? Since we only run the `deploy` phase on tags, it'd be nice to know whether the book keeping of the deploy works separate from the actual process of deploying.

I'd be willing to suggest that we even move this stuff into a phase that [runs all the time][job lifecycle]: `script` or `after_success`. Nothing really needs to change for this PR, though. The way it's laid out currently solves the problem at hand, and that's more valuable than toiling in perfection.

[job lifecycle]: https://docs.travis-ci.com/user/job-lifecycle/

f-f: Would this mean _not_ moving the tar operations as the TODO suggests?

joneshf: Right.

Gabriel439:
@f-f: Alright, Travis is enabled with a `GITHUB_OAUTH_TOKEN` installed, but you'll need to trigger a new change to test it on this pull request

Gabriel439:
@f-f: Hmmm.  For some reason Travis doesn't seem to be registering the existence of your pull request.  You might have to close and reopen your pull request

.travis.yml
f-f: @joneshf do you know if the env is shared between stages (so that we can build in `script` and have a no-op here) or should I duplicate the tar stuff?

.travis.yml
```diff
@@ -1,100 +1,73 @@
...
-     brew install cabal-install;
-     brew install gnu-sed --with-default-names;
-   fi
+  # Using compiler above sets CC to an invalid value, so unset it
+  - unset CC
+  - CABALARGS=""
+  - export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$HOME/.local/bin:/opt/alex/$ALEXVER/bin:/opt/happy/$HAPPYVER/bin:$HOME/.cabal/bin:$PATH
+  - | # Install stack
+    mkdir -p ~/.local/bin
+    travis_retry curl --insecure -L https://get.haskellstack.org/stable/osx-x86_64.tar.gz | tar xz --strip-components=1 --include '*/stack' -C ~/.local/bin
```
Gabriel439: Is the `--insecure` necessary?

.travis.yml
```diff
@@ -1,100 +1,73 @@
...
-     cp -a $HOME/.cabsnap/lib $HOME/.cabsnap/share $HOME/.cabsnap/bin $HOME/.cabal/;
-   else
-     echo "cabal build-cache MISS";
-     rm -rf $HOME/.cabsnap;
-     mkdir -p $HOME/.ghc $HOME/.cabal/lib $HOME/.cabal/share $HOME/.cabal/bin;
-     cabal install --only-dependencies --enable-tests --enable-benchmarks;
-   fi
+install:
+  - echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]"
+  - if [ -f configure.ac ]; then autoreconf -i; fi
```
Gabriel439: What is this line for?

.travis.yml
```diff
@@ -1,100 +1,73 @@
...
+  - set -ex
+  - stack --no-terminal --install-ghc $ARGS test --bench --only-dependencies
```
Gabriel439: What is `$ARGS` set to?

Also, I don't think we need to run tests or benchmarks.  We already do that in the Nix build.  We only need the Travis build for generating OS X binaries

jneira: Ideally it should be enough but would not be possible to have specific errors or perf problems in OS X?

f-f: Note that this is building dependencies only, we're not running the tests of Dhall itself

f-f:
> When this is enabled, will be running this on pull requests or just the master branch and tags?

It will build PRs, master and tags. There is the following rule in the config, that might be confusing in this regard:
```yml
# Build only master and release tags
branches:
  only:
  - master
  - /^\d+\.\d+\.\d+(\.\d+)?$/
```

This is to prevent other branches from being built, because they are covered by Travis' "PR builds", otherwise you'd get two builds from every PR with a branch in this repo

.travis.yml
Gabriel439: Then shouldn't this dependency-only command also only use `build` instead of `test --bench`?

My understanding is that the actual build command for the `dhall` packages is just `stack build` (i.e. doesn't run tests or benchmarks), meaning that we don't need to install test-related dependencies or benchmark-related dependencies.

f-f: Yep, I changed it to `build` and removed the unneeded flags

.travis.yml
```diff
@@ -1,100 +1,72 @@
...
-     echo "cabal build-cache MISS";
-     rm -rf $HOME/.cabsnap;
-     mkdir -p $HOME/.ghc $HOME/.cabal/lib $HOME/.cabal/share $HOME/.cabal/bin;
-     cabal install --only-dependencies --enable-tests --enable-benchmarks;
-   fi
+install:
+  - echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]"
+  - set -ex
+  - stack --no-terminal --install-ghc build --only-dependencies
+  - set +ex
```
Gabriel439: Note that this restore only works if the settings before the `set -ex` were `set +ex` (which is probably the case, but in theory it could not be! :slightly_smiling_face: )

A safer way to restore settings is to save them before modifying them and then restore the saved settings, like this (taken from https://superuser.com/a/946451/236574)"

```bash
SAVED_OPTIONS=$(set +o)
set -ex
stack --no-terminal --install-ghc build --only-dependencies
eval "$SAVED_OPTIONS"
```

.travis.yml
joneshf: I think that's what the [`skip_cleanup`][] flag controls. There's only that blurb about it though.

[`skip_cleanup`]: https://docs.travis-ci.com/user/job-lifecycle/#deploying-your-code

.travis.yml
f-f: Thanks, TIL! 🙂

.travis.yml
f-f: @joneshf cool thanks, let's try that and fix it in next release if it doesn't work

paulyoung:
This is great! Can we expect a release soon that includes these binaries?

f-f:
@paulyoung yep, once [Dhall 9](https://github.com/dhall-lang/dhall-lang/pull/635) is out (sometime in the next days) there will be a matching release for the Haskell package too

Gabriel439:
@f-f: Thank you for doing this! 🙂

f-f:
You're welcome 🙂

paulyoung:
@f-f it occurred to me that the binaries built by Travis might be available somewhere in the meantime. Is that the case?

f-f:
@paulyoung the binaries are indeed build at every CI run, but they are not uploaded anywhere, so they get "lost" after the build is done (and there's no way to recover them as far as I know)

paulyoung:
Now that Dhall 9 has been released, is there a timeframe for a new dhall-haskell release?

Gabriel439:
@paulyoung: ~3 days.  I'm going to begin the process of cutting the release tonight

Gabriel439:
@paulyoung: The executables are available now from the releases page: https://github.com/dhall-lang/dhall-haskell/releases

PR #1105, opened 2019-07-13
Build macOS binaries with Travis

f-f:
Replaces #1101

### DIFF
--- a/.travis-functions.sh
+++ b/.travis-functions.sh
@@ -1,0 +1,3 @@
+get_cabal_version() { cat $1/$1.cabal | grep '^Version: ' | gsed -e 's/^Version: //g'; }
+
+mk_release_name() { echo "$1-$(get_cabal_version $1)-x86_64-macos.tar.bz2"; }
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ install:
 
 script:
   - export PATH="${PATH}:$(pwd)/artifacts"
-  - stack build --copy-bins --local-bin-path ./bin;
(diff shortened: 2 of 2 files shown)


Gabriel439:
@f-f: Okay, I think I figured out what is going on.  Builds that depend on shared secrets can't be triggered from pull requests from forked repositories (otherwise they could exploit the shared secret):

https://github.com/medialize/ally.js/issues/95

f-f:
@Gabriel439 but the branch is in this repo?

Gabriel439:
@f-f: Oh, weird.  Then I'm not sure.  For reference, Travis did trigger on another pull request that I pushed.  I'm still trying to figure out why it doesn't trigger on yours

Gabriel439:
It seems to only trigger on my branches and pull requests, like this one I opened:

https://github.com/dhall-lang/dhall-haskell/pull/1107

Maybe it's because the token belongs to me?

f-f:
@Gabriel439 I think that's unlikely. We have the same setup in Spago (with private token and everything) and it triggers on every PR by anyone

Gabriel439:
@f-f: I added you as an admin for this repository so you should be able to manage the Travis settings yourself now.  Check them out and see if anything looks out of the ordinary

f-f:
@Gabriel439 thanks! I was wondering if it could be because we are still on `travis-ci.org` but everything should move to `travis-ci.com`

f-f:
Update: found it, apparently Travis cannot parse the yml file and silently fails

f-f:
Closing this one and reopening #1101, sorry for the noise

PR #1113, opened 2019-07-15
Preparing `Dhall.Import` for "Semi-semantic" caching

EggBaconAndSpam:
This PR consists of a number of smaller changes I would like to make to `Dhall.Import` to prepare for the next step, which is to reorganise the code to load imports recursively. I copy the commit messages of the two main commits below.

There are a few questions that came up along the way:
- I believe `Local parent (File (Directory dir) file)` to be equivalent to `Local Here (File (Directory (".." : dir)) file)`; if that is the case, could we not just get rid of the `Parent` constructor entirely? 
- Is there a reason for using `throwIO` as well as `throwM`, and not simply sticking with the latter?

Finally, I feel that 'resolved' would be the best way to describe fully 'chained' imports whose headers have been typechecked and normalised, in analogy to 'DNS resolution'; `Chained` is very much a compromise. It is a shame that `Resolved` was already appropriated to mean something a bit more vague... ;) 

#### Commit: Add `Chained` type to capture fully chained imports

Until now we used `Import` two mean two different things:
- The syntactic construct; e.g. `./a.dhall` corresponds to the following
  AST:
  ```
  Embed
    (Import
      (ImportHashed Nothing (Local Here (Directory ["."]) "a.dhall"))
      Code)
  ```

- The physical location the import is pointing to, computed by
'chaining' the syntactical import with the the 'physical' parent import.
For example the syntactic import `./a.dhall` might actually refer to the
remote file `http://host/directory/a.dhall`.

This commit adds a `Chained` newtype on top of `Import` to make this
distinction explicit at type level.

#### Commit: Typecheck and normalise http headers earlier

Previously we would typecheck and normalise http headers in
`exprFromImport`, i.e. while loading the import. This commit adds the
invariant that any headers in 'Chained' imports are already typechecked
and normalised, and moves this step into `loadWith` accordingly.

This causes a subtle difference in behaviour when importing remote files
with headers `as Location`: previously, nonsensical expressions like
`http://a using (0 0) as Location` were valid, while they would now cause
a type error.


dhall/src/Dhall/Import.hs
```diff
@@ -390,29 +390,23 @@ instance Canonicalize Import where
...
-  -> Text
-  -> Expr s a
-  -> Maybe (CI Data.ByteString.ByteString, Data.ByteString.ByteString)
-toHeader key₀ key₁ (RecordLit m) = do
-    TextLit (Chunks [] keyText  ) <- Dhall.Map.lookup key₀ m
-    TextLit (Chunks [] valueText) <- Dhall.Map.lookup key₁ m
+toHeader :: Expr s a -> Maybe HTTPHeader
+toHeader (RecordLit m) = do
+    TextLit (Chunks [] keyText  ) <-
+        Dhall.Map.lookup "header" m <|> Dhall.Map.lookup "mapKey" m
```
Gabriel439: This is not exactly the same semantics as the standard since it permits `List { header : Text, mapValue : Text }`

You would need to do something like:

```haskell
toHeader (RecordLit m) = do
    let alternative0 = do
            TextLit (Chunks [] keyText) <- Dhall.Map.lookup "header" m
            TextLit (Chunks [] keyValue) <- Dhall.Map.lookup "value" m
            return (keyText, keyValue)

    let alternative1 = do
            TextLit (Chunks [] keyText) <- Dhall.Map.lookup "mapKey" m
            TextLit (Chunks [] keyValue) <- Dhall.Map.lookup "mapValue" m
            return (keyText, keyValue)

    (keyText, valueText) <- alternative0 <|> alternative1

    …
```

dhall/src/Dhall/Import/Types.hs
```diff
@@ -33,28 +34,35 @@ import qualified Dhall.Context
 import qualified Data.Map      as Map
 import qualified Data.Text
 
+-- | A fully 'chained' import, i.e. if it contains a relative path that path is
+--   relative to the current directory. If it is a remote import with headers
+--   those are well-typed and in normal form. Note that the type itself isn't
+--   precise enough to capture either of these invariants.
```
Gabriel439: Even though the type might not capture those invariants, you could make it so that the API exposed by the module captures those invariants.  The module could expose only safe primitives for building and transforming `Chained` imports

EggBaconAndSpam: That was the original idea. I messed that up when trying to get dhall-lsp-server to compile.

dhall/src/Dhall/Import.hs
EggBaconAndSpam: That the header expression is of a valid type (so either `List { header : Text, value Text}` or `List { mapKey : Text, mapValue Text}`) is one of the invariants of the `Chained` type. In the end we call `toHeader` with a header from a `Chained` remote import, so that should be alright.

I guess I should make that more explicit in the comment describing the `Chained` type.

Gabriel439:
@f-f: See the Travis failure on this pull request.  The build appears to time out during uploading the cache: https://travis-ci.org/dhall-lang/dhall-haskell/builds/559476240

dhall/src/Dhall/Import.hs
```diff
@@ -605,8 +621,58 @@ getCacheDirectory = alternative₀ <|> alternative₁
...
+                        Annot expr expected
+
+            _ <- case (Dhall.TypeCheck.typeOf annot) of
+                Left err -> throwM err
+                Right _ -> return ()
+
+            return (Dhall.Core.normalize expr)
+
+    let handler₀ (e :: SomeException) = do
+            {- Try to normalize using the preferred @mapKey@ /
```
Gabriel439:
```suggestion
            {- Try to type-check using the preferred @mapKey@ /
```

f-f:
@Gabriel439 oh wow. Caches are 800MB in size, which is less than what we get in Spago, but it's the first time I see this error. I tried removing caching for `~/.local/bin`, which isn't really needed

dhall/src/Dhall/Import.hs
EggBaconAndSpam: Well spotted!

dhall/src/Dhall/Import.hs
EggBaconAndSpam: I added a comment to clarify these assumptions.

EggBaconAndSpam:
> There are a few questions that came up along the way:
> 
>     * I believe `Local parent (File (Directory dir) file)` to be equivalent to `Local Here (File (Directory (".." : dir)) file)`; if that is the case, could we not just get rid of the `Parent` constructor entirely?
> 
>     * Is there a reason for using `throwIO` as well as `throwM`, and not simply sticking with the latter?

Any insights on this, @Gabriel439?

Gabriel439:
> I believe `Local parent (File (Directory dir) file)` to be equivalent to `Local Here (File (Directory (".." : dir)) file)`; if that is the case, could we not just get rid of the `Parent` constructor entirely?

See: https://github.com/dhall-lang/dhall-haskell/pull/718

> Is there a reason for using `throwIO` as well as `throwM`, and not simply sticking with the latter?

As far as I know, we use `throwM` for code that is polymorphic over `MonadThrow` and `throwIO` for code that is either `IO` or polymorphic over `MonadIO`.

For the case of `IO` code, there's no particular reason we prefer `throwIO` over `throwM`.  It's just that a lot of that code predates the use of `MonadThrow` in `Dhall.Import`

EggBaconAndSpam:
Thanks for the clarifications!

Gabriel439:
@EggBaconAndSpam: You're welcome! :slight

Issue #89, opened 2017-07-21
Feature request: remove use of TH and libraries that require TH


vmchale:
As it stands, I like Dhall a lot, but unfortunately I can't cross compile it. If TH were removed, I would be able to use it in one of my bigger projects.

Thanks!

PR #1119, opened 2019-07-16
Add instance Lift (Expr s a)

ocharles:
This allows Exprs to be lifted in template-haskell. This is useful to build a `[dhall| ... |]` quasiquoter


dhall/dhall.cabal
```diff
@@ -559,6 +559,7 @@ Library
         scientific                  >= 0.3.0.0  && < 0.4 ,
         template-haskell                           < 2.15,
         text                        >= 0.11.1.0 && < 1.3 ,
+        th-lift-instances           >= 0.1.13   && < 0.2 ,
```
ocharles: Needed to automatically derive `Lift` in various cases. If we don't want to pollute downstream users of `dhall` with orphan instances, I can look into writing these out by hand.

Gabriel439: I'm not that familiar with `Lift`, but my reasoning is that if these instances are basically unambiguous (i.e. there's only one right way to define them), then I don't think our downstream users will mind

dhall/src/Dhall/Src.hs
```diff
@@ -1,6 +1,8 @@
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveGeneric      #-}
+{-# LANGUAGE DeriveLift         #-}
 {-# LANGUAGE OverloadedStrings  #-}
+{-# LANGUAGE TemplateHaskell    #-}
```
Gabriel439: Note that @vmchale at one point requested that Dhall not use `TemplateHaskell` or `QuasiQuotes` due to issues with cross-compilation.  See: https://github.com/dhall-lang/dhall-haskell/issues/89

@vmchale: Is this still a concern for you?

Gabriel439: Actually, even if it were, we could guard this behind a `cabal configure` flag so that users can compile out Template Haskell support if necessary.   But let's still confirm whether or not we even need to do so

ocharles: It wouldn't be hard to remove it from this file.

dhall/dhall.cabal
ocharles: Yea, the idea of `Lift` is "given this Haskell value, give me source code that would produce the same value". So `lift [1,2,3] = App (:) 1 (App (:) 2 (App (:) 3 [])))`. I think we can trust these instances to be canonical, but just missing from the respective libraries for whatever reason.

dhall/src/Dhall/Src.hs
Gabriel439: Oh, if it's easy to do, then go ahead.  The real issue here is that I just don't really understand what are the exact requirements for Dhall to be easy-to-cross-compile.  For example, it's not clear to me if depending on `th-lift-instances` matters or not for the original use case @vmchale had in mind.

quasicomputational: This can actually get away with just `TemplateHaskellQuotes`, which is cross-compilation friendly.

sjakobi:
Sorry for the merge conflicts, @ocharles! :/

ocharles:
No problem! I have a little more work to do anyway, should be easy to fix up.

dhall/src/Dhall/Src.hs
ocharles: I've changed this to `TemplateHaskellQuotes`, I hadn't even heard of that extension!

ocharles:
Anyone got any ideas on the build failure? Is this Cabal messing up? It seems to be in the "Preprocessing" stage.

AndrasKovacs:
FWIW, when I built your branch with stack just now, I got the same error, and added `th-lift-instances-0.1.13@sha256:2852e468511805cb25d9e3923c9e91647d008ab4a764ec0921e5e40ff8a8e874` to `stack.yaml`. Then I got one more similar error about some `th` package, and added that to `stack.yaml` as well, and then it worked.

ocharles:
Oh, that's a slightly different error but I do probably need to fix the `stack.yml` to use the newer versions of `th-lift-instances` and `th-lift`.

ocharles:
I have a feeling that even though `DeriveLift` is supported on 7.10, this doesn't build on 7.10 because Cabal wasn't updated in time. https://github.com/haskell/cabal/pull/2831 is the PR that teaches Cabal about the extension. @Gabriel439 what do you want to do about this?

Gabriel439:
@ocharles: I think the only option is to guard all of the `Lift`-related changes behind a Cabal macro.  We currently have to support GHC 7.10 for Eta

quasicomputational:
Cabal actually doesn't care - that list is only relevant for `other-extensions` / `default-extensions`, and even then it does the sensible thing of passing unknown values through to GHC, which is what really counts.

ocharles:
@quasicomputational are you sure? Because this extension has existed since 7.4, yet the build is failing on 7.10

quasicomputational:
[GHC's manual says new in 8.0](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/glasgow_exts.html#extension-DeriveLift), so actually, yeah, this won't work without either conditionally exposing the `Lift` instance (which IMO is a bad idea) or defining it manually.

ocharles:
Huh. [This](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-DeriveLift) says since 7.2.1

ocharles:
I guess this was a bug and the documentation has been fixed in HEAD?

sjakobi:
> I guess this was a bug and the documentation has been fixed in HEAD?

Right: https://gitlab.haskell.org/ghc/ghc/commit/1ad3c8245cc114c3f862633d027361700fba3e50

ocharles:
Ok, I've taken a different approach using the default instance for `Lift` for any types that have a `Data` instance. All our types do, so this should work just as well.

Gabriel439:
@ocharles: It looks like we might need to replace `TemplateHaskellQuotes` with another extension for older GHC versions

ocharles:
Yea I saw. Sorry, I'm traveling a lot this weekend so I'm just pushing
commits out when I get a few minutes spare.

Gabriel439:

ocharles:
Oh good, I finally got all CI systems happy!

PierreR:
#

PR #1121, opened 2019-07-16
dhall-lsp-server: Turn imports into clickable links

EggBaconAndSpam:
Note that remote imports were already clickable since they were recognised by VSCode as urls.

### DIFF
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -613,6 +613,7 @@ exprFromUncachedImport import_@(Import {..}) = do
     let resolveImport importType' = case importType' of
           Local prefix file -> liftIO $ do
               path   <- localToPath prefix file
+              absolutePath <- Directory.makeAbsolute path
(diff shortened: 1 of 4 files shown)


dhall/src/Dhall/Import.hs
```diff
@@ -451,11 +454,10 @@ localToPath prefix file_ = liftIO $ do
             return "/"
 
         Parent -> do
-            pwd <- Directory.getCurrentDirectory
-            return (FilePath.takeDirectory pwd)
+            return ".."
 
         Here -> do
-            Directory.getCurrentDirectory
+            return "."
```
EggBaconAndSpam: What was the rationale behind having `localToPath` return absolute paths in the first place?

EggBaconAndSpam: Never mind, I simply changed `exprFromUncachedImport` to call `makeAbsolute` on the resulting path, thus recovering the original behaviour.

PR #1123, opened 2019-07-17
dhall-lsp-server: Freezing imports

EggBaconAndSpam:
If the import in question was already frozen update to the newest version (i.e. "refreeze").


dhall-lsp-server/src/Dhall/LSP/Backend/Diagnostics.hs
```diff
@@ -119,15 +120,23 @@ positionToMegaparsec (line, col) = Megaparsec.SourcePos ""
                                      (Megaparsec.mkPos $ max 0 line + 1)
                                      (Megaparsec.mkPos $ max 0 col + 1)
 
+addRelativePosition :: Position -> Position -> Position
+addRelativePosition (x1, y1) (0, dy2) = (x1, y1 + dy2)
+addRelativePosition (x1, _) (dx2, y2) = (x1 + dx2, y2)
+
+-- addRelativePosition pos (relativePosition pos pos') = pos'
```
Gabriel439: If you enable `doctest` for the `dhall-lsp-server` package then you could turn this into a checked property test:

```haskell
-- | prop> addRelativePosition pos (relativePosition pos pos') == pos'
```

dhall-lsp-server/src/Dhall/LSP/Backend/Freezing.hs
```diff
@@ -0,0 +1,56 @@
...
+  loaded <- load fileid expr cache
+  case loaded of
+    Left err -> return (Left err)
+    Right (cache', expr') -> case typecheck expr' of
+      Left err -> return (Left err)
+      Right (wt,_) -> do
+        return (Right (cache', hashNormalToCode (normalize wt)))
+
+stripHash :: Import -> Import
+stripHash (Import (ImportHashed _ importType) mode) =
```
Gabriel439: This is outside of the scope of this pull request, but at some point we should add some lenses for dealing with the deeply nested `Import*` hierarchy, so that this could be factored into:

```haskell
strip lens = set lens empty

hash = importHashedLens . hashLens

stripHash = strip hash
```

dhall-lsp-server/src/Dhall/LSP/Backend/Freezing.hs
```diff
@@ -0,0 +1,56 @@
...
+  Import (ImportHashed Nothing importType) mode
+
+getImportHashPosition :: Src -> Maybe Range
+getImportHashPosition src@(Src left _ text)  = do
+  Src left' right' _ <- getImportHash src
+  let (x0, y0) = positionFromMegaparsec left
```
Gabriel439: Since you don't use `x0`/`y0`/`x1`/`y1` individually, you could instead pass around the positions as a whole:

```haskell
let p0 = positionFromMegaparsec left
let p1 = positionFromMegaparsec left'
let off1 = positionToOffset text (relativePosition p0 p1)
```

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -358,6 +362,74 @@ executeAnnotateLet request = do
...
+  uri <- getCommandArguments request
+
+  fileIdentifier <- fileIdentifierFromUri uri
+  txt <- readUri uri
+  expr <- case parse txt of
+    Right e -> return e
+    Left _ -> throwE (Warning, "Could not freeze imports; did not parse.")
+
+  let importRanges = getAllImportsWithHashPositions expr
+  edits <- mapM (\(import_, Range (x1, y1) (x2, y2)) -> do
```
Gabriel439: Instead of:

```haskell
edits <- mapM (\... -> do
    ...
    ) importRanges
```

... you can do:

```haskell
edits <- forM importRanges $ \... -> do
    ...
```

... where `forM` is from `Control.Monad`

dhall-lsp-server/src/Dhall/LSP/Backend/Freezing.hs
f-f: This would be amazing, we have some pretty verbose pattern matches in Spago that we could totally get rid of if we had lenses (e.g. see [here](https://github.com/spacchetti/spago/blob/3eaba105a0a9ff59a35709e5d6336c1d81d73215/src/Spago/PackageSet.hs#L144))

dhall-lsp-server/src/Dhall/LSP/Backend/Diagnostics.hs
```diff
@@ -119,15 +120,23 @@ positionToMegaparsec (line, col) = Megaparsec.SourcePos ""
...
+relativePosition :: Position -> Position -> Position
```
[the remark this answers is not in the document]
EggBaconAndSpam: I agree that's better ;)

dhall-lsp-server/src/Dhall/LSP/Backend/Freezing.hs
EggBaconAndSpam: Well spotted :)

Issue #1146, opened 2019-07-22
Import alternatives do not recover from type errors


EggBaconAndSpam:
`BAD="0 0" dhall <<< "env:BAD ? 0"` results in the following error:
```
↳ env:BAD

Error: Not a function

1│ 0 0

BAD:1:1
```
This is behaviour is a bit surprising given that the alternative operator `?` successfully recovers from parsing errors and missing files, e.g. `BAD="42 ?" dhall <<< "env:BAD ? 0"` evaluates to 0. The relevant part of the [standard](https://github.com/dhall-lang/dhall-lang/blob/master/standard/imports.md) says the following, which I interpret to mean that we should allow `?` to recover from type errors:
> Resolution of expressions might not be always successful: [..] imports
> might not resolve in cases like:
> - an environment variable is not defined
> - [..]
> - typecheck error
> By using the ? operator, expressions are alternatively resolved [..]

The fix is trivial, but I'll wait a bit in case I'm misreading the situation.

PR #1152, opened 2019-07-25
Fix import alternatives to recover from type errors

EggBaconAndSpam:
Previously, `BAD="0 0" dhall <<< "env:BAD ? 0"` resulted in the
following error:
```
↳ env:BAD

Error: Not a function

1│ 0 0

BAD:1:1
```

According to the standard the above expression was supposed to evaluate
successfully to `0`. See #1146 for further discussion.


Gabriel439:
Thank you!

Issue #1098, opened 2019-07-12
Proposal: Semi-semantic caching


EggBaconAndSpam:
## Introduction

I propose ‘semi-semantic’ hashing as a solution to caching of non-frozen imports. The semi-semantic approach to caching is motivated by the fact that parsing is computationally trivial compared to typechecking and normalisation. See #888 for the detailed reasoning.

I use *semi-semantic* to mean that any two *semantically* different (i.e. differing in their normal forms) inputs are mapped to distinct hashes, while semantically equivalent inputs that differ *syntactically* may result in different hashes. We could compute the semi-semantic hash of a dhall import/file/expression as follows:
- Parse the input;
- compute the *semantic hashes* of all imports referenced in the AST, i.e. the hashes of their normal forms;
- compute the *syntactic hash* of the input (for example hashing the unparsed input, or alternatively by hashing the parsed AST);
- concatenate the syntactic hash of the input with the semantic hashes of its imports and hash the result.

### Update
The current implementation #1154 substitutes any imports by their normal form and hashes the resulting expression to obtain a *semi-semantic* hash. This is easier to implement and, since we store normal forms anyway, only incurs a constant-factor cost.

## Discussion

A cache indexed by semi-semantic hashes has the following properties:
- For a given input we can quickly find out if it is in the cache: we only need to parse the input – we don't need to typecheck or normalise it!
- The cache stays consistent, that is, we don't need to ‘invalidate’ old cache entries if their dependencies change!

Comparison to alternative approaches:
- Naive syntactic hashing is problematic because two syntactically equivalent dhall expressions may have different semantics depending on the context (e.g. if they contain relative imports).
- We can fix syntactic hashing by ‘resolving’ the absolute location of all imports before hashing. However, in this case we still need to worry about invalidating cached results if their imports change.
- Local caching as implemented in #1073. Here we simply dump all loaded imports after each invocation of dhall. The problem here is again keeping the cache consistent; additionally, the user experience is suboptimal since for example changing to a different directory causes dhall to start from scratch with an empty cache. Also, we leave cache files everywhere.

The main disadvantage of a ‘semi-semantic’ approach is that it
- incures a slight performance penalty when evaluating an expression with a clean cache since every import needs to be separately encoded and written to disk (currently we only ever encode and write ‘frozen’ imports to disk)
[...]

PR #1154, opened 2019-07-25
Implement semi-semantic caching

EggBaconAndSpam:
This completes the implementation of the "semi-semantic caching"
proposal (issue #1098).

We compute the semi-semantic hash of a dhall import/file/expression as
follows:

- Parse the input;
- compute the semantic hashes of all imports referenced in the AST, i.e.
the hashes of their normal forms;
- compute the syntactic hash of the input (hashing the parsed AST);
- concatenate the syntactic hash of the input with the semantic hashes
of its imports and hash the result.

The "semi-semantic" cache (normal forms, indexed by semi-semantic
hashes) has the following properties:

- For a given input we can quickly find out if it is in the cache: we
only need to parse the input – we don't need to typecheck or normalise
it!
- The cache stays consistent, that is, we don't need to ‘invalidate’ old
cache entries if their dependencies change!

### DIFF
--- a/dhall/src/Dhall/Import/Types.hs
+++ b/dhall/src/Dhall/Import/Types.hs
@@ -28,7 +28,6 @@ import Dhall.TypeCheck (X)
 import Lens.Family (LensLike')
 import System.FilePath (isRelative, splitDirectories)
 
-import qualified Crypto.Hash
 import qualified Dhall.Binary
 import qualified Dhall.Context
 import qualified Data.Map      as Map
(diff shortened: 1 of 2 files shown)


dhall/src/Dhall/Import/Types.hs
```diff
@@ -47,6 +48,18 @@ instance Pretty Chained where
...
+    , semanticHash :: Crypto.Hash.Digest Crypto.Hash.SHA256
+    -- ^ Its semantic hash, i.e. the hash of its encoded normal form.
+    }
+
+data ResolvedExpr = ResolvedExpr
+    { resolvedExpr :: Expr Src X
+    -- ^ The fully resolved expression, i.e. any imports where replaced by their
+    --   semantics. Note that the resolved expression has not yet been
+    --   typechecked nor normalised!
+    , imports :: [ImportSemantics]
```
Gabriel439: Why do you include the `imports` when hashing if they are already included as subexpressions of `resolvedExpr`?

EggBaconAndSpam: Hah, never thought about that! That makes the code simpler as well. Thanks!

EggBaconAndSpam:
There we go, a much smaller diff now!

EggBaconAndSpam:
Will do!

Issue #1135, opened 2019-07-19
Tests cannot be ran without an active network connection


ocharles:
For example, when trying to install Dhall with the Nix sandbox enabled:

```
      ./dhall-lang/tests/import/success/headerForwardingA.dhall:                                FAIL
        Exception: 
        ↳ https://test.dhall-lang.org/Bool/package.dhall using [ { mapKey = "Test", mapValue = "Example" } ]
        
        Error: Remote host not found
        
        
        15│ https://test.dhall-lang.org/Bool/package.dhall
        16│   using [ { mapKey = "Test", mapValue = "Example" } ]
        
        15:1
```

I suggest we spin up a local HTTP server for these tests.

Gabriel439:
@ocharles: How would we get the tests to use to that server?

ocharles:
One option is to rewrite all the tests to access `localhost` instead and document how test harnesses should setup a local server.

Gabriel439:
@ocharles: The main issue is that this greatly increases the test burden of implementation authors, especially the parts of the test that require CORS support or two domains, one of which is protected by authenticated headers

ocharles:
I suppose the other option is that `dhall-haskell` could be more configurable in terms of how it does HTTP, and we could just mock the responses out rather than even bothering with a server. Then when we run tests, we can implement a pure approach

ocharles:
Thanks for the response. I think there may be a solution here to fix this, but we don't need to keep the issue around. But if someone wants to propose a fix in a PR (e.g., me) - great!

ocharles:
Also, it looks like #1159 should have fixed this anyway, if I understand correctly.

EggBaconAndSpam:
> Also, it looks like #1159 should have fixed this anyway, if I understand correctly.


Yes! I forgot to mention that here...

Anton-Latukha:
I would bump this.

Accepting the https://github.com/dhall-lang/dhall-haskell/issues/1135#issuecomment-513241959 greatly.

But since thread gives further hints of resolving this.

Is this can be resolved now?

Nixpkgs escapes testing the package in its ecosystem because of _*this_ issue, since Nix builds by default do not permit the network use.

---

[...]

PR #1155, opened 2019-07-25
Fix `isNormalized` test

EggBaconAndSpam:
The test in question failed intermittently because `isNormalized` is
more thorough than `normalize`, in the sense that it throws exceptions
for more non-welltyped expressions. As a result, we need to use `spoon`
not just when computing the normal form of a raw expression, but also
when calling `isNormalized` on the result. Note that the test may still
randomly fail in the future, because normalizing non-welltyped
expressions needn't terminate!


EggBaconAndSpam:
Good idea!

EggBaconAndSpam:
Now this one fails:
```
      .\dhall-lang\tests\import\success\customHeadersA.dhall:                                   FAIL
        Exception: 
        Error: Unexpected HTTP status code:
```
This is (somewhat) related to #1135.

PR #1160, opened 2019-07-27
Fix CSS for `dhall-lang.org` in Safari

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/609

Safari requires a vertical alignment CSS directive in order to render the
input and output panes side-by-side

### DIFF
--- a/dhall-try/index.html
+++ b/dhall-try/index.html
@@ -19,12 +19,14 @@
         overflow: auto;
         width: 60ch !important;
         display: inline-block;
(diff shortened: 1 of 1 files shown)


Gabriel439:
Here is an example of what the site looks like after the fix in Safari:

Screen Shot 2019-07-26 at 8 01 34 PM

I also checked in Firefox and Chrome, too, to make sure that change didn't introduce any new regressions.  They all render correctly now

PR #1162, opened 2019-07-27
Change `ImportSemantics` to not necessarily be α-normalized

Gabriel439:
The motivation for this change is to avoid α-normalizing all imported
expressions.

For example, before this change you would get the following behavior
beginning with an empty cache

```
$ cat ./example.dhall
λ(a : Type) → a

$ dhall <<< './example.dhall'
λ(_ : Type) → _
```

The reason why is that the current code α-normalizes all imported
expressions, even when returning them fresh.

To fix this, I changed the `ImportSemantics` type to not require that
expressions are α-normalized.  Instead, the α-normalization only
happens at the last minute when interacting with the semantic cache, but
nowhere else.

I figured that this change would also be fine from the perspective of
the semi-semantic cache because false-negatives for this cache are
fine.  In particular, we probably don't mind if we get a cache miss for
the semi-semantic cache if the user renames a variable.

After this change imports are no longer α-normalized, whether loaded
from a hot or cold cache:

```
$ cat ./example.dhall
λ(a : Type) → a

$ dhall <<< './example.dhall'
λ(a : Type) → a

$ dhall <<< './example.dhall'
λ(a : Type) → a
```

### DIFF
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -517,14 +517,14 @@ loadImportWithSemanticCache
         Nothing -> do
             ImportSemantics { importSemantics } <- loadImportWithSemisemanticCache import_
 
-            let variants = map (\version -> encodeExpression version importSemantics)
+            let variants = map (\version -> encodeExpression version (Dhall.Core.alphaNormalize importSemantics))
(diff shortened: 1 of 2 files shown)


EggBaconAndSpam:
Looks good to me!

PR #1169, opened 2019-07-30
Remove most uses of `StandardVersion` from the API

Gabriel439:
We no longer support multiple versions of the standard, except for
supporting old integrity checks, so this change removes all inessential
uses of `StandardVersion` from the API and command-line interface.


EggBaconAndSpam:
Yay for simpler code!

PR #1173, opened 2019-07-30
Restore support for records containing both types and terms

ocharles:
... as standardized in https://github.com/dhall-lang/dhall-lang/pull/689


sjakobi:
This makes sense to me so far – but I don't know much about these things.

I think you should probably also adjust the typechecking for `/\`, `//` (and probably `//\\` too?). In both cases we currently check that the two records have the same kind, which I assume we don't want to check anymore. That would allow something like

    { a = 0 } // { a = Bool }

ocharles:
Thanks @sjakobi - I wasn't sure if that was explicitly checked in the type checker. I'll have a look and update that.

dhall/src/Dhall/TypeCheck.hs
```diff
@@ -474,47 +476,22 @@ typeWithA tpa = loop
...
-                kts <- Dhall.Map.unorderedTraverseWithKey process kvs'
+        let process k v = do
+                t <- loop ctx v
+                s <- fmap Dhall.Core.normalize (loop ctx t)
+                case s of
+                    Const _ -> return t
+                    _ -> Left (TypeError ctx e (InvalidFieldType k t))
 
-                return (Record (Dhall.Map.insert k0 t0 kts))
+        Record <$> Dhall.Map.unorderedTraverseWithKey process kvs
```
Gabriel439: I believe you also need an additional step to type-check the record type returned here.  i.e.:

```haskell
recordType <- Record <$> Dhall.Map.unorderedTraverseWithKey process kvs

_ <- loop ctx recordType

return recordType
```

dhall/src/Dhall/TypeCheck.hs
```diff
@@ -581,10 +558,6 @@ typeWithA tpa = loop
             Const constY -> return constY
             _            -> Left (TypeError ctx e (MustCombineARecord '∧' kvsY tKvsY))
 
-        if constX == constY
-            then return ()
-            else Left (TypeError ctx e (RecordMismatch '∧' kvsX kvsY constX constY))
```
sjakobi: The same change is necessary for `//` (`Prefer`).

dhall/src/Dhall/TypeCheck.hs
ocharles: Thanks, done.

ocharles:
This is getting there, but I'm stuck on what to do with one test case. The test input is:

```dhall
{ x = Type, y = Kind }
```

Which does have a type - it has type

```dhall
{ x : Kind, y : Sort }
```

However, the current test harness can't actually verify this, because the expected type is itself type checked, and `{ x : Kind, y : Sort }` does not have a type (because `Sort` doesn't have a type).

Should I reject `RecordLit`s that contain `Kind`s?

Gabriel439:
@ocharles: Yes, I believe you would need to reject records that contain `Kind`s.  This is actually not a consequence of the test harness limitations but rather a consequence of the rule that record literal's inferred type must itself type-check, and a record type with a `Sort` in it is not valid.

ocharles:
Thanks for the clarification - I'll see what I can do.

MonoidMusician:
Nice! Did you see this issue: https://github.com/dhall-lang/dhall-lang/issues/558 ? It might be nice to put it in with this change, but we can always take care of it in a separate PR too – up to you.

dhall/src/Dhall/TypeCheck.hs
```diff
@@ -456,47 +458,26 @@ typeWithA tpa = loop
...
-                c <- case s0 of
-                    Const c -> pure c
-                    _ -> Left (TypeError ctx e (InvalidFieldType k0 t0))
-                let process k t = do
-                        s <- fmap Dhall.Core.normalize (loop ctx t)
-                        case s of
-                            Const c' ->
-                                if c == c'
-                                then return ()
-                                else Left (TypeError ctx e (FieldAnnotationMismatch k t c k0 t0 c'))
```
sjakobi: I think the `FieldAnnotationMismatch` error is obsolete now.

dhall/src/Dhall/TypeCheck.hs
```diff
@@ -456,47 +458,26 @@ typeWithA tpa = loop
...
+                s <- fmap Dhall.Core.normalize (loop ctx t)
+                case s of
+                    Const _ -> return t
+                    _ -> Left (TypeError ctx e (InvalidFieldType k t))
 
-                        return t
+        recordType <- Record <$> Dhall.Map.unorderedTraverseWithKey process kvs
 
-                kts <- Dhall.Map.unorderedTraverseWithKey process kvs'
+        _ <- loop ctx recordType
```
sjakobi: Why is this check necessary? If it is necessary, do we have a test for it?

dhall/src/Dhall/TypeCheck.hs
```diff
@@ -456,47 +458,26 @@ typeWithA tpa = loop
...
-                    Const c -> pure c
-                    _       -> Left (TypeError ctx e (InvalidFieldType k0 v0))
-
-                let process k v = do
-                        t <- loop ctx v
-                        s <- fmap Dhall.Core.normalize (loop ctx t)
-                        case s of
-                            Const c'
-                                | c == c'   -> return ()
-                                | otherwise -> Left (TypeError ctx e (FieldMismatch k v c k0 v0 c'))
```
sjakobi: Same here.

dhall/src/Dhall/TypeCheck.hs
```diff
@@ -546,19 +527,15 @@ typeWithA tpa = loop
             _          -> Left (TypeError ctx e (MustCombineARecord '∧' kvsY tKvsY))
 
         ttKvsX <- fmap Dhall.Core.normalize (loop ctx tKvsX)
-        constX <- case ttKvsX of
+        _      <- case ttKvsX of
             Const constX -> return constX
```
sjakobi: Just return `()`.

dhall/src/Dhall/TypeCheck.hs
```diff
@@ -546,19 +527,15 @@ typeWithA tpa = loop
...
             _            -> Left (TypeError ctx e (MustCombineARecord '∧' kvsX tKvsX))
 
         ttKvsY <- fmap Dhall.Core.normalize (loop ctx tKvsY)
-        constY <- case ttKvsY of
+        _      <- case ttKvsY of
             Const constY -> return constY
```
sjakobi: Same here.

dhall/src/Dhall/TypeCheck.hs
sjakobi: Or is this check necessary at all?

dhall/src/Dhall/TypeCheck.hs
```diff
@@ -621,19 +590,15 @@ typeWithA tpa = loop
...
             _            -> Left (TypeError ctx e (MustCombineARecord '⫽' kvsY tKvsY))
 
-        if constX == constY
-            then return ()
-            else Left (TypeError ctx e (RecordMismatch '⫽' kvsX kvsY constX constY))
```
sjakobi: `RecordMismatch` should be obsolete now.

dhall/src/Dhall/TypeCheck.hs
```diff
@@ -621,19 +590,15 @@ typeWithA tpa = loop
             _          -> Left (TypeError ctx e (MustCombineARecord '⫽' kvsY tKvsY))
 
         ttKvsX <- fmap Dhall.Core.normalize (loop ctx tKvsX)
-        constX <- case ttKvsX of
+        _      <- case ttKvsX of
             Const constX -> return constX
```
sjakobi: Same comments as for `Combine`.

dhall/src/Dhall/TypeCheck.hs
Gabriel439: It's required by the standard, but it might be redundant now that we permit mixed records.  In other words, this additional check required by the standard might not actually catch anything that the other checks do so there wouldn't be a differentiating test, but the standard would need to be fixed.

Gabriel439: Either way I'll remove it as I'm reasonably sure it's no longer necessary and the standard change can follow

dhall/src/Dhall/TypeCheck.hs
Gabriel439: It is not necessary any longer, so I'll remove it.  One invariant of the type system is that the type of the type of something will always be a type-checking constant (i.e. `Const`) so if we don't care about the specific constant then we no longer need to check this

Gabriel439:
@ocharles: I opened a pull request against your branch to fix the test failure and address the review feedback: https://github.com/circuithub/dhall-haskell/pull/1

ocharles:
Thanks for that! It was a little hard to do the review but I trust you and
am not too invested in how this work is done - the standard and
understanding the changes was more important to me. As such, I've just gone
ahead and merged your PR. Also, I'm not sure if you're aware but GitHub
will let you push changes directly to my branch, so feel free to push more
fixes directly here. I will also try and help next week to push this over
the line.

Cheers!

PR #1190, opened 2019-08-02
dhall-lsp-server: Implement completion support

EggBaconAndSpam:
Completes the following:
- environment variables
- local imports
- identifiers in scope (as well as built-ins)
- record projections from 'easy-to-parse' records (of the form `ident1.ident2[.ident3...]`)
- union constructors from 'easy-to-parse' unions

This is the hackiest, but also probably the most useful feature to date.


dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
```diff
@@ -0,0 +1,186 @@
...
+
+        values' | Const k <- kind', k `elem` [Kind, Sort] =
+                  fmap (shift 1 (V x 0)) $ insert x a' values
+                | otherwise = fmap (shift 1 (V x 0)) $ insert x holeExpr values
+
+    in buildCompletionContext' context' values' e'
+
+  -- fall back to annotated type if body doesn't type check; can't do 'dependent let'
+  | Just _A <- mA
+  , Right _ <- typeWithA absurd context _A =
```
Gabriel439: I don't think you need this extra type-checking step.  One of the soundness rules of Dhall is that an inferred type like `_A` is also safe to normalize and won't diverge

dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
```diff
@@ -0,0 +1,186 @@
...
+ where inc x (V y i) | x == y = V x (i + 1)
+                     | otherwise = V y i
+
+-- | Complete identifiers from the given completion context.
+completeFromContext :: CompletionContext -> [Completion]
+completeFromContext (CompletionContext context _) =
+  let context' = toList context
+      completeReserved keyword
+        | Right expr <- exprFromText "" keyword
+        , Right typ <- typeOf (fmap undefined expr) =
```
Gabriel439: Instead of `fmap undefined expr`, you can do:

```haskell
typeOf (do _ <- expr; holeExpr)
```

This is because `Expr` is a `Monad` where `>>=` replaces all occurrences of `Embed`

dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
```diff
@@ -0,0 +1,186 @@
...
+          Completion keyword (Just typ)
+        | otherwise = Completion keyword Nothing
+      reserved = map completeReserved $ HashSet.toList reservedIdentifiers
+  in [ Completion (pretty var) (if typ == holeExpr then Nothing else Just typ)
+     | (var, (_, typ)) <- zip (contextToVariables context') (toList context) ]
```
Gabriel439:
```suggestion
     | (var, (_, typ)) <- zip (contextToVariables context') context' ]
```

dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
```diff
@@ -0,0 +1,186 @@
...
+
+-- | Complete union constructors and record projections.
+completeProjections :: CompletionContext -> Expr Src X -> [Completion]
+completeProjections (CompletionContext context values) expr =
+  -- substitute 'dependent lets', necessary for completion of unions
+  let values' = toList values
+      subs = filter ((/= holeExpr) . snd) $ zip (contextToVariables values') (map snd values')
+      expr' = foldl (\e (x,val) -> subst x val e) expr subs
+
+  in case typeWithA absurd (traceShow (map fst $ toList context) $ context) expr' of
```
Gabriel439: Did you mean to keep this `traceShow`?

dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
```diff
@@ -0,0 +1,186 @@
...
+      Left _ -> []
+      Right _A ->
+        let expr'' = normalize expr'
+        in completeUnion expr'' expr'' ++ completeRecord (normalize _A)
```
Gabriel439: I think `completeUnion` only needs to take `expr''` as an argument once, if you change it to:

```
completeUnion _A@(Union m) = ...
```

dhall-lsp-server/src/Dhall/LSP/Backend/Parsing.hs
```diff
@@ -122,3 +130,112 @@ setSourcePos :: SourcePos -> Parser ()
...
+
+-- | An expression that is guaranteed not to typecheck. Can be used a
+-- placeholder type to emulate 'lazy' contexts, when typechecking something in a
+-- only partly well-typed context.
+holeExpr :: Expr s a
+-- The illegal variable name ensures that it can't be bound by the user!
+holeExpr = Var (V "" 0)
+
+-- | Approximate the type-checking context at the end of the input. Tries to
+-- parse as many binders as possible. Very messy!
```
Gabriel439: This function is deliciously evil :slightly_smiling_face:

dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
```diff
@@ -0,0 +1,186 @@
...
+
+-- | Complete file names.
+completeLocalImport :: FilePath -> FilePath -> IO [Completion]
+completeLocalImport relativeTo prefix = do
+  let dir = takeDirectory relativeTo </> takeDirectory prefix
+  exists <- doesDirectoryExist dir
+  if not exists
+  then return []
+  else do
+    files <- listDirectory dir
```
Gabriel439: `listDirectory` is a potentially very expensive operation if the user auto-completes the wrong directory (think: `/nix/store/<tab>`), so I suggest timing out this operation if it takes too long so that the user's editor can recover gracefully.

dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
EggBaconAndSpam: We need this here because the input is not necessarily well-typed. In particular, `_A` may not be a well-formed type.

dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
```diff
@@ -0,0 +1,186 @@
...
+  , Right kind <- typeWithA absurd context _A =
+    let _A' = normalize _A
+        kind' = normalize kind
+
+        context' = fmap (shift 1 (V x 0)) $ insert x _A' context
+
+        a' = normalize a
+
+        e' | Const k <- kind', k `elem` [Kind, Sort] =
+             subst (V x 0) a' e  -- 'dependent let'
```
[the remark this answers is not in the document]
EggBaconAndSpam: Yes!

dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
EggBaconAndSpam: Neat!

dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
EggBaconAndSpam: ups.

dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
EggBaconAndSpam: thanks!

dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
EggBaconAndSpam: No, I'm using the copy of the initial argument to complete the resulting type while recursing on the other, e.g. `< Left | Right>.Left` has type `<Left | Right>`.

dhall-lsp-server/src/Dhall/LSP/Backend/Completion.hs
EggBaconAndSpam: Hadn't thought of that, good point.

PR #1235, opened 2019-08-13
[dhall format] Prefer unquoted URLs

### LINKED ISSUE #1109: `dhall format` should prefer percent-encoded URL paths instead of quoted paths

philandstuff:
This updates `dhall format` to prefer unquoted URLs.

I included the test cases described in #1109, but along the way I
discovered a standard bug so I have opened dhall-lang/dhall-lang#704 to
fix it.  This change depends on that upstream fix.

Fixes #1109.


Gabriel439:
@philandstuff: Feel free to skip newly-added test cases that we haven't addressed yet

PR #1237, opened 2019-08-17
:help command in repl

### LINKED ISSUE #1236: :help command in repl
Hi all, loving the project! As a new dhall learner, I opened the repl and was presented with a cryptic turnstyle. The next thing I typed was "let x = 1" and got a strange error about unexpected end of input. I had no idea where to go from there. I think there are probably some easy wins here to help new users along!

I propose that we add a `:help` command and advertise it on repl startup. This could function like ghci :help by printing all available magic commands and their descriptions. Additionally, we can print simple information about particular repl quirks, like how let bindings need the magic :let command. I am imagining an experience like

```
$ dhall repl
Type :help for more information.
⊢ :help
Welcome to dhall repl vX.Y.Z!
Type any of some subset of expressions, blah blah, use the :let command blah blah.
The following commands are available:
:help - blah
:let - blah
:quit - blah
```

This seems pretty easy to add and I'd be happy to do so if we find agreement. Thanks!

ejconlon:
Resolves #1236 . Yields the following output:

```
$ stack exec dhall repl
Welcome to the Dhall v1.25.0 REPL! Type :help for more information.
⊢ :help
Type any expression to normalize it or use one of the following commands:
:help
    Print help text and describe options
:type EXPRESSION
    Infer the type of an expression
:hash EXPRESSION
    Hash the normalized value of an expression
:let IDENTIFIER = EXPRESSION
    Assign an expression to a variable
:clear
    Clear all bound variables
:load [FILENAME]
    Load bound variables from a file
:save [FILENAME | FILENAME = EXPRESSION]
    Save bound variables or a given expression to a file
:set OPTION
    Set an option. Currently supported: --explain
:unset OPTION
    Unset an option
:quit
    Exit the REPL
⊢
```

Happy to follow any process or style guidelines. Also happy to change the wording and formatting of the text.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ cabal.project.local~
 .HTF/
 .ghc.environment.*
 .cache
+.history
+stack.yaml.lock
--- a/dhall/src/Dhall/Repl.hs
+++ b/dhall/src/Dhall/Repl.hs
@@ -4,6 +4,7 @@
 {-# language NamedFieldPuns    #-}
 {-# language OverloadedStrings #-}
 {-# language RecordWildCards   #-}
+{-# language ScopedTypeVariables   #-}
 
 module Dhall.Repl
     ( -- * Repl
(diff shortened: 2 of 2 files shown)


dhall/src/Dhall/Repl.hs
```diff
@@ -4,6 +4,7 @@
 {-# language NamedFieldPuns    #-}
 {-# language OverloadedStrings #-}
 {-# language RecordWildCards   #-}
+{-# language ScopedTypeVariables   #-}
```
basile-henry: The rest of the language extensions look like they're aligned. This is most likely done using `stylish-haskell`, it might be worth a try! :)

dhall/src/Dhall/Repl.hs
```diff
@@ -398,26 +399,91 @@ cmdQuit _ = do
   liftIO (putStrLn "Goodbye.")
   liftIO (throwIO Interrupt)
 
+help
+  :: ( Haskeline.MonadException m, MonadFail m, MonadIO m, MonadState Env m )
+  => HelpOptions m -> [String] -> m ()
+help hs _ = do
+  liftIO (putStrLn "Welcome to the Dhall REPL.")
```
basile-henry: I think this would be better in the greeting than in the help output.

Maybe we could have a big Dhall in ASCII art?

```
  ____  _           _ _ 
 |  _ \| |__   __ _| | |
 | | | | '_ \ / _` | | |
 | |_| | | | | (_| | | |
 |____/|_| |_|\__,_|_|_|               
```
There are many other styles available here if that's something we might be interested in:
http://patorjk.com/software/taag/#p=display&f=Standard&t=Dhall

dhall/src/Dhall/Repl.hs
```diff
@@ -398,26 +399,91 @@ cmdQuit _ = do
...
+      "Clear all bound variables"
+      (dontCrash . clearBindings)
+  , HelpOption
+      "load"
+      "[FILENAME]"
+      "Load bound variables from a file"
+      (dontCrash . loadBinding)
+  , HelpOption
+      "save"
+      "[FILENAME]"
```
basile-henry: `:save` also support saving a specific expression. Would the following argument syntax work?

```suggestion
      "[FILENAME | FILENAME = EXPRESSION]"
```

dhall/src/Dhall/Repl.hs
```diff
@@ -506,8 +572,8 @@ completeFunc reversedPrev word
 
 
 greeter :: MonadIO m => m ()
-greeter =
-  return ()
+greeter = do
+  liftIO (putStrLn "Type :help for more information.")
```
basile-henry: Your issue mentions showing the version of Dhall on greeting. I think that's a great idea! :smile: 

Here's how we are getting the version of `dhall-haskell` for the command line interface if that helps:
https://github.com/dhall-lang/dhall-haskell/blob/8c3db821dbc0ad8cfd01d09a58e88126bda717ab/dhall/src/Dhall/Main.hs#L71
https://github.com/dhall-lang/dhall-haskell/blob/8c3db821dbc0ad8cfd01d09a58e88126bda717ab/dhall/src/Dhall/Main.hs#L359

dhall/src/Dhall/Repl.hs
ejconlon: Whoops I can delete this actually. Long story...

dhall/src/Dhall/Repl.hs
ejconlon: Moved to greeting with version. Personally, I prefer regular text so as not to increase scrollback too much.

.gitignore
```diff
@@ -20,4 +20,6 @@ cabal.project.local
 cabal.project.local~
 .HTF/
 .ghc.environment.*
-.cache
\ No newline at end of file
+.cache
+.history
```
ejconlon: `.history` comes from the repl and `stack.yaml.lock` comes from newer stack versions, I guess.

dhall/src/Dhall/Repl.hs
basile-henry: I saw the colourful ASCII art of the Unison language REPL yesterday and thought it was nice and fun. But I don't have any strong preferences for having one of those in the Dhall REPL.

PR #1311, opened 2019-09-13
Improve GHCJS support (cabal build, SHA256 hashing)

patrickmn:
This fixes the GHCJS build when using cabal, and adds SHA256 hashing support to GHCJS both when running in the browser and at compile time (THRunner on Node).

I have tested these changes using cabal on GHC and GHCJS, as well as etlas, but I could use some help to test nix/test suite/try-dhall. I would also love your blunt opinion on the organizational/style choices.

Motivation:
  - The `dhall` package doesn't compile using cabal/GHCJS (outside nix) due to the dependency on http-client (given -with-http) and cryptonite (breaking since, I think, 1.24.)
  - Importing anything with a SHA256 digest causes an error

Summary of changes:
  - Don't require `cryptonite`, `http-types`, `http-client`, or `http-client-tls` when compiling with GHCJS
  - Add `ghc-src` and `ghcjs-src` for platform-specific code to avoid a huge CPP mess
  - Add `ghc-src/Dhall/Crypto.hs` (using cryptonite) and `ghcjs-src/Dhall/Crypto.hs` (using JS FFI)
  - Split `src/Dhall/Import/HTTP.hs` into `ghc-src/Dhall/Import/HTTP.hs` and `ghcjs-src/Dhall/Import/HTTP.hs`
  - Add `src/Dhall/URL.hs` (for renderURL, used by both Import/HTTP modules)

Random thoughts:
  - Not happy about `typeof process` to detect Node vs. Browser, but it was the most reliable thing I could find (suggestions to look for CommonJS seem bad)
  - I don't think the use of `unsafePerformIO` in the GHCJS FFI call is necessary. You could just remove `IO` from the type signature, but that would make the intent less clear...

### DIFF
--- a/dhall/dhall.cabal
+++ b/dhall/dhall.cabal
@@ -449,7 +449,6 @@ Library
         cborg-json                                 < 0.3 ,
         containers                  >= 0.5.0.0  && < 0.7 ,
         contravariant                              < 1.6 ,
-        cryptonite                  >= 0.23     && < 1.0 ,
         data-fix                                   < 0.3 ,
         deepseq                                    < 1.5 ,
         Diff                        >= 0.2      && < 0.4 ,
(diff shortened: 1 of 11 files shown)


Gabriel439:
@patrickmn: I'm also adding you as a collaborator so that you can merge once CI Passes (either manually or by adding the `merge me` label to your pull request)

patrickmn:
Thank you both! `Dhall.Crypto (SHA256Hash(..), sha256DigestFromByteString, sha256Hash)` is exported now. Looks like the builds were successful.

Gabriel439:
@patrickmn: Yeah, you can go ahead and "Squash and merge" now, cleaning up the commit message if necessary

PR #1320, opened 2019-09-17
Warn about missing cache directory

jneira:
* As suggested by the [standard](https://github.com/dhall-lang/dhall-lang/blob/master/standard/imports.md).
* Not sure about the concrete message, feedback very welcome!
* Tested pointing `$XDG_CACHE_HOME` to a directory without read/write permissions
* The place where emitting the war is in each dhall execution, maybe it will be better to restrict to commands that couls use cache (most of them?)

### DIFF
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -107,6 +107,7 @@ module Dhall.Import (
     , hashExpression
     , hashExpressionToCode
     , writeExpressionToSemanticCache
+    , warnAboutMissingCache
     , assertNoImports
     , Status(..)
     , SemanticCacheMode(..)
@@ -202,6 +203,7 @@ import qualified Dhall.Pretty.Internal
 import qualified Dhall.TypeCheck
 import qualified System.Environment
 import qualified System.Info
+import qualified System.IO
 import qualified System.Directory                 as Directory
 import qualified System.FilePath                  as FilePath
 import qualified Text.Megaparsec
@@ -746,6 +748,26 @@ toHeader _ = do
 getCacheFile
     :: (Alternative m, MonadIO m) => FilePath -> Dhall.Crypto.SHA256Digest -> m FilePath
 getCacheFile cacheName hash = do
+    cacheDirectory <- getOrCreateCacheDirectory cacheName
+
+    let cacheFile = (cacheDirectory </> cacheName) </> ("1220" <> show hash)
+
+    return cacheFile
+
+warnAboutMissingCache :: (Alternative m, MonadIO m) => m ()
+warnAboutMissingCache = alternative₀ <|> alternative₁
+  where
+    alternative₀ = getOrCreateCacheDirectory "dhall" >> return ()
+
+    alternative₁ = liftIO (System.IO.hPutStrLn System.IO.stderr warning)
+
+      where warning = "It has not been possible to get a cache directory with read/write permission.\n"
+                   ++ "You can enable cache pointing the $XDG_CACHE_HOME environment variable\n"
+                   ++ "to a directory with the required permission.\n"
+                   ++ "A subdirectory named \"dhall\" will be created inside if it isn't exist.\n"
+
+getOrCreateCacheDirectory :: (Alternative m, MonadIO m) => FilePath -> m FilePath
+getOrCreateCacheDirectory cacheName = do
     let assertDirectory directory = do
             let private = transform Directory.emptyPermissions
                   where
@@ -774,16 +796,16 @@ getCacheFile cacheName hash = do
 
                     liftIO (Directory.setPermissions directory private)
 
-    cacheDirectory <- getCacheDirectory
+    cacheBaseDirectory <- getCacheBaseDirectory
 
-    assertDirectory (cacheDirectory </> cacheName)
+    let directory = cacheBaseDirectory </> cacheName
+    
+    assertDirectory directory
 
-    let cacheFile = (cacheDirectory </> cacheName) </> ("1220" <> show hash)
+    return directory
 
-    return cacheFile
-
-getCacheDirectory :: (Alternative m, MonadIO m) => m FilePath
-getCacheDirectory = alternative₀ <|> alternative₁
+getCacheBaseDirectory :: (Alternative m, MonadIO m) => m FilePath
+getCacheBaseDirectory = alternative₀ <|> alternative₁
   where
     alternative₀ = do
         maybeXDGCacheHome <- do
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -107,7 +107,7 @@ module Dhall.Import (
     , hashExpression
     , hashExpressionToCode
     , writeExpressionToSemanticCache
-    , warnAboutMissingCache
+    , warnAboutMissingCaches
     , assertNoImports
     , Status(..)
     , SemanticCacheMode(..)
@@ -148,6 +148,7 @@ import Control.Monad.Trans.State.Strict (StateT)
 import Data.ByteString (ByteString)
 import Data.CaseInsensitive (CI)
 import Data.List.NonEmpty (NonEmpty(..))
+import Data.Maybe (catMaybes)
 import Data.Semigroup (Semigroup(..))
 import Data.Text (Text)
 import Data.Void (absurd)
@@ -754,17 +755,27 @@ getCacheFile cacheName hash = do
 
     return cacheFile
 
-warnAboutMissingCache :: (Alternative m, MonadIO m) => m ()
-warnAboutMissingCache = alternative₀ <|> alternative₁
+warnAboutMissingCaches  :: (Alternative m, MonadIO m) => m ()
+warnAboutMissingCaches = do
+    warnings <- traverse warnAboutMissingCache ["dhall", "dhall-haskell"]
+
+    case catMaybes warnings of
+        warn:_ -> liftIO (System.IO.hPutStrLn System.IO.stderr warn)
+        []     -> return () 
+      
+
+warnAboutMissingCache :: (Alternative m, MonadIO m) => FilePath -> m (Maybe String)
+warnAboutMissingCache cacheName = alternative₀ <|> alternative₁
   where
-    alternative₀ = getOrCreateCacheDirectory "dhall" >> return ()
+    
+    alternative₀ = getOrCreateCacheDirectory cacheName >> return Nothing
 
-    alternative₁ = liftIO (System.IO.hPutStrLn System.IO.stderr warning)
+    alternative₁ = return (Just warning)
 
-      where warning = "It has not been possible to get a cache directory with read/write permission.\n"
+      where warning = "It has not been possible to get a cache directory \"" ++ cacheName ++ "\" with read/write permission.\n"
                    ++ "You can enable cache pointing the $XDG_CACHE_HOME environment variable\n"
                    ++ "to a directory with the required permission.\n"
-                   ++ "A subdirectory named \"dhall\" will be created inside if it isn't exist.\n"
+                   ++ "A subdirectory named \"" ++ cacheName ++ "\" will be created inside if it isn't exist.\n"
 
 getOrCreateCacheDirectory :: (Alternative m, MonadIO m) => FilePath -> m FilePath
 getOrCreateCacheDirectory cacheName = do
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -823,8 +823,8 @@ getOrCreateCacheDirectory showWarning cacheName = do
             catch (liftIO (Directory.doesDirectoryExist dir))
                   (handler "check the existence of" dir)
 
-    let existsPath path =
-            catch (liftIO (Directory.doesPathExist path))
+    let existsFile path =
+            catch (liftIO (Directory.doesFileExist path))
                   (handler "check the existence of" path)
 
     let createDirectory dir =
@@ -839,9 +839,9 @@ getOrCreateCacheDirectory showWarning cacheName = do
                     assertPermissions dir
 
                 else do
-                    existsPath' <- existsPath dir
+                    existsFile' <- existsFile dir
 
-                    if existsPath'
+                    if existsFile'
                         then do
                             let message =
                                      "The given path:\n"
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -773,7 +773,7 @@ warnAboutMissingCache cacheName = alternative₀ <|> alternative₁
     alternative₁ = return (Just warning)
 
       where warning = "It has not been possible to get a cache directory \"" ++ cacheName ++ "\" with read/write permission.\n"
-                   ++ "You can enable cache pointing the $XDG_CACHE_HOME environment variable\n"
+                   ++ "You can provide a cache directory by pointing the $XDG_CACHE_HOME environment variable\n"
                    ++ "to a directory with the required permission.\n"
                    ++ "A subdirectory named \"" ++ cacheName ++ "\" will be created inside if it isn't exist.\n"
 
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -70,7 +70,7 @@
     > $ export BAR='"Hi"'
     > $ export BAZ='λ(x : Bool) → x == False'
     > $ dhall <<< "{ foo = env:FOO , bar = env:BAR , baz = env:BAZ }"
-    > { bar : Text, baz : ∀(x : Bool) → Bool, foo : Integer }
+> { bar : Text, baz : ∀(x : Bool) → Bool, foo : Integer }
     >
     > { bar = "Hi", baz = λ(x : Bool) → x == False, foo = 1 }
 
@@ -140,7 +140,7 @@ module Dhall.Import (
 
 import Control.Applicative (Alternative(..))
 import Codec.CBOR.Term (Term(..))
-import Control.Exception (Exception, SomeException, toException)
+import Control.Exception (Exception, SomeException, IOException, toException)
 import Control.Monad (guard)
 import Control.Monad.Catch (throwM, MonadCatch(catch), handle)
 import Control.Monad.IO.Class (MonadIO(..))
@@ -748,7 +748,7 @@ toHeader _ = do
     empty
 
 getCacheFile
-    :: (Alternative m, MonadIO m) => FilePath -> Dhall.Crypto.SHA256Digest -> m FilePath
+    :: (MonadCatch m, Alternative m, MonadIO m) => FilePath -> Dhall.Crypto.SHA256Digest -> m FilePath
 getCacheFile cacheName hash = do
     cacheDirectory <- getOrCreateCacheDirectory cacheName
 
@@ -756,7 +756,7 @@ getCacheFile cacheName hash = do
 
     return cacheFile
 
-warnAboutMissingCaches  :: (Alternative m, MonadIO m) => m ()
+warnAboutMissingCaches  :: (MonadCatch m, Alternative m, MonadIO m) => m ()
 warnAboutMissingCaches = do
     warnings <- traverse warnAboutMissingCache ["dhall", "dhall-haskell"]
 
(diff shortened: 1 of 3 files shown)


dhall/src/Dhall/Import.hs
```diff
@@ -746,6 +748,26 @@ toHeader _ = do
...
+    cacheDirectory <- getOrCreateCacheDirectory cacheName
+
+    let cacheFile = cacheDirectory </> ("1220" <> show hash)
+
+    return cacheFile
+
+warnAboutMissingCache :: (Alternative m, MonadIO m) => m ()
+warnAboutMissingCache = alternative₀ <|> alternative₁
+  where
+    alternative₀ = getOrCreateCacheDirectory "dhall" >> return ()
```
Gabriel439: We should also warn if the `dhall-haskell` subdirectory does not exist, since it is used for the semi-semantic cache

dhall/src/Dhall/Import.hs
```diff
@@ -746,6 +748,26 @@ toHeader _ = do
...
     :: (Alternative m, MonadIO m) => FilePath -> Dhall.Crypto.SHA256Digest -> m FilePath
 getCacheFile cacheName hash = do
+    cacheDirectory <- getOrCreateCacheDirectory cacheName
+
+    let cacheFile = cacheDirectory </> ("1220" <> show hash)
+
+    return cacheFile
+
+warnAboutMissingCache :: (Alternative m, MonadIO m) => m ()
+warnAboutMissingCache = alternative₀ <|> alternative₁
```
Gabriel439: You should probably warn inside the `get*` functions whenever something goes wrong so that the warning can be more specific about the reason for failure.  If you warn at the top level then it's hard to say for sure why the cache failed, but if you warn at the point of failure then you can narrow down the warning to a permissions problem or a specific environment variable being missing.

jneira: Mmm, i had considered that alternative but then the warn will be emitted in each call to `getCacheFile` and afaiu it could be called several times in each dhall execution

jneira: There is basically three cases: no env var is set, the pointed dir doesnt exist or it has not the required permission. I could try to handle all three but keeping a top level warn.

jneira:
* I've added the warn for `dhall-haskell` but only if there is no one for `dhall` (they likely will fail at once)
  * Not sure if code is the most elegant, i am a little bit lost with monad transformers
* On being more precise with warnings:  i had thought in two alternatives:
  * Parametrize  `getOrCreateCacheDirectory`and `getCacheBasedirectory` functions (with a Bool?) to don't emit warnings when `getCacheFile` is used and add inline warnings within them
  * Change their signatures to some sort of `Either CacheError Path` (using EitherT?)
  * With both, to being precise, the code would be somewhat complex: the cause of missing cache could be a condition (missing env var or no appropiate permissions) or a `IOException` (in `getPermission`, `setPermission`, `createDirectory`) and we have to take in account that we could be creating missing parent directories

dhall/src/Dhall/Import.hs
```diff
@@ -746,6 +749,36 @@ toHeader _ = do
...
+warnAboutMissingCache :: (Alternative m, MonadIO m) => FilePath -> m (Maybe String)
+warnAboutMissingCache cacheName = alternative₀ <|> alternative₁
+  where
+    
+    alternative₀ = getOrCreateCacheDirectory cacheName >> return Nothing
+
+    alternative₁ = return (Just warning)
+
+      where warning = "It has not been possible to get a cache directory \"" ++ cacheName ++ "\" with read/write permission.\n"
+                   ++ "You can enable cache pointing the $XDG_CACHE_HOME environment variable\n"
```
sjakobi:
```suggestion
                   ++ "You can provide a cache directory by pointing the $XDG_CACHE_HOME environment variable\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -746,6 +749,36 @@ toHeader _ = do
...
+                   ++ "to a directory with the required permission.\n"
```
sjakobi:
```suggestion
                   ++ "to a directory with the required permissions.\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -746,6 +749,36 @@ toHeader _ = do
...
+                   ++ "A subdirectory named \"" ++ cacheName ++ "\" will be created inside if it isn't exist.\n"
```
sjakobi:
```suggestion
                   ++ "A subdirectory named \"" ++ cacheName ++ "\" will be created inside if it doesn't already exist.\n"
```

I think a user would already expect `dhall` to create directories in the cache directory – IMHO it doesn't need to be said explicitly. So maybe this sentence could simply be removed.

dhall/src/Dhall/Import.hs
```diff
@@ -746,6 +749,36 @@ toHeader _ = do
...
+
+warnAboutMissingCache :: (Alternative m, MonadIO m) => FilePath -> m (Maybe String)
+warnAboutMissingCache cacheName = alternative₀ <|> alternative₁
+  where
+    
+    alternative₀ = getOrCreateCacheDirectory cacheName >> return Nothing
+
+    alternative₁ = return (Just warning)
+
+      where warning = "It has not been possible to get a cache directory \"" ++ cacheName ++ "\" with read/write permission.\n"
```
sjakobi: I think it would be better if we would report the full patch of the cache directory here.

dhall/src/Dhall/Import.hs
jneira: My goal was users don't try to set the base cache dir like `set $XDG_CACHE_HOME=$HOME/.cache/dhall` although the env variable name suggests that it is a general dir for cache.

dhall/src/Dhall/Import.hs
jneira: Agree, but i will have to change the logic to make available the possible result of `getOrCreateCacheDirectory`. 
I'll give a try...

jneira:
Hi, i've tried to address the suggestion of @sjakobi and at same time i've separated the two main cases: env vars not informed and directories being not accesible.

I have to change the logic of `getOrCreateCacheDirectory`to make it ignore the possible `IOException`'s throwed by `Directory` actions:
* To follow the standars that says the implementation should not use the cache if the directory doesn't have the required permissions (and no throwing an error)
* To make the new warning makes sense at all! 

The output for first case (pretty unusual nowadays):
```console
D:\ws\dhall\dhall-haskell>set XDG_CACHE_HOME=

D:\ws\dhall\dhall-haskell>set LOCALAPPDATA=

D:\ws\dhall\dhall-haskell>echo 1 | dhall
It has not been possible to get a cache base directory from environment.
You can provide a cache base directory by pointing the $XDG_CACHE_HOME environment variable
to a directory with read and write permissions.

1
```

The output of second case:
```console
D:\ws\dhall\tests>set XDG_CACHE_HOME=D:\dirWithoutPermissions

D:\ws\dhall\tests>echo 1 | dhall
It has not been possible to get or create the default cache directory:
   D:\dirWithoutPermissions\dhall
Usually it is caused by permissions issues. You should make it readable and writable
or provide another cache base directory setting the $XDG_CACHE_HOME environment variable.

1
```

Trying to freeze a file (before this dhall throwed an error instead):
```console
D:\ws\dhall\tests>type test.dhall
./letrec.dhall

D:\ws\dhall\tests>dhall freeze --inplace test.dhall --all
It has not been possible to get or create the default cache directory:   D:\cache\dhall
Usually it is caused by permissions issues. You should make it readable and writable
or provide another cache base directory setting the $XDG_CACHE_HOME environment variable.

D:\ws\dhall\tests>type test.dhall
./letrec.dhall sha256:6c79d024292a8fa306f5906fa917079c48a87d5a5c7466fb9ba01690d636eaf5
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,8 +748,55 @@ toHeader _ = do
...
+warnAboutMissingCaches = do
+    warnings <- traverse warnAboutMissingCache ["dhall", "dhall-haskell"]
+
+    case catMaybes warnings of
+        warn:_ -> liftIO (System.IO.hPutStrLn System.IO.stderr warn)
+        []     -> return () 
+      
+
+warnAboutMissingCache :: (MonadCatch m, Alternative m, MonadIO m) => FilePath -> m (Maybe String)
+warnAboutMissingCache cacheName = do
```
jneira: It *have to be* another way to do it instead raw pattern matching (or using `maybe`) but i didn't get it :worried:

dhall/src/Dhall/Import.hs
```diff
@@ -744,8 +748,55 @@ toHeader _ = do
...
+                Maybe.runMaybeT (getOrCreateCacheDirectory cacheName)
+
+            case mbCacheDir of
+                Just _ -> return Nothing
+                
+                Nothing -> return (Just (warningCacheDir expectedCacheDir))
+
+        Nothing -> return (Just warningEnvVar)
+ 
+    where warningEnvVar =
```
Gabriel439: You can simplify this by pushing these warnings further down into the `get*` functions (i.e. push `warningEnvVar` into the `getCacheBaseDirectory` and pushing `warningCacheDir` into the `getOrCreateCacheDirectory` function).  Then you don't need separate top-level functions for warnings.

jneira: Mmm, my idea was not touch the `get*` function cause it is called by `getCacheFile` and the message would be shown several times (one for each hashed import, right?)

So those functions [have to change](https://github.com/dhall-lang/dhall-haskell/pull/1320#issuecomment-532600886), adding a parameter to activate warnings or returning an `Either Warning Path`. Not sure if it worths it, but what would be better? (or maybe i miss some another option)

Gabriel439: I think you could still create a `getOrCreateOrWarnCacheDirectory` utility which wraps `getOrCreateCacheDirectory`

jneira: Thanks for the suggestion!, this way i can leverage the `Alternative` class

jneira:
Much better this way, thanks again @Gabriel439

dhall/src/Dhall/Import.hs
```diff
@@ -774,16 +807,32 @@ getCacheFile cacheName hash = do
...
+    return directory
+
+getCacheBaseDirectoryOrWarn :: (Alternative m, MonadIO m) => m FilePath
+getCacheBaseDirectoryOrWarn = do
+    let message =
+            "\n" 
+         <> "\ESC[1;33mWarning\ESC[0m: " 
+         <> "It hasn't been possible get a cache base directory from environment.\n"
+         <> "\n"
+         <> "You can provide a cache base directory by pointing the $XDG_CACHE_HOME "
```
sjakobi:
```suggestion
         <> "You can provide a cache base directory by pointing the $XDG_CACHE_HOME\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,8 +746,39 @@ toHeader _ = do
...
+
+    let message =
+            "\n" 
+         <> "\ESC[1;33mWarning\ESC[0m: " 
+         <> "It hasn't been possible to get or create the default cache directory:\n" 
+         <> "\n"
+         <> "  " ++ cacheBaseDir </> cacheName ++ "\n"
+         <> "\n"
+         <> "Usually it is caused by permissions issues.\n"
+         <> "You should make it readable and writable or provide another cache base directory "
```
sjakobi:
```suggestion
         <> "You should make it readable and writable or provide another cache base directory by\n" 
```

dhall/src/Dhall/Import.hs
jneira: Mmm, i removed them cause they already went out  the 80 default col width of the msys console and i didnt like the aspect making the lines shorter.
But it is matter of taste...

sjakobi: Ah, never mind then!

jneira: Anyway i want to paste some screenshots to show how looks

jneira:
image

dhall/src/Dhall/Import.hs
sjakobi: Irregardless of the newline, the `by` should be added.

sjakobi:
The yellow of the `Warning` in the messages is slightly hard to read on light backgrounds, but I don't mind too much. Once the missing word above is sorted, I think this is fine to merge.

dhall/src/Dhall/Import.hs
```diff
@@ -744,8 +746,39 @@ toHeader _ = do
...
+    cacheBaseDir <- getCacheBaseDirectoryOrWarn
+
+    let message =
+            "\n" 
+         <> "\ESC[1;33mWarning\ESC[0m: " 
+         <> "It hasn't been possible to get or create the default cache directory:\n" 
+         <> "\n"
+         <> "  " ++ cacheBaseDir </> cacheName ++ "\n"
+         <> "\n"
+         <> "Usually it is caused by permissions issues.\n"
```
Gabriel439: I think we can still do more to push the error messages closer to the point of the error.  As a general rule of thumb if we can't confidently diagnose what went wrong then the errors aren't sufficiently localized

jneira: Mmm agree, i'll give another try, but imho sometimes the added complexity to get more precision for relatively infrequent errors doesn' worth it.
I will inline the warnigs inside the main functions adding a parameter to let `getCacheFile` ignore them.

jneira: Ok, i've made the inline of warnings within `get*` actions and it will warn about the "expected" errors (directory not having required perms or path pinting to a file instead a directory) and the possible `IOException`'s.
The flag to trigger the warnings is a `Bool` but if we establish a log system like proposed in #1328, it would change to a log level (= WARNING).

jneira:
I am afraid that in windows, for some cases, the code continue thowing an error (with no warnings) when trying to read or write the file in the cache directory. 
The cause is `Directory.getPermissions` does not work correctly and it returns that user has read/write permission when in fact it has not. When `dhall` or `dhall-haskell` exists but permissions are wrong the function `getOrCreateDirectory` return them happily and the error is throwed when effectively reading or writing the file inside.
But fix it could be in another pr...

dhall/src/Dhall/Import.hs
```diff
@@ -744,46 +746,139 @@ toHeader _ = do
...
-    let cacheFile = (cacheDirectory </> cacheName) </> ("1220" <> show hash)
+                            setPermissions dir
+    
+    cacheBaseDirectory <- getCacheBaseDirectory showWarning
 
-    return cacheFile
+    let directory = cacheBaseDirectory </> cacheName
+
+    let message =
+             "It hasn't been possible to get or create the default cache directory:\n"
```
Gabriel439:
```suggestion
             "Could not get or create the default cache directory:\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,46 +746,139 @@ toHeader _ = do
...
+          <> "\n"
+          <> "  " <> directory <> "\n"
+          <> "\n"
+          <> "You can enable cache creating it if needed and setting read, write and search\n"
```
Gabriel439:
```suggestion
          <> "You can enable caching by creating it if needed and setting read, write and search\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -812,6 +907,18 @@ getCacheDirectory = alternative₀ <|> alternative₁
 
         where isWindows = System.Info.os == "mingw32"
 
+    alternative₂ = do
+        let message =
+                "\n"
+             <> "\ESC[1;33mWarning\ESC[0m: "
+             <> "It hasn't been possible get a cache base directory from environment.\n"
```
Gabriel439:
```suggestion
             <> "Could not locate a cache base directory from the environment.\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,46 +746,139 @@ toHeader _ = do
...
+                  <> "\ESC[1;33mWarning\ESC[0m: "
+                  <> message
+
+            when showWarning (liftIO (System.IO.hPutStrLn System.IO.stderr warning))
+
+            empty
+
+    let handler action dir (ioex :: IOException) = do
+            let ioExMsg =
+                     "when trying to " <> action <> "\n"
```
Gabriel439:
```suggestion
                     "When trying to " <> action <> ":\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,46 +746,139 @@ toHeader _ = do
...
+                  <> "\n"
+                  <> "  " <> dir <> "\n"
```
Gabriel439:
```suggestion
                  <> "↳ " <> dir <> "\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,46 +746,139 @@ toHeader _ = do
...
+                  <> "\n"
+                  <> "the following IO exception was thrown:\n"
```
Gabriel439:
```suggestion
                  <> "... the following exception was thrown:\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,46 +746,139 @@ toHeader _ = do
...
-                then do
-                    permissions <- liftIO (Directory.getPermissions directory)
+            if accessible permissions
+                then
+                    return ()
+                else do
+                    let message =
+                             "the directory:\n"
+                          <> "\n"
+                          <> "  " <> dir <> "\n"
```
Gabriel439:
```suggestion
                          <> "↳ " <> dir <> "\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,46 +746,139 @@ toHeader _ = do
...
+                          <> "\n"
+                          <> "has not read, write and search permissions.\n"
+                          <> "Its current permissions are:\n"
```
Gabriel439:
```suggestion
                          <> "The directory's current permissions are:\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,46 +746,139 @@ toHeader _ = do
...
-                    assertDirectory (FilePath.takeDirectory directory)
+                    existsPath' <- existsPath dir
 
-                    liftIO (Directory.createDirectory directory)
+                    if existsPath'
+                        then do
+                            let message =
+                                     "the given path:\n"
+                                  <> "\n"
+                                  <> "  " <> dir <> "\n"
```
Gabriel439:
```suggestion
                                  <> "↳ " <> dir <> "\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,46 +746,139 @@ toHeader _ = do
...
 
                 else do
-                    assertDirectory (FilePath.takeDirectory directory)
+                    existsPath' <- existsPath dir
 
-                    liftIO (Directory.createDirectory directory)
+                    if existsPath'
+                        then do
+                            let message =
+                                     "the given path:\n"
```
Gabriel439:
```suggestion
                                     "The given path:\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,46 +746,139 @@ toHeader _ = do
...
 
-            if directoryExists
-                then do
-                    permissions <- liftIO (Directory.getPermissions directory)
+            if accessible permissions
+                then
+                    return ()
+                else do
+                    let message =
+                             "the directory:\n"
```
Gabriel439:
```suggestion
                             "The directory:\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,46 +746,139 @@ toHeader _ = do
...
+
+    let handler action dir (ioex :: IOException) = do
+            let ioExMsg =
+                     "when trying to " <> action <> "\n"
+                  <> "\n"
+                  <> "  " <> dir <> "\n"
+                  <> "\n"
+                  <> "the following IO exception was thrown:\n"
+                  <> "\n"
+                  <> "  " <> show ioex <> "\n"
```
Gabriel439:
```suggestion
                  <> "↳ " <> show ioex <> "\n"
```

dhall/src/Dhall/Import.hs
```diff
@@ -744,46 +746,139 @@ toHeader _ = do
...
+    
+    cacheBaseDirectory <- getCacheBaseDirectory showWarning
 
-    return cacheFile
+    let directory = cacheBaseDirectory </> cacheName
+
+    let message =
+             "It hasn't been possible to get or create the default cache directory:\n"
+          <> "\n"
+          <> "  " <> directory <> "\n"
```
Gabriel439:
```suggestion
          <> "↳ " <> directory <> "\n"
```

jneira:
@Gabriel439 @sjakobi Thanks both for your suggestions and patience 😉
And sorry for my poor english skills... i am always learning that part too

jneira:
Wow, bitten again in hydra by `directory-1.2.2.0`... fixing

PR #1330, opened 2019-09-20
Different with-http CPP check; fix remote imports in GHCJS

patrickmn:
I submitted https://github.com/dhall-lang/dhall-haskell/pull/1311 before studying `Import.hs` carefully enough, and my GHCJS build worked perfectly (with local imports, doh!) so I didn't notice.

Since I removed the GHCJS dependency on http-client, the CPP check for `MIN_VERSION_http_client` no longer works. I noticed the existence of `-DWITH_HTTP` in the tasty test suite and added that at the library level. (I think it's clearer too.)

Also, fortunately, `Manager` is not actually needed in Import.hs, just the check for `WITH_HTTP`.

### DIFF
--- a/dhall/dhall.cabal
+++ b/dhall/dhall.cabal
@@ -429,6 +429,9 @@ Library
       Build-Depends: semigroups == 0.18.*
       Build-Depends: transformers == 0.4.2.*
       Build-Depends: fail == 4.9.*
+    if flag(with-http)
+      CPP-Options:
+        -DWITH_HTTP
     if impl(ghcjs)
       Hs-Source-Dirs: ghcjs-src
       Build-Depends:
--- a/dhall/doctest/Main.hs
+++ b/dhall/doctest/Main.hs
@@ -25,7 +25,8 @@ main = do
         writeFile "file1" "./file2"
 
         Test.DocTest.doctest
-            [ "--fast"
+            [ "-DWITH_HTTP"
+            , "--fast"
             , "-i" <> (prefix </> "src")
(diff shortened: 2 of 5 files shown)


patrickmn:
Doctest again. Revisiting tomorrow.

patrickmn:
Btw stuck on this trying to enter nix shell to run doctest locally:

```
...
5pfkp8gqca-source/pkgs/development/haskell-modules/make-package-set.nix:75:37, called from /nix/store/xz3aw1hswykigk6ivipisj5pfkp8gqca-source/pkgs/development/haskell-modules/make-package-set.nix:115:28:
while evaluating 'makeOverridable' at /nix/store/xz3aw1hswykigk6ivipisj5pfkp8gqca-source/lib/customisation.nix:72:24, called from /nix/store/xz3aw1hswykigk6ivipisj5pfkp8gqca-source/pkgs/development/haskell-modules/make-package-set.nix:98:8:
while evaluating 'drvScope' at /nix/store/xz3aw1hswykigk6ivipisj5pfkp8gqca-source/pkgs/development/haskell-modules/make-package-set.nix:88:18, called from /nix/store/xz3aw1hswykigk6ivipisj5pfkp8gqca-source/lib/customisation.nix:74:12:
moving build output '/nix/store/x8sk4h2ylq3k0xmawqh8v8bar2ljx8mg-cabal-dedupe-src' from the sandbox to the Nix store: Permission denied
```

patrickmn:
Looks like that worked. Thank you :)

PR #1340, opened 2019-09-24
Make sure `dhall-nix` is included in releases

jhenahan:
Saw on Slack that the Darwin release was missing for `dhall-to-nix`, and from what I've dug up it looks like this should fix that.


jhenahan:
@sjakobi `fpchat`! Invite if you're not already there: https://fpchat-invite.herokuapp.com/

Gabriel439:
@jhenahan: Thank you!  For some reason I thought `dhall-nix` couldn't be build on OS X

jhenahan:
I was mainly going off the Travis build log which looked clean, and I know I've built it at least once on Darwin (under Nix, at least). Looks like the build's green, still, so I'm hopeful I haven't just helped put a wildly broken binary out into the world. 😆 Thanks for merging!

PR #1357, opened 2019-09-30
Allow haskell-lsp 0.17.

quasicomputational:
This is related to #1249, since earlier versions don't support GHC
8.8.


quasicomputational:
Oops - yep. Fenceposts are hard.

PR #1369, opened 2019-10-02
Fix `dhall-lsp-server` to specify a UTF8 locale

Gabriel439:
Related to https://github.com/dhall-lang/dhall-haskell/issues/1356#issuecomment-536840612


Gabriel439:
I'll go ahead and merge this as I'm reasonably confident this fixes the problem based on it fixing similar issues in #1396

PR #1434, opened 2019-10-18
Do not warn about cache on --no-cache.

StevenXL:
First time contributor to the project. (Trying to get my Hacktoberfest PRs in!)

This does not address the main point of #1404 but it is a point brought up in that issue.

P.S. I am partial to the use of `when` and `not`  instead of `unless`, but not a big deal if maintainers would like to change that.


StevenXL:
> Welcome and cheers! :)

Thank you - I just realized why this issue was brought up in #1404 - once we add the `--no-cache` flag to the payload of `Resolve` and `Type`, the code here will have to change; luckily, the change is minor and I will get to it.

Issue #1472, opened 2019-10-24
Should dhall-to-yaml quote special strings?


amarrella:
Hi!

Up until dhall-1.26.1 / dhall-json 1.4.1 when doing 
```sh
echo '{foo = "*"}' | ./dhall-to-yaml
```
this generated
```yaml
foo: '*'
```

With dhall-1.27.0 / dhall-json 1.5.0 this generates
```yaml
foo: *
```

I think the YAML spec is unclear on this case, but this breaks a bunch of stuff, for example kubernetes crontab definitions.

I know i can use the `--quoted` flag but this behaviour is a bit surprising, and the quoted flag quotes everything in the yaml file making visual inspection a bit harder. Would it be possible to revert the behaviour back to 1.26.1? (not sure if this is a dhall issue or a dependency issue)

Thanks in advance :)

Gabriel439:
@amarrella: How are you building and/or installing the `dhall-json` executable?  The reason I ask is that `dhall-json` can now be built with one of two dependencies (`aeson-yaml` with a BSD license or `HsYAML` with a GPLv3 license) depending on the presence or absence of an `-fgpl` `cabal configure` flag, and I want to reverse engineer which dependency your `dhall-json` was built against using so we know what needs to be fixed.

amarrella:
@Gabriel439: I'm using the binary from the release page. Both linux and osx version seem to have the same behaviour

Gabriel439:
Okay, so those are both built with `HsYAML`.  Let me see how to fix this

amarrella:
Thanks! Meanwhile a workaround seems to be putting escaped quotes inside the string
```sh
❯ echo '{foo = "\"*\""}' | dhall-to-yaml
foo: "*"
```

Gabriel439:
Another workaround at the moment is that if you build against the `aeson-yaml` dependency (i.e. without the `-fgpl` `cabal configure` flag) then it also gives the right behavior:

```
$ dhall-to-yaml <<< '{ foo = "*" }'
foo: "*"
```

Gabriel439:
Fix is up here: https://github.com/dhall-lang/dhall-haskell/pull/1474

amarrella:
Awesome! Thanks again @Gabriel439 ! :)

Gabriel439:
@amarrella: You're welcome! 🙂

PR #1474, opened 2019-10-25
Fix `dhall-to-yaml` to quote special strings

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/1472

This also refactors the code to a form that was easier for me to understand


amarrella:
@sjakobi: the best info that I found is that the parser treats it as an `alias indicator`: https://yaml.org/refcard.html

PR #1487, opened 2019-10-29
Flake demo

ehmry:
I've been playing around with nix flakes (https://github.com/NixOS/rfcs/pull/49) and it was not difficult to make a flake out of dhall-haskell. The biggest change I made was to convert the dhall-lang submodule into a subtree. This makes it possible to build the repo without a recursive checkout and in pure mode, it is also possible fetch from a Gitub archive rather than do a Git clone.

The disadvantage is that this introduces a subtree workflow, which I think most people are not familiar with (I am not).

Using the flake is quite simple once the `nixFlake` version of Nix is available:
```shell
nix flake add dhall-haskell 'git+https://github.com/dhall-lang/dhall-haskell.git?ref=flake'
  # add Dhall to local flake registry

nix build dhall-haskell # build the "dhall" utility
nix flake check dhall-haskell # run Dhall coverage test
```

Nix flakes are still experimental, I've created this branch simply as a test and demo.


Gabriel439:
@ehmry: We would need to keep the `dhall-lang` subdirectory as a `git submodule` since we need to periodically synchronize it with https://github.com/dhall-lang/dhall-lang

basile-henry:
My understanding of nix flakes is that they can depend on other flakes. As such, wouldn't it make more sense to also add nix flakes to https://github.com/dhall-lang/dhall-lang and depend on that instead of using a git subtree?

ehmry:
A flake here is now redundant as the current version of dhall-haskell is in nixpkgs master, and nixpkgs master is now a flake.

PR #1613, opened 2019-12-20
Allow context and normalizer to be passed to various main entrypoints

brendanhay:
In order to usefully extend Dhall with new builtins we must currently reimplement `dhall`, `dhall-json`, `dhall-yaml`, and any other CLI to whatever degree you can stomach. In order to make it somewhat more realistic to extend the starting context and normalizer used in all of these binaries I've done a pass over the main entrypoints to offer up `*With` variants in the spirit of `Dhall.Import.loadWith`, `Dhall.TypeCheck.typeWith`, and `Dhall.Normalize.normalizeWith`. 

Additionally the entrypoints are now more consistently 'importable' into your own potentially monolithic binary that could in theory tie all of the command-line parsers together and call the appropriate `mainWith` library function.

To ensure backwards compatibility all of the existing functions have not been modified and simply call the respective `*With` variant where necessary.

* `dhall:Dhall.Main`: `commandWith`
* `dhall:Dhall.Diff`: `diffWith`
* `dhall-json:Dhall.JSON`: `dhallToJSONWith`, `convertToHomogeneousMapsWith`, `codeToValueWith`
* `dhall-json:Dhall.JSON.Yaml`: `dhallToYamlWith`
* `dhall-to-json` now has an identical structure to `dhall-to-yaml` with `Main.hs` living in the library proper. This allows trivial reimpelmentation of both the `dhall-json` and `dhall-yaml` binaries.

Overall this is kinda gross. The implicit empty context and normalizer in `Dhall.Import.loadRelativeTo`, `Dhall.Normalize.normalize`, and `Dhall.TypeCheck.typeOf` leads to basically every externally useful bit of functionality being either difficult or downright impossible to extend. It sort of smells like all of this is pointing towards a more explicit interface that requires you to specify context/normalizer in order to ensure proper extension, otherwise you cannot guarantee that something in the bowels of Dhall isn't calling the implicit variants and ignoring your customizations.


brendanhay:
I'm throwing this up here to just prove a point - that it is in fact unrealistic to currently extend Dhall beyond simple normalization and type-checking, as you either lose too much of the good stuff from any existing Dhall workflow, or you end up maintaining essentially a fork of `dhall`-the-binary and company.

dhall/src/Dhall/Main.hs
```diff
@@ -813,4 +837,4 @@ command (Options {..}) = do
 main :: IO ()
 main = do
     options <- Options.Applicative.execParser parserInfoOptions
-    command options
+    commandWith Dhall.Context.empty Nothing options
```
Gabriel439:
```suggestion
    command options
```

... since you already defined `command` for this purpose

dhall-json/src/Dhall/JSON/Yaml.hs
```diff
@@ -64,22 +66,41 @@ parseQuoted =
...
+    -- ^ The source file path. If no path is given, imports
+    -- are resolved relative to the current directory.
+    -> Text
+    -- ^ Input text.
+    -> IO ByteString
+dhallToYaml mpath =
+    dhallToYamlWith Dhall.Context.empty Nothing mpath
+
+dhallToYamlWith
+    :: _
```
Gabriel439: Did you mean to leave these type holes here?

Gabriel439:
@brendanhay: Sorry, I missed your comment that this was just a proof of concept about the difficulty of accurately threading things around.

One idea I had to deal with the issue of internal calls accidentally using the implicit versions is to use type classes instead of explicit context/normalizer passing, in order to ensure coherence.  In other words, you would have something like:

```haskell
class Normalizable m expr where
    normalizer :: expr s a -> m (Maybe (expr s a))
```

That would ensure that you always consistently use the right normalizer no matter where you are in the code base (and it would reduce the amount of dictionary passing as a convenience)

dhall-json/src/Dhall/JSON/Yaml.hs
brendanhay: No - I was waiting for `nix-build` to enlighten me via `stdout`, but apparently it never came

brendanhay:
I'm happy to tidy this up as appropriate if it looks like a worthwhile interim solution, otherwise I'm also happy to attempt a stab at implementing the type-class based approach.

Gabriel439:
@brendanhay: I think the typeclass approach is more promising

PR #1659, opened 2020-02-06
Added support for substitutions in inputExpr/inputExprWithSettings

mmhat:
Support for substitutions was somehow missed in the original
implementation. This commit fixes that.

In order to avoid code duplication common parts of inputWithSettings and
inputExprWithSettings were moved to a new helper function inputHelper.


mmhat:
@Gabriel439 Was there a particular reason for the code duplication of `inputWithSettings` and `inputExprWithSettings`?

Gabriel439:
@mmhat: No particular reason.  These functions evolved over time from multiple contributors so the duplication might have been missed

PR #1663, opened 2020-02-07
Append newline to REPL expressions

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/1655


Gabriel439:
@sjakobi: You're welcome! 🙂

PR #1664, opened 2020-02-09
Support template haskell for multiple datatypes

Gabriel439:
Inspired by this discussion: https://discourse.dhall-lang.org/t/makehaskelltypefromunion-and-fields-with-sum-types/150

This adds a new `makeHaskellTypes` utility which can generate multiple
types which can refer to each other.  The trick is that the type
compares nested Dhall types against top-level ones and replaces them
with Haskell datatype references when the types are judgmentally
equivalent.

### DIFF
--- a/dhall/src/Dhall/TH.hs
+++ b/dhall/src/Dhall/TH.hs
@@ -1,4 +1,7 @@
+{-# LANGUAGE DeriveTraversable #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternGuards     #-}
+{-# LANGUAGE RecordWildCards   #-}
 {-# LANGUAGE TemplateHaskell   #-}
 
 -- | Template Haskell utilities
@@ -6,6 +9,8 @@ module Dhall.TH
     ( -- * Template Haskell
       staticDhallExpression
     , makeHaskellTypeFromUnion
+    , makeHaskellTypes
+    , HaskellType(..)
     ) where
 
 import Data.Monoid ((<>))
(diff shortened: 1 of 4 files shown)


srid:
@Gabriel439 I'm using this to see how it works in my code base. `makeHaskellTypes` successfully generates Haskell types for the 3 Dhall types (see the PR linked above). 

However, as briefly mentioned in Discourse, the only problematic parts are newtypes ... specifically newtypes over non-Dhall types. 

The [original types](https://github.com/srid/website/pull/14/commits/a90c457000cab3ad866a23f653bb3cd37a7d0d40#diff-6da90772fd4ca97e2718d288c0c277e6L51-L66) use two newtypes -- `Markdown` and `Day` -- which are both over a Haskell library type. Originally the builtin type `Text` is used in the Dhall type def, which are loaded using the `string` parser in [their `FromDhall` instances](https://github.com/srid/website/blob/45b6cd0895661643246f613fdac1b16c555b724b/src/Website/Tidbit.hs#L33-L37), which in turn converts the `Text` to `Markdown` or `Day` (using MMark.parse and parseTimeOrError respectively).

I also think needing to define parallel types that differ only in some fields using newtypes beats the point of generating the original type using TH. I don't know what a good solution to this would be. We could hack the arguments to the TH function to specify custom types for record fields, but this won't map well to sum type constructor arguments. Unfortunately newtypes are unavoidable if we are to parse some well-known types like dates.

Gabriel439:
@srid: This is a pretty common issue for IDLs in general (e.g. protocol buffers, thrift, etc.).  Usually the generated code doesn't exactly match what you want and you have to perform some additional conversions to the actual types you want to use or settle for the default type

PR #1666, opened 2020-02-11
Add `cachix` step to `./scripts/release.sh`

Gabriel439:
See: https://discourse.dhall-lang.org/t/cookbook-request-simple-dhall-pinning-with-nix-shell/147


ari-becker:
Wonderful, thank you for adding this!

Two points of consideration:
1. What about OS X binaries?
2. Adding documentation around `dhall.cachix.org` to the `README`?

I wouldn't want either one to hold up merging this in... if you decide to merge this in first then let's just create issues for them so that they don't fall off the map.

Gabriel439:
@ari-becker: I just uploaded OS X binaries to `cachix` for version 1.30.0 and updated the `README` with instructions for using that cache

PR #1681, opened 2020-02-21
Fix numeric parsing for GHCJS

Gabriel439:
The fix from #1647 broke GHCJS support for Dhall.  Specifically, any Dhall
code containing numeric literals would fail to parse when using GHCJS.

The consequence of this is that the final example from the online demo
currently does not work because it uses at least one number.

The root cause was this issue:

https://github.com/ghcjs/ghcjs-base/issues/62

The change to use `Text.Megaparsec.Char.Lexer.float` triggered the above
problem, because `float` depended transitively on `Data.Char.toTitle` which
was not working for the version of GHCJS that we currently build against.

The fix is to reimplement the logic for parsing double literals by
following the standard grammar instead of reusing an existing utility.
However, I cribbed quite a bit from the existing logic for
`Text.Parser.Char.double` along the way.

### DIFF
--- a/dhall/src/Dhall/Parser/Token.hs
+++ b/dhall/src/Dhall/Parser/Token.hs
@@ -123,13 +123,16 @@ import qualified Control.Monad
 import qualified Data.Char                  as Char
 import qualified Data.Foldable
 import qualified Data.HashSet
+import qualified Data.List                  as List
 import qualified Data.List.NonEmpty
+import qualified Data.Scientific            as Scientific
 import qualified Data.Text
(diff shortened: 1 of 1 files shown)


Gabriel439:
@sjakobi: I can add a comment, but don't think we can enable the tests.  I left a comment at one point saying that the test suite doesn't work on GHCJS:

https://github.com/dhall-lang/dhall-haskell/blob/91696782fd7d510102a23b2e39c96dacee0230b9/nix/shared.nix#L91

... although I don't remember the exact reason why

PR #1689, opened 2020-02-27
Fix typos in manpage

vmchale:
This fixes a couple irritation in the `dhall` manpage.


Gabriel439:
@vmchale: Thank you! 🙂

PR #1696, opened 2020-03-04
From/ToDhall no longer takes InterpretOptions argument

akrmn:
Coming from `aeson`, I wanted to define `FromDhall` instances for my types which built on `GHC.Generics` but with tweaked options (say, drop prefixes). Ideally, I would write something like

```haskell
{- dhall type: { Name : Text, Email : Text }
-}
data Person = Person
  { personName :: Text
  , personEmail :: Text
  } 
  deriving (Generic)

instance FromDhall Person where
  autoWith = genericAutoWith defaultInterpretOptions 
    { fieldModifier = Text.drop (Text.length "person") 
    }
```

The problem with this is that `autoWith` itself takes a `InterpretOptions` argument, which is set to `defaultInterpretOptions` by `auto`, the most common entry point. This means that the only way to enforce a canonical `FromDhall` instance for a type is by ignoring the argument to `autoWith`, which is confusing for callers. In practice, this PR is only reflecting in the types the fact that all instances of `FromDhall` and `ToDhall` necessarily ignore the `InterpretOptions`, either because they don't need it or because they overwrite it. The only field that was actually used, `inputNormalizer`, was kept as an input.

- `InterpretOptions` split into smaller `InterpretOptions` record and `InputNormalizer` newtype.
- `FromDhall` method `autoWith` and `ToDhall` method `injectWith` no longer take `InterpretOptions` arguments, instead, they take an `InputNormalizer`, since the only field that instances of these classes ever used was `inputNormalizer` (only used by instances for `(->)`).
-  `GenericFromDhall` method `genericAutoWithNormalizer` (renamed from `genericAutoWith`) and `GenericToDhall` method `genericToDhallWithNormalizer`(renamed from `genericToDhallWith`) now take an `InputNormalizer` in addition to an `InterpretOptions`.

### DIFF
--- a/dhall/src/Dhall.hs
+++ b/dhall/src/Dhall.hs
@@ -75,6 +75,7 @@ module Dhall
     , list
     , vector
     , function
+    , functionWith
     , setFromDistinctList
     , setIgnoringDuplicates
     , hashSetFromDistinctList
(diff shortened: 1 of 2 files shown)


akrmn:
Awesome! Thanks a lot, Gabriel

Gabriel439:
@akrmn: You're welcome! 🙂

patrickmn:
@akrmn Catching a codebase that used a custom fieldModifier up with this change, and it does not appear that `defaultInputNormalizer` is exported, which I need for `autoWith` (right?)

Gabriel439:
@patrickmn: Yeah, it's not currently exported, but it's not hard to locally define:

https://hackage.haskell.org/package/dhall-1.31.0/docs/src/Dhall.html#defaultInputNormalizer

patrickmn:
@Gabriel439 for that I'd need the newtype to be exported, right?

Or, in other words, how would I externally make an instance with a custom field modifier? (I can't use DerivingVia.)

Gabriel439:
@patrickmn: The newtype is exported: https://hackage.haskell.org/package/dhall-1.31.0/docs/Dhall-Core.html#t:ReifiedNormalizer

Gabriel439:
@patrickmn: Oh sorry, wrong newtype.  Never mind

Gabriel439:
@patrickmn: I have a fix up here: https://github.com/dhall-lang/dhall-haskell/pull/1727

I can also cut a 1.31.1 point release for this

patrickmn:
Awesome, will finish my port as soon as it's merged just in case there's anything else pre-1.31.1.

patrickmn:
All good. Thank you!

Gabriel439:
@patrickmn: You're welcome! :slightly_smiling_face:

Issue #1711, opened 2020-03-25
Invalid Dhall.Decoder based on order of ADT constructors


SiriusStarr:
Attempting to import the following Dhall:
```
λ(Recursive : Type)
    → let RecursiveF =
            < AF : Recursive
            | BF
            | CF
            | DF
            | WorksIfFirstF :  Text
            | WorksIfSecondF : Text
            | EF
            >
      in    λ(Fix : RecursiveF → Recursive)
          → Fix (RecursiveF.WorksIfSecondF "" )
```
into the Haskell type:
```
data Recursive
  = A Recursive
  | B
  | C
  | D
  | WorksIfSecond NonemptyText
  | WorksIfFirst NonemptyText
  | E
  deriving (Generic, Show)
```
produces the error:
```
Error: Invalid Dhall.Decoder                                               
                                                                                
Every Decoder must provide an extract function that succeeds if an expression   
matches the expected type.  You provided a Decoder that disobeys this contract  
                                                                                
The Decoder provided has the expected dhall type:                               
                                                                                
↳ < WorksIfFirstF : Text | EF >
                                                                                
and it couldn't extract a value from the well-typed expression:                 
                                                                                
↳ < AF : _@1
  | BF
  | CF
  | DF
  | EF
  | WorksIfFirstF : Text
  | WorksIfSecondF : Text
  >.WorksIfSecondF
    ""
```

This error does not occur if any of `A`, `B`, `C`, `D`, or `E` are removed from both the Haskell and Dhall, or if `WorksIfFirst` and `WorksIfSecond` are flipped in the Haskell.  In any of these cases, the expected "failed extraction" error occurs (as intended).

Relevant Haskell instances below:
```
newtype NonemptyText = NonemptyText {getNonemptyText :: Text} deriving (Eq, Ord, Show)

validateText :: Decoder Text -> Decoder NonemptyText
validateText (Decoder extractIn expectedIn) = Decoder extractOut expectedIn
  where
    extractOut e = case extractIn e of
      Success t ->
        if T.null t
          then extractError "Text cannot be empty."
          else Success $ NonemptyText t
      Failure f -> Failure f

instance FromDhall NonemptyText where
  autoWith _ = validateText strictText

data Recursive
  = A Recursive
  | B
  | C
  | D
  | WorksIfFirst NonemptyText
  | WorksIfSecond NonemptyText
  | E
  deriving (Generic, Show)

TH.makeBaseFunctor ''Recursive

[...]

PR #1720, opened 2020-03-28
Improve error message for failed union extractions:

Gabriel439:
Related to https://github.com/dhall-lang/dhall-haskell/issues/1711

The derived `FromDhall` instance for unions used `ealt` to combine
results from the left and right "halves" of a union, and `ealt`
essentially prefers the last error if there is more than one.

This leads to confusing error messages when combining unions with
custom decoders.  For example, using the example from:

https://github.com/dhall-lang/dhall-haskell/issues/1711#issuecomment-605516095

... gives the following error message:

```
Error: Invalid Dhall.Decoder

Every Decoder must provide an extract function that succeeds if an expression
matches the expected type.  You provided a Decoder that disobeys this contract

The Decoder provided has the expected dhall type:

↳ < C | D >

and it couldn't extract a value from the well-typed expression:

↳ < A : List {} | B | C | D >.A [ {=}, {=} ]
```

... because the decoder splits the union into two smaller unions:

* `< A : List {} | B >`
* `< C | D >`

... and then tries to decode the union constructor using both smaller
unions.  Both halves fail, but in different ways:

* The first half fails due to Haskell-side validation for the expected
  `Set` that rejects the `List` since it contains duplicates

* The second half fails due the constructor not matching the union

... but only the error message for the former half is preserved.
(I'm oversimplifying a bit, but this is the basic idea)

After this change, both error messages are preserved and you now get:

```
Test.hs: Multiple errors were encountered during extraction:

Error: Failed extraction

The expression type-checked successfully but the transformation to the target
type failed with the following error:

One duplicate element in the list: ()

Error: Invalid Dhall.Decoder

Every Decoder must provide an extract function that succeeds if an expression
matches the expected type.  You provided a Decoder that disobeys this contract

The Decoder provided has the expected dhall type:

↳ < C | D >

and it couldn't extract a value from the well-typed expression:

↳ < A : List {} | B | C | D >.A [ {=}, {=} ]
```

It's still not ideal, but at least the error message that the user was
expecting is preserved, even if it is polluted by the error message for
the other half of the union.

To further improve the error message we'd need to refactor the
`ExtractError`s type to distinguish between union-decoding errors
and custom-decoder errors and give priority to the latter errors.

### DIFF
--- a/dhall/src/Dhall.hs
+++ b/dhall/src/Dhall.hs
@@ -127,7 +127,7 @@ import Control.Exception (Exception)
 import Control.Monad.Trans.State.Strict
 import Control.Monad (guard)
 import Data.Coerce (coerce)
-import Data.Either.Validation (Validation(..), ealt, eitherToValidation, validationToEither)
+import Data.Either.Validation (Validation(..), eitherToValidation, validationToEither)
 import Data.Fix (Fix(..))
 import Data.Functor.Contravariant (Contravariant(..), (>$<), Op(..))
 import Data.Functor.Contravariant.Divisible (Divisible(..), divided)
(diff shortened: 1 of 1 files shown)


Gabriel439:
@sjakobi: The decoding logic for the union constructor should essentially never fail (if it does: that's a bug).  Internally there might be a failure when trying to decode some subset of the union, but some other subset always should succeed.

Gabriel439:
I'll close this in favor of https://github.com/dhall-lang/dhall-haskell/pull/1725

PR #1725, opened 2020-03-29
GHC Generics instance for `:+:` now uses `union`

akrmn:
This solves the problem at the root of #1711. Instead of branching into both `L1` and `R1`, a `UnionDecoder` is constructed for the datatype, and is then consumed with `union`, guaranteeing that the only `Decoder` used is the one that matches the constructor.


Gabriel439:
@akrmn: I think your fix uncovered a latent bug that was there previously, which is that this line:

https://github.com/dhall-lang/dhall-haskell/blob/2bf814fb1b445bdca40d90467670ad7d755abecb/dhall/src/Dhall.hs#L1218

... should actually be:

```haskell
    expected_ = Var (V "_" 1)
```

akrmn:
@Gabriel439, I'm not sure if that's the problem, since the errors seem to occur after the actual types have matched the expected types. I'll dig a bit more later today

akrmn:
@Gabriel439, I think I found the problem. It does seem related to the line you mentioned above, though changing it to `expected_ = Var (Dhall.Core.V "_" 1)` gives unbound variable errors.

I think the problem is in the interaction of `union` with the `FromDhall` instances for `Result` and/or `Fix`. The `union` combinator has this assertion:

https://github.com/dhall-lang/dhall-haskell/blob/bf6af02f167df4a32af2466fedb092241e89552a/dhall/src/Dhall.hs#L2465-L2466

This is a situation where it arises. Note that `Generics` doesn't appear to be involved.

```haskell
{-# LANGUAGE DeriveGeneric #-}
{-# LANGUAGE DeriveFunctor #-}
{-# LANGUAGE DeriveAnyClass #-}
{-# LANGUAGE QuasiQuotes #-}

module Main where

import Dhall
import Text.Heredoc
import Numeric.Natural
import Data.Fix (Fix(..))

data ExprF expr
   = LitF Natural
   | ParF expr
   deriving (Eq, Functor, Show)

instance FromDhall expr => FromDhall (ExprF expr) where
  autoWith _ = exprF

exprF :: FromDhall expr => Decoder (ExprF expr)
exprF = union
  (  (LitF <$> constructor "LitF" natural)
  <> (ParF <$> constructor "ParF" auto)
  )

main :: IO ()
main = do
  result <-
    Dhall.input
      Dhall.auto
      [here|
λ(Expr : Type)
→ let ExprF =
        < LitF : Natural
        | ParF : Expr
        >
in   λ(Fix : ExprF → Expr)
     → let Lit = λ(x : Natural) → Fix (ExprF.LitF x)
       in Lit 3
|] ::
      IO (Fix ExprF)
  print result
```

But the `Union`s are not `judgmentallyEqual`, since 

`rest` = `fromList [("ParF",Just (Var (V "_" 1)))]`

`Dhall.Map.delete fld expect` = `fromList [("ParF",Just (Var (V "result" 0)))]`

so we get the error.

akrmn:
Getting back to this, I'm not really sure why the assertion is needed. Shouldn't the types have matched by the time we're extracting the values?

Gabriel439:
@akrmn: If we throw that error it's because somewhere in the `Dhall` module we're incapable of decoding the expression because we found a completely unexpected constructor in the implementation of the `extract` function.

Or, to put it another way, it's not like the extraction code is needlessly applying unnecessary validations that we could just choose to omit; if it throws that error it's because the code could not proceed further because the expression to extract is somehow wrong.

Gabriel439:
@akrmn: I figured out the problem.  See the commit message for https://github.com/dhall-lang/dhall-haskell/pull/1725/commits/a9dd77f3decf0777b74da701878c4f4541b5b5fa

akrmn:
@Gabriel439: I see, thanks for the explanation and for the fix!

Gabriel439:
@akrmn: You're welcome! 🙂

PR #1746, opened 2020-04-09
Don't allow mixed-style records for using expressions

TimWSpence:
Previously it was valid to write

import XXX using [ { header = "foo", mapValue = "bar" } ]
or
import XXX using [ { mapKey = "foo", value = "bar" } ]

whereas only

import XXX using [ { header = "foo", value = "bar" } ]
or
import XXX using [ { mapKey = "foo", mapValue = "bar" } ]

should be allowed

### DIFF
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -138,7 +138,7 @@ module Dhall.Import (
     , HashMismatch(..)
     ) where
 
-import Control.Applicative (Alternative(..))
+import Control.Applicative (Alternative(..), liftA2)
 import Control.Exception (Exception, SomeException, IOException, toException)
 import Control.Monad (when)
 import Control.Monad.Catch (throwM, MonadCatch(catch), handle)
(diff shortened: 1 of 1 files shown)


TimWSpence:
@Gabriel439 as we discussed in https://discourse.dhall-lang.org/t/valid-expressions-for-using-headers/205/3

However, I checked and the mixed style is actually rejected by the type checker. I think this code more reflects the intention but it's probably not strictly necessary due to the type checker doing the work so not offended if you don't want to bother merging!

TimWSpence:
Also, I wasn't sure if you wanted me to add any failure tests to dhall-lang for it?

TimWSpence:
Thanks @Gabriel439 !!! It's a bank holiday here so I'll probably be occupied entertaining the family but will look into the tests and merge early next week

Gabriel439:
@TimWSpence: Actually, it looks like the mixed-style was actually correctly forbidden before, albeit at a downstream step here:

https://github.com/dhall-lang/dhall-haskell/blob/b54787ecbd4a4e2ef3516344da9655aace327566/dhall/src/Dhall/Import.hs#L950-L961

I ran into this when trying to write a regression test and found that the Haskell implementation was actually rejecting it properly:

```
$ dhall <<< 'https://example.com using [ { mapKey = "X-Foo", value = "Foo" } ] as Text'

…
{ - mapValue : …
, + value : …
, …
}

1│                           [ { mapKey = "X-Foo", value = "Foo" } ] : List { mapKey : Text, mapValue : Text }

(stdin):1:27

1│ https://example.com using [ { mapKey = "X-Foo", value = "Foo" } ] as Text

(stdin):1:1
```

However, I still think this is a good change regardless, so I'll go ahead and merge this in without the regression test.

TimWSpence:
👍 thanks @Gabriel439 !

Gabriel439:
@TimWSpence: You're welcome! 🙂

PR #1749, opened 2020-04-11
Use `input` instead of `stdin` for error messages

Gabriel439:
`stdin` is an abbreviation that might not be obvious to less experienced
programmers, so I thought `input` might be more clear

### DIFF
--- a/dhall-bash/exec/Main.hs
+++ b/dhall-bash/exec/Main.hs
@@ -50,7 +50,7 @@ main = do
     (if unHelpful explain then Dhall.detailed else id) (handle (do
         inText <- Data.Text.IO.getContents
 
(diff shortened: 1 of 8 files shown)


Gabriel439:
@sjakobi: Yeah, I thought about that, but I went with "input" since I figured that it wouldn't be clear to the same less technical users why the input was "standard", especially given that we don't have other inputs that aren't already named (like files)

Profpatsch:
Bikeshed, but `stdin` is a well-known tokens that immediately gives any half-experienced programmer a mental image about the thing in question. `input` can be anything.

Gabriel439:
@Profpatsch: The few places where we used `stdin` (now `input`) were places where I believe the meaning was unambiguous

PR #1759, opened 2020-04-22
Change formatting for functions and their types

Gabriel439:
This is based on a suggestion from the functional programming
Slack

The basic difference is that now instead of this:

```dhall
  λ(x : A)
→ λ(y : B)
→ c
```

... we now do this:

```dhall
λ(x : A) →
λ(y : B) →
  c
```

### DIFF
--- a/dhall/src/Dhall/Pretty/Internal.hs
+++ b/dhall/src/Dhall/Pretty/Internal.hs
@@ -307,20 +307,33 @@ hangingBraces n docs =
 
     combineLong x y = x <> y <> Pretty.hardline
 
+unsnoc :: [a] -> Maybe ([a], a)
+unsnoc       []   = Nothing
+unsnoc (x0 : xs0) = Just (go id x0 xs0)
+  where
+    go diffXs x      []  = (diffXs [], x)
+    go diffXs x (y : ys) = go (diffXs . (x:)) y ys
+
 -- | Pretty-print anonymous functions and function types
-arrows :: CharacterSet -> [(Doc Ann, Doc Ann)] -> Doc Ann
-arrows ASCII =
-    enclose'
-        ""
-        "    "
-        (" " <> rarrow ASCII <> " ")
-        (rarrow ASCII <> "  ")
-arrows Unicode =
-    enclose'
-        ""
-        "  "
-        (" " <> rarrow Unicode <> " ")
-        (rarrow Unicode <> " ")
+arrows :: CharacterSet -> [ Doc Ann ] -> Doc Ann
+arrows characterSet docs = Pretty.group (Pretty.flatAlt long short)
+  where
+    long = Pretty.align (mconcat (Data.List.intersperse Pretty.hardline docs'))
+      where
+        docs' = case unsnoc docs of
+            Nothing -> docs
+
+            Just (init_, last_) -> init' ++ [ last' ]
+              where
+                 appendArrow doc = doc <> space <> rarrow characterSet
+
+                 init' = map appendArrow init_
+
+                 last' = space <> space <> last_
+
+    short = mconcat (Data.List.intersperse separator docs)
+      where
+        separator = space <> rarrow characterSet <> space
 
 combine :: CharacterSet -> Text
 combine ASCII   = "/\\"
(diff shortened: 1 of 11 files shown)


ari-becker:
I'll remind you of my earlier comment: https://github.com/dhall-lang/dhall-haskell/pull/1729#issuecomment-606128463

It seems like the auto-formatter's behavior is changing nearly every time that we upgrade and it results in enormous changesets across our codebase, which cause enormous headaches. Either we only update the whitespace when we need to update the underlying code, resulting every so often in huge PRs that we need to carefully review, or we take a break to go repository by repository and open PRs solely for whitespace changes, which halts more productive work (I could script it to speed it up but I feel like I shouldn't *need* to script it).

I don't have a vote, but I would like the maintainers to understand the effects that changes like this have on their userbase. The value of an auto-formatter lies in trusting the auto-formatter to output stable, predictable formatting. In my opinion, PRs like this one undermine that trust.

f-f:
I'm also fairly concerned about the formatter moving very quickly. The context in my case is that [purescript package-sets](https://github.com/purescript/package-sets) are written entirely in Dhall, and we ask people to run `dhall format` before they commit changes.
I'm fine with porting the codebase to new versions of the formatter, but the problem in our case is that if users have a Dhall version that is not the absolute latest (which we use in CI), then the diff in their PR is going to be several thousand lines, and contributors get confused, and we have to ask them to run `nix-shell` with the right versions, etc (we can't recommend `nix-shell` as the only way to contribute because we have Windows users as well).

So my point in all of this is that formatter changes are great (e.g. I very much agree with this change), but I was wondering if we could "bundle" them to reduce impact on reformatting whitespace downstream?
Since lots of our users apparently use `stack` or `nixpkg`, then maybe a good criteria could be "hold changes to formatting until a new version can land on Stackage/nixpkgs", or something like that?

Gabriel439:
@f-f @ari-becker: I'll try to first address the specific concern (about this pull request) followed by the more general concern (about the approach to formatting changes)

For this specific pull request the formatting change is pretty narrow since it only affects function arguments.  I implemented this to ensure that the collateral "damage" was minimal (which you can see for yourself by studying the diff to the formatting golden tests)

Also, there is one thing you can do to mitigate the impact of formatting on things like `git blame`.  For example, you can use `git`'s `blame.ignoreRevsFile` to list revisions which are just reformatting changes and then `git` blame will work correctly and "see through" reformatting changes (assuming you have reformatting changes isolated to their own revisions).  I have not yet checked if GitHub respects this option when rendering blames, but I did verify this using the `git` command line and it does indeed work as advertised.

Now, to address the more general question of: "How does Gabriel decide whether to put up a pull request for a formatting change?".  The usual criterion I use is the same as my criterion for other potentially disruptive changes, which is:

* Do I perceive that this breakage is inevitable? (meaning that I can't stall indefinitely until people give up)
* Do I perceive that existing users agree (albeit reluctantly) that the change is necessary?

If the answer to both of those questions is "yes", then I tend to swift about making the change (e.g. "rip off the bandaid").  The rationale is that if the change is truly inevitable then prolonging the agony doesn't make things any easier.  Quite the opposite: prolonging the agony makes things worse because the longer things drag on then the greater the disruption will be when the change is finally be made (since more Dhall code has been written in the interim which will be disrupted).

timbertson:
I've had similar thoughts on formatting, as I (using nix) am often a bit behind colleagues using the latest binaries downloaded from github. I was wondering if it'd be possible to set a language version for `dhall format` - i.e. `dhall format --version=<version>`, so that people can use new features without breaking evaluation on older versions. The same option could decouple "update dhall" from "reformat all my stuff" (I appreciate the improvements, I just want more control over it).

It wouldn't solve the noise, but at least people could collaborate using the lowest version between them instead of forcing big bang updates.

I'm assuming this would be a right pain to implement though, so consider it an idea more than a request ;)

Gabriel439:
@timbertson: Yeah, it would be very difficult to implement and maintain, not only because it would require specifying multiple parallel formatting algorithms but also because we'd have to support multiple older ASTs, too.  The closest you can get is to use an older version of the `dhall` executable when formatting your code.

PR #1767, opened 2020-04-29
Change formatting of `if` expressions

Gabriel439:
* Remove blank lines inside the expression
* Don't indent the outermost `if`


ari-becker:
If you'll permit me a little bikeshedding, what about:
```dhall
if   Foo
then Bar
else Baz
```
i.e. adding whitespace after the `if` so that the expressions align?

PR #1778, opened 2020-05-08
Fix version numbers and update `cabal` version to 1.10

Gabriel439:
I realized that I forgot to update the version numbers for `dhall-{json,yaml}`
as part of #1777 

Also, Hackage is starting to reject Haskell packages that have a
`Cabal-Version` with a minimum bound less than 1.10, which this
change fixes.


Gabriel439:
@sjakobi: `dhall-yaml` technically had a breaking change because the `Options` type was publicly exposed and changed in the course of the schema inference pull request

PR #1797, opened 2020-05-16
(small) nix/shared.nix: pull cabal2nix from non-static pkgs

Profpatsch:
See commit messages, this strips quite some build time from the static binaries.

Actually, what I forgot to mention in the commit message, building cabal2nix statically without using the custom binary cache is not possible because the tests of pyopenssl in that version are flaky in that nixpkgs version, so this came out of looking into that.


Profpatsch:
Eh, the hydra build was killed.

PR #1801, opened 2020-05-19
fix(parser): record fields should not include language keywords

### LINKED ISSUE #1750: Haskell implementation not rejecting keyword labels
The standard requires record labels to not be keywords:

https://github.com/dhall-lang/dhall-lang/blob/master/standard/dhall.abnf#L226

... but the Haskell implementation does not correctly reject them.  For example, this expression:

```bash
$ dhall <<< '{ if : Text }'
{ if : Text }
```

... should have been rejected.  See also: https://github.com/travisbrown/dhallj/issues/17

german1608:
# Summary

Fixes #1750 

# Solution

Split `reservedKeywords` from `reservedIdentifiers`

Currently there are failing tests because [standard says that `Some` is a reserved keyword](https://github.com/dhall-lang/dhall-lang/issues/1012). Meanwhile, I'll have this PR as a draft

### DIFF
--- a/dhall/tests/format/applicationMultilineA.dhall
+++ b/dhall/tests/format/applicationMultilineA.dhall
@@ -18,7 +18,7 @@
 , someList =
     Some
       [ aaaaaaaaaaaaaaaaaaaaaaaaaa, bbbbbbbbbbbbbbbbbbbb, cccccccccccccccccccccccc, dddddddddddddd ]
-, merge =
+, merge1 =
     merge
       aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
       bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
--- a/dhall/tests/format/applicationMultilineA.dhall
+++ b/dhall/tests/format/applicationMultilineA.dhall
@@ -27,7 +27,7 @@
       a
       b
       ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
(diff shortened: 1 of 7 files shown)


german1608:
I've created [this](https://github.com/dhall-lang/dhall-lang/pull/1013) PR to add tests for this on common repository

german1608:
Actual footage of me with all the issues/prs references I've made last hour:

Gabriel439:
As @sjakobi noted in https://github.com/dhall-lang/dhall-lang/issues/1012#issuecomment-630650616 you will need to add an exception for `Some` for record labels.  Otherwise, this looks great to me!

german1608:
I added a `anyLabelOrSome` to have our parser impl similar as official grammar specs.

There were other places where I had to use `anyLabelOrSome`, like union types and field projections.

I've had to ammend some test files because it was using invalid record field labels.

I've added tests for this in dhall-lang/dhall-lang#1013

dhall/tests/format/applicationMultilineA.dhall
```diff
@@ -18,7 +18,7 @@
 , someList =
     Some
       [ aaaaaaaaaaaaaaaaaaaaaaaaaa, bbbbbbbbbbbbbbbbbbbb, cccccccccccccccccccccccc, dddddddddddddd ]
-, merge =
+, merge1 =
```
Gabriel439, line 21: For future reference, another thing you can do to keep the original name is to quote the label in backticks:

```dhall
, `merge` =
```

... but this is fine, too

PR #1821, opened 2020-05-29
Allow http-client 0.7

felixonmars:
Builds fine and all tests pass here.


Gabriel439:
Thank you!  I also added a matching revision to Hackage

Issue #1838, opened 2020-06-04
dhall-nix: Please, bump dependency HNix to 0.8.0 or stabilize in Nixpkgs


Anton-Latukha:
<details><summary>Current Nixpkgs DHall-Nix Nix build tail:</summary>
<p>

```
Configuring dhall-nix-1.1.14...
CallStack (from HasCallStack):
  die', called at libraries/Cabal/Cabal/Distribution/Simple/Configure.hs:1022:20 in Cabal-3.0.1.0:Distribution.Simple.Configure
  configureFinalizedPackage, called at libraries/Cabal/Cabal/Distribution/Simple/Configure.hs:475:12 in Cabal-3.0.1.0:Distribution.Simple.Configure
  configure, called at libraries/Cabal/Cabal/Distribution/Simple.hs:625:20 in Cabal-3.0.1.0:Distribution.Simple
  confHook, called at libraries/Cabal/Cabal/Distribution/Simple/UserHooks.hs:65:5 in Cabal-3.0.1.0:Distribution.Simple.UserHooks
  configureAction, called at libraries/Cabal/Cabal/Distribution/Simple.hs:180:19 in Cabal-3.0.1.0:Distribution.Simple
  defaultMainHelper, called at libraries/Cabal/Cabal/Distribution/Simple.hs:116:27 in Cabal-3.0.1.0:Distribution.Simple
  defaultMain, called at Setup.hs:2:8 in main:Main
Setup: Encountered missing or private dependencies:
dhall >=1.31 && <1.33, hnix ==0.7.*

builder for '/nix/store/qsikazg9k94wxif0y2j4a6ck813d4jaf-dhall-nix-1.1.14.drv' failed with exit code 1
```
</details>

`HNix 0.8.0`: Breaking change, `NAtom` type (& module) - received an `NURI Text` data constructor.

---

Soon we would probably release minor fix, fixing Hackage documentation.

---

Sorry in advance. There seen, and would be many breakage changes and so - major versions. Probably a lot.
But since the project is far from working state - I think those breakages would be/is no biggy. And since the project does not work - now is the best time to do needed breakages as soon as they can be done, and the only time we can and should do them, since the project does not work really and even still somewhat far from being cleaned-up into shape.

---

If something - we can stabilize the DHall-Nix on some HNix version by creating a separate HNix version in Nixpkgs.

Gabriel439:
[...]

PR #1841, opened 2020-06-05
Build against `hnix-0.8`

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/1838

### DIFF
--- a/dhall-nix/dhall-nix.cabal
+++ b/dhall-nix/dhall-nix.cabal
@@ -31,7 +31,7 @@ Library
         containers                              < 0.7 ,
         data-fix                                < 0.3 ,
         dhall                     >= 1.31    && < 1.33,
-        hnix                      >= 0.7     && < 0.8 ,
+        hnix                      >= 0.7     && < 0.9 ,
         lens-family-core          >= 1.0.0   && < 2.2 ,
         neat-interpolation                      < 0.6 ,
         text                      >= 0.8.0.0 && < 1.3


Gabriel439:
I verified that `dhall-nix` still builds against the newer version of `hnix` without changes

Anton-Latukha:
@sjakobi

Woops: I've looked at the project description, searched Reports requesting it, and opened Report that askes to add `0.8`.

Also, I think this can not look better than #1822, since it is seen you did work in there.

---

I'm also getting lost in Nix files sometimes. Moreover for that, some of Nix files in projects are manually crafted, and some are automatically generated - and often it is too obscure which are which, especially if people write Nix files without comments.

---

Sidenote: I do not know about `libsodium` situation. If `hnix-store-core-0.2` depends on unavailable in Nixpkgs `libsodium` version, that is often the case the package for separate version is explicitly created in Nixpkgs, but of course properly for `hnix-store-core-0.2` to follow upstream package versions.

Also another sidenote about `nhix-store`s: `hnix-store-remote` is `0.1.0` in Nixpkgs, and it is indeed a broken package there.

---

Gabriel439:
@Anton-Latukha: Generally, I try to avoid deviating from the default package set as much as possible to avoid maintaining a large number of pinned dependencies, because then upgrading to newer versions of Nixpkgs becomes more difficult.  I simply locally verify that the project builds against the newer version of the dependency, which it does in this case.

The only thing I had to do in order to build locally against `hnix-0.8` (using Nix) was to add the `some` package to the package set and increase the upper bound and that was it.

Usually the only exception I make to this rule is if there is some performance improvement to using a newer version of a dependency

Anton-Latukha:
Yes, Nixpkgs and Nix tooling could still receive some (a lot of) work, simplifications, and automation.

You know that Nixpkgs is currently objectively the best software store, technologically, as also by size, and effort put in as also the consolidation of that effort.

What I would say - we can learn from Nixpkgs mistakes.

For example, yesterday I thought: In Nixpkgs the "default" `pandoc` is old versioin `2.7.4` there, and there is a separate package `pandoc_2_9_2`. Developers stopped supporting old`2.7.4`. But people that use `pandoc` in Nixpkgs - stuck on old version, since they somehow should know there exists  `pandoc_2_9_2`, so they should manually switch to it, and then when Nixpkgs migration happens - switch back.

It should be the other way around.

Newer versions of package should be always a default tip. There no stopping the constant progress, so tip moves forward constantly.
The default tip - should always be current, if not even `master` packages. And old/stable versions should be branched-out into a separate tails, or use old package description revision, and the package manager, if the code says to it (please, `pandoc < 2.9.0`) - should prioritize finding and picking some newest `pandoc < 2.9.0` version there is in the repository automatically.

Those processes are and should be internal to the package management, and not given to manual external human work.

Gabriel439:
@Anton-Latukha: I don't think this repository is a good forum for discussing how Nixpkgs should operate

Anton-Latukha:
True. I would pardon the explanation, it is too late, not possible to change the core Nixpkgs processes.

PR #1842, opened 2020-06-05
Build against `doctest-0.17`

Gabriel439:
Related to https://github.com/commercialhaskell/stackage/issues/5406


Gabriel439:
I verified that the respective repositories build and pass tests against `doctest-0.17`

Issue #1784, opened 2020-05-10
LSP Freeze Import broken with imports with fallback


SiriusStarr:
Right-clicking on an import (in VSCode) and selecting "Freeze (refreeze) import" no longer works with the new LSP version (1.0.7) with imports that have fallbacks, instead failing with the error "You weren't pointing at an import!"  Whether or not a (valid or invalid) hash already exists does not affect this behavior.

For instance, given a file `A.dhall` containing simply `0`, the following cannot be frozen through the LSP:
```
./A.dhall ? ./A.dhall
```

```
  ./A.dhall sha256:d0a48d41dcc572839faa9499b6c316b8d0d7115efac2ac770f062cd5fd6d0181
? ./A.dhall
```

```
  ./A.dhall sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
? ./A.dhall
```

and the following freeze normally:
```
./A.dhall
```

```
./A.dhall sha256:d0a48d41dcc572839faa9499b6c316b8d0d7115efac2ac770f062cd5fd6d0181
```

```
./A.dhall sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
```

Tracing messages to the LSP doesn't seem to reveal anything different between the broken and working cases:
```
[Trace - 2:09:18 PM] Sending request 'workspace/executeCommand - (249)'.
Params: {
    "command": "dhall.server.freezeImport",
    "arguments": [
        {
            "position": {
                "line": 0,
                "character": 5
            },
            "textDocument": {
                "uri": "file:///home/user/broken1.dhall"
            }
        }
    ]
}


[Trace - 2:09:18 PM] Received notification 'window/showMessage'.
Params: {
    "type": 2,
    "message": "You weren't pointing at an import!"
}
```
versus
```
[Trace - 2:10:50 PM] Sending request 'workspace/executeCommand - (256)'.
Params: {
    "command": "dhall.server.freezeImport",
    "arguments": [
        {
            "position": {
                "line": 0,
                "character": 6
            },
            "textDocument": {
                "uri": "file:///home/user/working1.dhall"
            }
        }
    ]
}


[Trace - 2:10:50 PM] Received request 'workspace/applyEdit - (7)'.
Params: {
    "edit": {
        "changes": {
            "file:///home/userworking1.dhall": [
                {
                    "range": {
                        "start": {
                            "line": 0,
                            "character": 9
                        },
                        "end": {
[...]

PR #1849, opened 2020-06-13
Fix double-nesting of `Note` constructors

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/1784

In #1824 I introduced a bug in computing source spans for `Note` constructors.
The (very indirect) consequence of this bug was that doubly-nested `Note`
constructors would show up in the syntax tree.

For example, the following command would generate *far* more `Note`
constructors than you would expect before this fix:

```bash
$ dhall haskell-syntax-tree --noted <<< 'A B'
```

Among other things, that broke the LSP server, which assumes that
`Note` constructors are only nested once.

I added a regression test to prevent this from recurring in the
future.

### DIFF
--- a/dhall/src/Dhall/Parser/Expression.hs
+++ b/dhall/src/Dhall/Parser/Expression.hs
@@ -64,6 +64,14 @@ src parser = do
     after       <- getSourcePos
     return (Src before after tokens)
 
+-- | Same as `src`, except also return the parsed value
+srcAnd :: Parser a -> Parser (Src, a)
+srcAnd parser = do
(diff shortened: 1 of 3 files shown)


TristanCacqueray:
The test failure was caused by `Error: Remote host not found URL: https://test.dhall-lang.org/random-string`. Otherwise the change works as described, `A B` now results in 4 notes (from 16)

Gabriel439:
I'm going to disable that test for now, since it keeps causing us issues

Issue #1796, opened 2020-05-15
*-to-dhall: generate auto-completed record using a --schemas file


sbrunk:
I'm trying to convert a bunch of kubernetes resources (currently generated via helm) to dhall/dhall-kubernetes.

Given the existing resources, using `yaml-to-dhall` as described  by @Gabriel439 in this [SO answer](https://stackoverflow.com/questions/61010406/is-it-possible-to-get-a-kubernetes-object-in-json-or-yaml-with-all-optional-fiel) seems like a promising approach to bootstrap the conversion.

But, as also explained in that answer, *`yaml-to-dhall` doesn't yet take advantage of support for default values*. This means that the generated dhall file contains `None` for all optional values not set in the original yaml. It makes the imported definitions quite large and somewhat convoluted.

So I wonder if it would be feasible to add an option to `yaml-to-dhall` to omit unset fields in the generated dhall definition?

german1608:
If I understand your suggestion, you're saying that `yaml-to-dhall` could have a flag like `--omit-null` for example, to omit those files.

The only thing I see is that the generated dhall definition doesn't have the type provided by, in that SO Answer, Deployment.dhall and won't be compatible with that since the absence of a field in a dhall record isn't the same as if that field is `None` there.

```
Welcome to the Dhall v1.32.0 REPL! Type :help for more information.
⊢ ./schema.dhall 

{ o : Optional Natural }

⊢ {} : ./schema.dhall 

Error: Expression doesn't match annotation

- { … : … }
+ Type

1│ {} : ./schema.dhall

(input):1:1
⊢ { o = None Natural} : ./schema.dhall 

{ o = None Natural }

⊢ { o = Some 1} : ./schema.dhall 

{ o = Some 1 }

⊢ { o = Some 1 } : ./schema.dhall 

{ o = Some 1 }```

So yes I think it will be possible but you'll lose backwards compatibility between generated dhall and original type

sbrunk:
Oh I see. I've just started with dhall and I'm still in the process of figuring it out.

So I guess what I'm **actually** looking for is for the generated dhall records to be typed w.r.t to `kubernetes.Deployment.Type` from Deployment.dhall in that case.

I guess that's not possible (yet)?

In that case something like `--omit-null` would still be helpful in a partially manual conversion process as it would remove a lot of noise.

Gabriel439:
[...]

PR #1853, opened 2020-06-13
*-to-dhall: generate auto-completed record using a --schemas file

TristanCacqueray:
Fixes: #1796

### DIFF
--- a/dhall-json/json-to-dhall/Main.hs
+++ b/dhall-json/json-to-dhall/Main.hs
@@ -216,7 +216,7 @@ main = do
                       Just schemasText -> do
                           parsedSchemas <- Dhall.Core.throws (Dhall.Parser.exprFromText "(input)" schemasText)
 
-                          dhallFromJSONSchemas parsedSchemas expression
+                          useSchemas parsedSchemas expression
 
                       Nothing ->
                           return (fmap absurd expression)
--- a/dhall-json/json-to-dhall/Main.hs
+++ b/dhall-json/json-to-dhall/Main.hs
@@ -221,13 +221,6 @@ main = do
                       Just filePath -> do
                           inputSchemas <- loadSchemas filePath
 
-{- TODO: remove debug...
-                          putStrLn $ "value        => " <> show value
-                          putStrLn $ "finalSchema  => " <> show finalSchema
-                          putStrLn $ "inputSchemas => " <> show inputSchemas
-                          putStrLn $ "expr         => " <> show expression
--}
-
                           Dhall.Core.throws (dhallFromJSONSchemas filePath inputSchemas expression)
 
                       Nothing -> return $ fmap absurd $ expression
--- a/dhall-json/json-to-dhall/Main.hs
+++ b/dhall-json/json-to-dhall/Main.hs
@@ -30,8 +30,8 @@ import qualified Options.Applicative                       as Options
 import qualified System.Console.ANSI                       as ANSI
 import qualified System.Exit
 import qualified System.IO                                 as IO
-import qualified Dhall
 import qualified Dhall.Core
+import qualified Dhall.Parser
 import qualified Dhall.Pretty
 import qualified Paths_dhall_json                          as Meta
 
@@ -51,7 +51,7 @@ parserInfo = Options.info
 data Options
     = Default
         { schema     :: Maybe Text
-        , schemas    :: Maybe FilePath
+        , schemas    :: Maybe Text
         , conversion :: Conversion
         , file       :: Maybe FilePath
         , output     :: Maybe FilePath
@@ -105,8 +105,8 @@ parseOptions =
     parseSchemas =
         Options.strOption
             (   Options.long "schemas"
-            <>  Options.help "List of schemas to look for record completion output"
-            <>  Options.metavar "FILE"
+            <>  Options.help "Record of schemas to use in generated output"
+            <>  Options.metavar "RECORD"
             )
 
     parseVersion =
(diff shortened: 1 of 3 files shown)


TristanCacqueray:
Given a `schemas.dhall` file with:
```dhall
{ Person = { Type = { name : Text, alive : Bool }, default = { alive = True } }
}
```

Here are some examples:
```console
$ json-to-dhall --schemas ./schemas.dhall <<< '{"name": "Toto"}'
let schemas = ./schemas.dhall in schemas.Person::{ name = "Toto" }
$ json-to-dhall --schemas ./schemas.dhall <<< '{"name": "Toto", "alive": true}'
let schemas = ./schemas.dhall in schemas.Person::{ alive = True, name = "Toto" }
$ json-to-dhall --schemas ./schemas.dhall <<< '{"name": "Toto", "alive": 42}'
let schemas = ./schemas.dhall in schemas.Person::{ alive = 42, name = "Toto" }
$ json-to-dhall --schemas ./schemas.dhall <<< '{"nop": "Toto"}'

Error: Input does not match any schema
```

This doesn't check record completion type, nor it removes un-necessary defaults. And most importantly, this doesn't check for sub-expression record completion. But that's a start :-)

dhall-json/src/Dhall/JSONToDhall.hs
```diff
@@ -1103,6 +1109,93 @@ dhallFromJSON (Conversion {..}) expressionType =
...
+          })
+      (mkRecordCompletion name attrs)
+    mkRecordCompletion :: Text -> Map.Map Text ExprS -> ExprS
+    mkRecordCompletion name attrs = D.RecordCompletion (D.Field (D.Var (D.V "schemas" 0)) name) (D.RecordLit attrs)
+    mkImport = D.Import {
+      importMode = D.Code,
+      importHashed = D.ImportHashed {
+          hash = Nothing,
+          importType = D.Local D.Here (D.File {directory = D.Directory {components = []},
+                                               file = Text.pack (drop 2 filePath) })
```
TristanCacqueray, line 1195: TODO: make this work for path that are not local relative (e.g. not `./file.dhall`)

dhall-json/json-to-dhall/Main.hs
```diff
@@ -46,9 +49,14 @@ parserInfo = Options.info
...
 -- | All the command arguments and options
+data Schemas
+    =  SchemaName (Maybe Text)
+    |  SchemasFile FilePath
+    deriving Show
+
 data Options
     = Default
-        { schema     :: Maybe Text
+        { schemas    :: Schemas
```
Gabriel439, line 59: I think we should use a separate command-line option for specifying the schemas used to simplify the file

The reason why is that the existing singular schema argument is used to specify the expected Dhall type of the complete file; if the file does not match the declared schema then the conversion rejects the file.

These other new schemas are used for a different purpose: simplifying the generated Dhall code.  If none of them match then it's not an error.

The other reason we want to separate the two concepts is because the user is allowed to omit the old singular schema and let the compiler infer the correct Dhall type.  However, we'd still like to be able to simplify the inferred type with these latter new schemas if they happen to match against the type.

TristanCacqueray: I assumed those would be conflicting options, either the conversion match a single type, or it find the type from a list. I now see how both concepts can be used together, it would actually make the Main.hs modification simpler.

TristanCacqueray:
Please find an update to the implementation, here is what I managed to support.

Given a `schema.dhall` file containing:

```dhall
let Address =
      { Type = { country : Text, city : Optional Text }
      , default.city = None Text
      }

in  { Person =
        { Type = { name : Text, addr : Address.Type, alive : Bool }
        , default.alive = True
        }
    , Address
    }
```

We can now:

```ShellSession
$ json-to-dhall -- --schemas ./schemas.dhall <<< '{"name": "Alice", "addr": {"country": "Switzerland"} }'
```
```dhall
let schemas = ./schemas.dhall

in  schemas.Person::{
    , name = "Alice"
    , addr = schemas.Address::{ country = "Switzerland" }
    }
```

However there is an issue when Optional values are provided, e.g.: `{"name": "Bob", "addr": {"country": "Switzerland", "city":  "Zurich"} }` can't be found in the schemas because city is first decoded as a Text, not an Optional Text.

Using `json-to-dhall '(./schemas.dhall).Person.Type' --schemas ./schemas.dhall` command line SCHEMA argument doesn't help either because the Person type is missing the `alive` attribute.

This works though:

```ShellSession
$ json-to-dhall -- '(./schemas.dhall).Person.Type' --schemas ./schemas.dhall <<< '{"name": "Eve", "alive": true, "addr": {"country": "Switzerland", "city": "Zurich"} }'
```
```dhall
let schemas = ./schemas.dhall

in  schemas.Person::{
    , name = "Eve"
    , addr = { city = Some "Zurich", country = "Switzerland" }
    }
```

If I remember correctly, I need to use the `rewriteOf` function, but I don't think I understand how... If someone wants to take over, please feel free to do so, otherwise I now need help to figure out the next step for this feature.

Gabriel439:
@TristanCacqueray: Since I cannot push to your repository I created a pull request with my suggested changes:

https://github.com/TristanCacqueray/dhall-haskell/pull/1

If you can add me to your repository I can then iterate on your branch directly

TristanCacqueray:
That's amazing, thank you @Gabriel439 . You should now be able to push directly.

Gabriel439:
In the course of implementing this I realized that there is nothing YAML-specific about this feature, so my plan is to upstream this functionality into the `dhall` package as a `dhall` subcommand (e.g. `dhall using-schemas` or something similar)

Gabriel439:
I created a new pull request upstreaming this functionality into the `dhall` package so I'll close this pull request in favor of that one: https://github.com/dhall-lang/dhall-haskell/pull/1902

Issue #1847, opened 2020-06-11
LSP - Unused binding should have targeted highlighting


drewolson:
I've been really enjoying using my editor's integration with the dhall LSP (vim + coc). However, the linter seems to be overzealous in highlighting my code when there is an unused binding. All of the code in the file is highlighted rather than the specific unused binding. I've attached a screenshot.

![dhall-lint](https://user-images.githubusercontent.com/3536/84449903-54c8f180-ac14-11ea-9961-67b25e28c9d1.png)

Thanks!

Gabriel439:
I have a fix up here: https://github.com/dhall-lang/dhall-haskell/pull/1854

PR #1854, opened 2020-06-13
Narrow source span for unused let warnings

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/1847

### DIFF
--- a/dhall-lsp-server/tests/Main.hs
+++ b/dhall-lsp-server/tests/Main.hs
@@ -1,4 +1,5 @@
-{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE OverloadedStrings     #-}
 
 import Control.Monad.IO.Class (liftIO)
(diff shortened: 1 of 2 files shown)


Gabriel439:
Here is a screenshot showing the new narrowed-down hints:

Screen Shot 2020-06-13 at 3 52 47 PM

dhall-lsp-server/tests/Main.hs
```diff
@@ -48,15 +50,36 @@ lintingSpec fixtureDir =
...
+                    , _end = Position { _line = 4, _character = 38 }
+                    }
+                , _severity = Just DsHint
+                , _code = Nothing
+                , _source = Just "Dhall.Lint"
+                , _message = "Unused let binding 'carl'"
+                , _tags = Nothing
+                , _relatedInformation = Nothing
+                }
+            ]
```
sjakobi, line 81: I'm slightly concerned that these `Diagnostic` values might need frequent updating if the type changes shape. 

But we can still pick out the important fields once that happens!

PR #1863, opened 2020-06-16
feat(dhall-docs): parse markdown from header

german1608:
# Summary

This PR includes:

* `getAllDhallFiles` now only filters by the `.dhall` extension.
* Headers are extracted and parsed using `mmark`. The result is then written directly to each html directory.

Sadly, I can't provide examples at this moment because of connectivity issues, but if you'd like to test locally here is an example that I used that includes frequently used markdown features:

https://gist.github.com/german1608/64e0e912851a24058218ae3cbb136c2c

The first character of each line can't be indented because the tool parses weirdly adding some `<pre>` html elements. A solution for this could be to re-indent before passing the file, but I'd like to keep it that way because is more of a WYSWYG philosophy.


german1608:
Mmmm I don't understand the [last](https://hydra.dhall-lang.org/build/61735/nixlog/1) error log.

it is compiling with ghc 8.6, so it should have `containers-0.6.0.1`.

It fails building `text-metrics-0.3.0` which was required by mmark, with the following boundaries error:

```containers >=0.5.6.2 && <0.6```

_but_ the boundaries on that package on [hackage](http://hackage.haskell.org/package/text-metrics) says:

```containers >=0.5 && <0.7```

why do they differ on the hydra build?

**Comment by sjakobi** [comment]
@german1608 Take a look at the Hackage revisions for `text-metrics-0.3.0`:

http://hackage.haskell.org/package/text-metrics-0.3.0/revisions/

The containers bounds were widened only in the last revision. I assume that  the nixpkgs version we use still includes the old revision.

I think the fix would be to add the derivation for the latest version of `text-metrics` to `nix/packages`.

**Comment by german1608** [comment]
After adding 7 .nix files to `nix/packages`, I think that this is ready for your review :smile:

**Comment by german1608** [comment]
Rendered markdown in dark-mode:
image

Rendered markdown in light-mode:
image

File: dhall-docs/src/Dhall/Docs.hs [hunk 1]
```diff
@@ -94,6 +98,15 @@ parserInfoOptions =
         <>  Options.Applicative.progDesc progDesc
         )
 
+showParseError
```
sjakobi, line 101: I'd maybe call this `showMarkdownParseError` or similar, to prevent confusion with Dhall code parse errors.

I assume that eventually we'll have to add some dedicated error types here, but it's probably too early right now.

dhall-docs/src/Dhall/Docs.hs
```diff
@@ -106,20 +119,32 @@ getAllDhallFilesAndHeaders
     :: Path Abs Dir -- ^ Base directory to do the search
     -> IO [(Path Abs File, Header)]
 getAllDhallFilesAndHeaders baseDir = do
-    files <- snd <$> Path.IO.listDirRecur baseDir
+    files <- filter hasDhallExtension . snd <$> Path.IO.listDirRecur baseDir
     Data.Maybe.catMaybes <$> mapM readDhall files
   where
+    hasDhallExtension :: Path Abs File -> Bool
+    hasDhallExtension absFile = case Path.splitExtension absFile of
```
sjakobi, line 126: I somewhat wonder whether there might be a faster alternative to `splitExtensions` here, but it probably doesn't matter, at least for now.

dhall-docs/src/Dhall/Docs.hs
```diff
@@ -106,20 +119,32 @@ getAllDhallFilesAndHeaders
...
-                    _ -> Nothing
-
+        contents <- Text.IO.readFile filePath
+        case exprAndHeaderFromText filePath contents of
+            Right (header, _) -> return $ Just (absFile, removeComments header)
+            Left ParseError{..} -> do
+                putStrLn $ showParseError unwrap
+                return Nothing
+
+    removeComments :: Header -> Header
```
sjakobi, line 140: Maybe call this `stripComment` or `stripCommentSyntax`?!

sjakobi: The type seems wrong – the output isn't a valid `Header` anymore! Maybe output `Text` or a dedicated newtype?

dhall-docs/src/Dhall/Docs.hs
```diff
@@ -106,20 +119,32 @@ getAllDhallFilesAndHeaders
...
+    removeComments (Header header)
+        | "--" `Data.Text.isPrefixOf` strippedHeader = Header $ Data.Text.drop 2 strippedHeader
```
sjakobi, line 142: Aside: I thought we wanted to distinguish between documentation comments and "internal" comments. Is this still planned?

sjakobi, line 142: `Data.Text.stripPrefix` and `stripSuffix` might be handy in this context.

dhall-docs/src/Dhall/Docs/Html.hs
```diff
@@ -35,24 +33,10 @@ data DocParams = DocParams
...
-    | otherwise = strippedHeader
-
-  where
-    strippedHeader = Data.Text.strip header
-
 -- | Generates an @`Html` ()@ with all the information about a dhall file
 filePathHeaderToHtml
-    :: (Path Abs File, Header) -- ^ (source file name, parsed header)
-    -> DocParams               -- ^ Parameters for the documentation
+    :: (Path Abs File, Html ()) -- ^ (source file name, parsed header)
```
sjakobi, line 38: Update the haddocks too.

dhall-docs/src/Dhall/Docs/Markdown.hs
```diff
@@ -0,0 +1,29 @@
...
+
+type ParseError = ParseErrorBundle Text MMarkErr
+
+{-| Takes a text that could contain markdown and returns the generated HTML.
+    If an error occurs while parsing, it also returns the error information.
+-}
+markdownToHtml
+    :: Path Abs File -- ^ Used by `Mmark.parse` for error messages
+    -> Text          -- ^ Text to parse
+    -> (Maybe ParseError, Html ())
```
sjakobi, line 25: I think it might be better to return `Either ParseError (Html ()))` here.

The current type is a bit confusing IMHO. When we get a parse error, the returned HTML is _not_ generated from the parsed markdown, contrary to what the function name seems to say!

german1608:
> A few comments, some a bit more brusque than intended, I feel. Sorry!
> 
Don't worry! I see your reviews as a source of knowledge.

dhall-docs/src/Dhall/Docs.hs
german1608: I used that function to show both dhall and markdown parse error as warnings (should rename the `Error` suffix to `Warning` I guess.

I think I'll create a dedicated type for markdown error so we don't use the same function for the both kind of parse errors

dhall-docs/src/Dhall/Docs.hs
german1608: I guess that `fileExtension` is shorter, but it calls `splitExtension` under the hood as well:

https://github.com/commercialhaskell/path/blob/a75ddfc86319cd2187eb8965bdb11b52690fa002/src/Path/Include.hs#L459-L460

dhall-docs/src/Dhall/Docs.hs
german1608: I tend to forget (rookie haskell errors, I think) that ADT represents stuff in real world and as well they should be valid. You're right, output is not a `Header` anymore.

I think I'll add a dedicated type for this purpose

dhall-docs/src/Dhall/Docs.hs
german1608: Although on the discourse [thread](https://discourse.dhall-lang.org/t/dhocs-cli-utility/267/24) for `dhall-docs` we didn't reach an agreement about it, I think it will be better to add the distintion.

Currently I don't because of a simple reason: The `prelude` comment headers are "internal" comments if we say so. I'd have to update prelude later. But for testing purposes, I'd prefer to add the distintion for "internal" and documentation comments later.

dhall-docs/src/Dhall/Docs.hs
sjakobi: Alternatively we could possibly use [`isSuffixOf`](http://hackage.haskell.org/package/base-4.14.0.0/docs/Data-List.html#v:isSuffixOf), but, again, I don't know whether that is actually faster, or whether we really _need_ anything faster.

dhall-docs/src/Dhall/Docs.hs
sjakobi: No worries! :)

dhall-docs/src/Dhall/Docs.hs
german1608: Agree. Shall I ask again on the thread? Or add another thread?

sjakobi: I'd be curious what @Gabriel439  thinks about this first. Continuing the discussion in the same thread seems fine to me though.

dhall-docs/src/Dhall/Docs.hs
german1608: Oh, another solution can be to use `turtle` module to filter using `find`, check this example https://hackage.haskell.org/package/turtle-1.5.19/docs/Turtle-Prelude.html

The only downside is that it returns `FilePath`, and I won't like to do a lot of path conversions, that may uglify the code.

Also I don't know if that is actually faster. `splitExtension` implementation looks roughly O(n), and we execute it on each file. `find` executes the specified pattern on each file, but I'm not sure how much is the complexity difference. Maybe @Gabriel439 can illustrate us a little bit?

Another thing is that I tested with the biggest dhall package I found: dhall-kubernetes. And it isn't really that slow, so we can stick with `splitExtensions` with no problem

dhall-docs/src/Dhall/Docs.hs
german1608: I ended up returning just `Text`, but I'm sure that in the future probably I'll have a dedicated data type to represent _extracted documentation_

german1608:
@sjakobi I think I reviewed all your comments, left unresolved those where we mentioned @Gabriel439 to see his thoughts. Let me know if you have any other regards.

dhall-docs/src/Dhall/Docs.hs
Gabriel439: I personally would stick with `splitExtension`.  It's highly unlikely that it would be the bottleneck

dhall-docs/src/Dhall/Docs.hs
Gabriel439: We could do the same thing that Haskell does and use `{-|` or `-- |` to denote comments that should be interpreted by the documentation generator

dhall-docs/src/Dhall/Docs/Html.hs
```diff
@@ -35,24 +33,10 @@ data DocParams = DocParams
...
-    | otherwise = strippedHeader
-
-  where
-    strippedHeader = Data.Text.strip header
-
 -- | Generates an @`Html` ()@ with all the information about a dhall file
 filePathHeaderToHtml
-    :: (Path Abs File, Header) -- ^ (source file name, parsed header)
-    -> DocParams               -- ^ Parameters for the documentation
+    :: (Path Abs File, Html ()) -- ^ (source file name, header document as html)
```
sjakobi, line 38: Nit: Why is this a tuple instead of two separate parameters?

dhall-docs/src/Dhall/Docs.hs
```diff
@@ -162,32 +178,57 @@ saveHtml inputAbsDir outputAbsDir packageName t@(absFile, _) = do
...
+        "The original non-markdown text will be pasted in the documentation"
+
+    stripCommentSyntax :: Header -> Text
+    stripCommentSyntax (Header h)
+        | Just s <- Data.Text.stripPrefix "--" strippedHeader
+            = s
+        | Just commentPrefixStripped <- Data.Text.stripPrefix "{-" strippedHeader
+        , Just commentSuffixStripped <- Data.Text.stripSuffix "-}" commentPrefixStripped
+            = commentSuffixStripped
+        | otherwise = strippedHeader
```
sjakobi, line 209: In which scenario are we hitting this branch?

sjakobi, line 209: Why does this strip whitespace, while the other branches don't?

dhall-docs/src/Dhall/Docs.hs
```diff
@@ -150,7 +166,7 @@ saveHtml
     -> String                   -- ^ Package name
     -> (Path Abs File, Header)  -- ^ (Input file, Parsed header)
```
sjakobi, line 167: (Similar nit regarding tuple vs separate params. Maybe you could instead use `uncurry` on the call site.)

dhall-docs/src/Dhall/Docs.hs
german1608: @Gabriel439 Cool, then I'll mark this discussion as solved

dhall-docs/src/Dhall/Docs.hs
german1608: I'm going to update the prelude to have this change and what we have discussed at #1868 after dhall-lang/dhall-lang#1026 is merged. Meanwhile, I'll keep it as it currently is.

dhall-docs/src/Dhall/Docs/Html.hs
german1608: I think that at the begginning I passed the whole tuple but it's not like that anymore - I'll split them

dhall-docs/src/Dhall/Docs.hs
german1608: I think that in this moment no scenario is going there - I tend to always fill all the possible branches. I can remove it if you think its better.

german1608: All of the other branches start with `strippedHeader`, so all of them are stripped.

german1608: mmm but text inside comments might be have whitespaces as well. I'll fix it

sjakobi: If the branch would only be used in case of a programmer mistake, we could add something like this:

```suggestion
        | otherwise = error $ "Unexpected header:\n" ++ Text.unpack h
```

PR #1871, opened 2020-06-21
feat(dhall-docs): add some dhall packages documentation to hydra pipelines

german1608:
As @Gabriel439 suggested, in order to avoid sending screenshots everytime at PRs about how the ui/ux currently looks, we can add another step to our hydra pipeline to build several dhall packages and serve them.

This PR attempts to do so. I apologize if I make sone rookie mistakes on `nix`, this was my first encounter with the language and I'm still learning.

@Gabriel439 I think you're the one that should review this PR since it looks like you have the most experience with it in the project. I'm open to any comments.

This PR also includes the `--version` flag on `dhall-docs`, I added it to test some stuff on `nix build` command.


german1608:
@Gabriel439 the build succeded, but I don't know how to see the link to generated docs. Maybe I misunderstood you about this task?

nix/dhall-docs-gen.nix
```diff
@@ -0,0 +1,18 @@
...
+  src = fetchurl {
+    url = "https://github.com/dhall-lang/dhall-lang/archive/v17.0.0.zip";
+    sha256 = "8c46f81d6131bf5dac715a3f107d0a43f4e8decaadd7ed36fbc4375b5b32eeaf";
+  };
+
+  buildInputs = [ dhall-docs unzip ];
+
+  installPhase = ''
+    mkdir -p $out/nix-support/hydra-build-products
+    dhall-docs --input ./Prelude --output $out/nix-support/hydra-build-products/Prelude > $out/nix-support/hydra-build-products/dhall-docs.log
```
Gabriel439, line 15: I believe if you add this line then Hydra will server the generated HTML and the log from its user interface:

```
echo "report html $out/nix-support/hydra-build-products/Prelude/index.html" >> "$out/nix-support/hydra-build-products"
echo "report log $out/nix-support/hydra-build-products/dhall-docs.log" >> "$out/nix-support/hydra-build-products"
```

german1608: I had to `report html` the folder where docs are generated, and it works! If you go here https://hydra.dhall-lang.org/build/62246 you can see the html report link, which is: https://hydra.dhall-lang.org/build/62246/download/1/docs/

I think that suffices the task. Currently the prelude at master doesn't have any .dhall files, so little to no files are actually generated.

german1608:
@Gabriel439 For the moment, I only added the prelude. `dhall-kubernetes` may be added for perf tests, but it acutally doesn't serve any docs since that project is automatically generated.

I may create another repo (or a directory in the `dhall-docs` package) that includes all of the features that we are aiming, like docs on headers, on attributes, etc

german1608:
I tested the preview on `master` and works like a charm :star2:

PR #1875, opened 2020-06-22
Version 1.33.0 → 1.33.1

Gabriel439:


Gabriel439:
The recent work to support `repline-0.4.0` needs to be published to Hackage in order to unblock `dhall-1.33.*` on Stackage since a simple revision will not work in this case

Gabriel439:
This is related to https://github.com/commercialhaskell/stackage/issues/5465

PR #1876, opened 2020-06-23
feat(dhall-docs): Add content-addressed storage for documentation

### LINKED ISSUE #1870: dhall-docs symlinks strategy
I've been thinking a little bit about generating the documentation like nix does for its packages, using symlinks. So I'll use this issue for a first proposal with the details of how it will be done, where docs will be saved and so on.

## Where docs will actually be saved?

`dhall-docs` will have a reserved folder in the filesystem to save its documents. I propose we use `$HOME/.dhall-docs` for it. There can be an environmental variable to override it, I see `DHALL_DOCS_STORE` as a good name.

Now, each package needs to have its own directory inside that `DHALL_DOCS_STORE`. I suggest using the package-name, extracted as the currently the tool does:

* Via the `--package-name` configuration flag, or
* Using the basename of the input directory.

Wo, for a package named `PACKAGE_NAME`, its documentation will be saved on `$DHALL_DOCS_STORE/$PACKAGE_NAME`.

## How it will be done

Actually, it will be the same as we currently do, but saving at `$DHALL_DOCS_STORE/$PACKAGE_NAME`. After that, the tool will wire up the symlinks and configure them in the output folder that the user specified. Note that the `--output` is different from `$DHALL_DOCS_STORE/$PACKAGE_NAME`, the prior just contains symlinks unlike the latest that contains the actual documentation.

i'll be aware of your thoughts

german1608:
Closes #1870 

## Known issues

- [x] If the symlink path already exists and it resolves to another symlink, the `resolveFile'` function of `path-io` resolves to the target of the symlink, and things break therefore.
- [x] I don't calculate the hash of the generated doc to use it as the name of the link. Do you have any idea (package that might help) to do so?

### DIFF
--- a/dhall-docs/src/Dhall/Docs.hs
+++ b/dhall-docs/src/Dhall/Docs.hs
@@ -18,6 +18,7 @@ module Dhall.Docs
 
 import Control.Applicative ((<|>))
 import Data.Monoid         ((<>))
+import Data.Text           (Text)
 import Data.Version        (showVersion)
 import Dhall.Docs.Core
 import Options.Applicative (Parser, ParserInfo)
@@ -111,9 +112,8 @@ defaultMain = \case
         Control.Monad.when outDirExists $ do
             isLink <- System.Directory.pathIsSymbolicLink docLink
             if isLink then System.Directory.removeFile docLink
-            else die $ "The specified --output-link (" <> docLink <> ") already "
-                    <> "exists and its not a symlink. The --output-link flag "
-                    <> "needs to either not exist or to be a symlink"
+            else die $ "The specified --output-link (" <> Data.Text.pack docLink
+                    <> ") already exists and its not a symlink."
 
         resolvedDocLink <- Path.IO.resolveDir' docLink
         let packageName = resolvePackageName resolvedPackageDir
@@ -121,9 +121,9 @@ defaultMain = \case
     Version ->
         putStrLn (showVersion Meta.version)
 
-die :: String -> IO a
+die :: Text -> IO a
 die e = do
-    Text.IO.hPutStrLn System.IO.stderr $ Data.Text.pack e
+    Text.IO.hPutStrLn System.IO.stderr e
 
     System.Exit.exitFailure
 
--- a/dhall-docs/src/Dhall/Docs.hs
+++ b/dhall-docs/src/Dhall/Docs.hs
@@ -113,7 +113,7 @@ defaultMain = \case
             isLink <- System.Directory.pathIsSymbolicLink docLink
             if isLink then System.Directory.removeFile docLink
             else die $ "The specified --output-link (" <> Data.Text.pack docLink
-                    <> ") already exists and its not a symlink."
+                    <> ") already exists and it's not a symlink."
 
         resolvedDocLink <- Path.IO.resolveDir' docLink
         let packageName = resolvePackageName resolvedPackageDir
--- a/dhall-docs/src/Dhall/Docs.hs
+++ b/dhall-docs/src/Dhall/Docs.hs
@@ -24,12 +24,15 @@ import Options.Applicative (Parser, ParserInfo)
 import Path                (Abs, Dir, Path)
 
 import qualified Control.Monad
+import qualified Data.Text
+import qualified Data.Text.IO        as Text.IO
 import qualified GHC.IO.Encoding
 import qualified Options.Applicative
 import qualified Path
 import qualified Path.IO
 import qualified Paths_dhall_docs    as Meta
 import qualified System.Directory
+import qualified System.Exit
 import qualified System.IO
 
 -- | Command line options
@@ -54,9 +57,8 @@ parseOptions =
        <> Options.Applicative.metavar "OUTPUT-LINK"
        <> Options.Applicative.help
             ( "Path to the link targeting the directory with the generated "
-           <> "documentation. If the path exists, is a directory or a symlink "
-           <> "to a directory, its overwritten with the target to the actual "
-           <> "documentation"
+           <> "documentation. The path needs to not exist or to be a symlink, "
+           <> "otherwise the tool won't generate any docs at all"
             )
        <> Options.Applicative.value "./docs" )
     <*> parsePackageNameResolver
@@ -105,16 +107,25 @@ defaultMain = \case
         GHC.IO.Encoding.setLocaleEncoding System.IO.utf8
         resolvedPackageDir <- Path.IO.resolveDir' packageDir
 
-        outLinkExists <- System.Directory.doesDirectoryExist docLink
-        Control.Monad.when outLinkExists $ System.Directory.removeFile docLink
+        outDirExists <- System.Directory.doesPathExist docLink
+        Control.Monad.when outDirExists $ do
+            isLink <- System.Directory.pathIsSymbolicLink docLink
+            if isLink then System.Directory.removeFile docLink
+            else die $ "The specified --output-link (" <> docLink <> ") already "
+                    <> "exists and its not a symlink. The --output-link flag "
+                    <> "needs to either not exist or to be a symlink"
 
         resolvedDocLink <- Path.IO.resolveDir' docLink
-
         let packageName = resolvePackageName resolvedPackageDir
         generateDocs resolvedPackageDir resolvedDocLink packageName
     Version ->
         putStrLn (showVersion Meta.version)
 
+die :: String -> IO a
+die e = do
+    Text.IO.hPutStrLn System.IO.stderr $ Data.Text.pack e
+
+    System.Exit.exitFailure
 
 -- | Entry point for the @dhall-docs@ executable
 main :: IO ()
--- a/dhall-docs/src/Dhall/Docs.hs
+++ b/dhall-docs/src/Dhall/Docs.hs
@@ -3,7 +3,6 @@
 -}
 {-# LANGUAGE LambdaCase        #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE PatternGuards     #-}
 {-# LANGUAGE RecordWildCards   #-}
 
 module Dhall.Docs
@@ -15,53 +14,30 @@ module Dhall.Docs
       -- * Execution
     , main
     , defaultMain
-    , getAllDhallFilesAndHeaders
-
-      -- * Miscelaneous
-    , saveHtml
-    , createIndexes
-    , resolveRelativePath
     ) where
 
 import Control.Applicative ((<|>))
 import Data.Monoid         ((<>))
-import Data.Text           (Text)
 import Data.Version        (showVersion)
-import Data.Void           (Void)
-import Dhall.Docs.Embedded
-import Dhall.Docs.Html
-import Dhall.Docs.Markdown
-import Dhall.Parser        (Header (..), ParseError (..), exprAndHeaderFromText)
+import Dhall.Docs.Core
 import Options.Applicative (Parser, ParserInfo)
-import Path                (Abs, Dir, File, Path, (</>))
-import Text.Megaparsec     (ParseErrorBundle (..))
+import Path                (Abs, Dir, Path)
 
 import qualified Control.Monad
-import qualified Data.ByteString
-import qualified Data.Map.Strict     as Map
-import qualified Data.Maybe
-import qualified Data.Text
-import qualified Data.Text.IO        as Text.IO
 import qualified GHC.IO.Encoding
-import qualified Lucid
 import qualified Options.Applicative
 import qualified Path
 import qualified Path.IO
 import qualified Paths_dhall_docs    as Meta
+import qualified System.Directory
 import qualified System.IO
-import qualified Text.Megaparsec
-
--- $setup
--- >>> :set -XQuasiQuotes
--- >>> import Path (absdir, absfile)
 
 -- | Command line options
 data Options
     = Options
         { packageDir :: FilePath         -- ^ Directory where your package resides
-        , outDir :: FilePath             -- ^ Directory where your documentation
-                                         --   will be placed
-        , packageNameResolver :: Path Abs Dir -> String
+        , docLink :: FilePath            -- ^ Link to the generated documentation
+        , resolvePackageName :: Path Abs Dir -> String
         }
     | Version
 
@@ -74,10 +50,10 @@ parseOptions =
        <> Options.Applicative.metavar "INPUT"
        <> Options.Applicative.help "Directory of your dhall package" )
     <*> Options.Applicative.strOption
-        ( Options.Applicative.long "output"
-       <> Options.Applicative.metavar "OUTPUT"
-       <> Options.Applicative.help "Directory where your docs will be generated"
-       <> Options.Applicative.value "docs" )
+        ( Options.Applicative.long "output-link"
+       <> Options.Applicative.metavar "OUTPUT-LINK"
+       <> Options.Applicative.help "Path for the link to the generated documentation"
+       <> Options.Applicative.value "./docs/index.html" )
     <*> parsePackageNameResolver
     ) <|> parseVersion
   where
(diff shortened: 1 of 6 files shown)


german1608:
Whenever you have time @Gabriel439 @sjakobi @Profpatsch could lend me a hand to solve my known issues please :)

german1608:
Mmm, appveyor complains if `Rel` is not in scope (showing the error in the Qusiquotes `Path.reldir`. hydra complains if `Rel` appears (unused import). 

what should I do?

sjakobi:
> ##
> * If the symlink path already exists and it resolves to another symlink, the `resolveFile'` function of `path-io` resolves to the target of the symlink, and things break therefore.

`resolveFile'` uses [`canonicalizePath`](http://hackage.haskell.org/package/path-io-1.6.0/docs/Path-IO.html#v:canonicalizePath) which AFAICT does the symlink resolving. Maybe you could use [`makeAbsolute`](http://hackage.haskell.org/package/path-io-1.6.0/docs/Path-IO.html#v:makeAbsolute) instead, which AFAICT doesn't do the symlink resolving?

german1608:
> > * If the symlink path already exists and it resolves to another symlink, the `resolveFile'` function of `path-io` resolves to the target of the symlink, and things break therefore.
> 
> `resolveFile'` uses [`canonicalizePath`](http://hackage.haskell.org/package/path-io-1.6.0/docs/Path-IO.html#v:canonicalizePath) which AFAICT does the symlink resolving. Maybe you could use [`makeAbsolute`](http://hackage.haskell.org/package/path-io-1.6.0/docs/Path-IO.html#v:makeAbsolute) instead, which AFAICT doesn't do the symlink resolving?

I thought of that, but let me give it a try. I also filed an issue on [path-io#53](https://github.com/mrkkrp/path-io/issues/53) to see if there is a function or something that does the trick.

dhall-docs/src/Dhall/Docs.hs
```diff
@@ -117,199 +93,20 @@ parserInfoOptions =
...
 defaultMain :: Options -> IO ()
 defaultMain = \case
     Options{..} -> do
         GHC.IO.Encoding.setLocaleEncoding System.IO.utf8
         resolvedPackageDir <- Path.IO.resolveDir' packageDir
-        resolvedOutDir <- Path.IO.resolveDir' outDir
 
-        let packageName = packageNameResolver resolvedPackageDir
+        outLinkExists <- System.Directory.doesFileExist docLink
+        Control.Monad.when outLinkExists $ System.Directory.removeFile docLink
```
german1608, line 104: @sjakobi I ended up doing this

dhall-docs/src/Dhall/Docs/Core.hs
```diff
@@ -0,0 +1,232 @@
...
+
+        dataDir <- getDataDir
+        Path.IO.ensureDir outDir
+        Control.Monad.forM_ dataDir $ \(filename, contents) -> do
+            let finalPath = Path.fromAbsFile $ outDir </> filename
+            Data.ByteString.writeFile finalPath contents
+
+        Path.IO.ensureDir $ Path.parent outLink
+        Path.IO.createFileLink
+            (outDir </> Data.Maybe.fromJust (Path.parseRelFile "index.html"))
```
Gabriel439, line 231: I generally recommend against using partial functions like `fromJust`.  I like to get into the habit of handling all errors with good error messages, even for errors that I believe are impossible (if only to incentivize myself to make invalid states unrepresentable)

german1608: I thought it was ok since it is impossible for `Path.parseRelFile "index.html"` to be `Nothing.

The best we can do here is use the `path` quasiquoters, but on hydra it complains about a missing import (`Rel` from `path) and on appveyor it requires it. I see that appveyor uses `stack`, which uses `ghc-8.6.5`, unlike hydra, that uses `8.6.1`. May this be a bug on ghc or I misunderstood something?

sjakobi: @german1608 Can you give us links to the failing jobs?

It's a weird issue, and this is not a very pretty workaround. (I see you're using `parse{Abs,Rel}{File,Dir}` on literals elsewhere too).

Did you try using template haskell instead of the quasiquoters?

sjakobi: A better alternative could be to disable the warning for the problematic GHC version with a bit of CPP.

german1608: I'll replicate the issues on this PR and sent here the links to builds, gimme a sec

german1608: I didn't tried using template haskell though... I'll give it a try after replicating the errors

german1608: Commit https://github.com/dhall-lang/dhall-haskell/pull/1876/commits/49e59f3b21e1422e03b6942c9ed8fc9f82c639e9 uses quasiquoters, and imports `Rel`.

Its hydra build [fails](https://hydra.dhall-lang.org/build/62671/nixlog/2) because of the redundant import

Its appveyor build [succeeds](https://ci.appveyor.com/project/Gabriel439/dhall-haskell/builds/33699055)

Commit https://github.com/dhall-lang/dhall-haskell/pull/1876/commits/ebb8a3664c2db4e5e871dbd82108ae513a9c70e1 uses quasiquoters, but doesn't import `Rel`.

Its hydra build [succeeds](https://hydra.dhall-lang.org/build/62677)

Its appveyor build [fails](https://ci.appveyor.com/project/Gabriel439/dhall-haskell/builds/33699326) because it can't found `Rel` on scope.

german1608: It also fails using template haskell :(

image

sjakobi: Thanks, @german1608, strange indeed! I wasn't aware that quasiquoters require additional imports! If you're using a *nix OS, can you reproduce the build failure with 8.6.5?

If the `Rel` import is in fact needed, I'd suggest to add `{-# OPTIONS_GHC -Wno-redundant #-}` to the problematic module (possibly behind CPP for the specific GHC version), or not to use `-Werror` by removing this line:

https://github.com/dhall-lang/dhall-haskell/blob/6e1fdd8696856660d74b9eeb41e8a91249d0c8cf/nix/shared.nix#L110

sjakobi: …or use `-Wwarn` in the specicific module to negate `-Werror`… http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/using-warnings.html#ghc-flag--Wwarn

german1608: With *nix you mean NixOS? I'm using ubuntu and `stack` is using `lts-14.27`, which is 8.6.5 already.

Yeah, a solution is to use some CPP but its strange for me that is failing on a greater version.

sjakobi: No, I was wondering whether you could reproduce the issue with a non-Windows OS, i.e. macOS or Linux, some Unix derivative.



> its strange for me that is failing on a greater version.

What do you mean?

german1608: > No, I was wondering whether you could reproduce the issue with a non-Windows OS, i.e. macOS or Linux, some Unix derivative.

Oh I see, well, as I said, I tested on ubuntu. The `-Wwarn` looks like it will work, but I don't like the idea of suppressing warnings. So I'll go with the CPP approach although I don't like to (ab)use it.

> > its strange for me that is failing on a greater version.
> 
> What do you mean?

I mean that nix uses ghc 8.6.1 and it doesn't fail without the import, while stack (appveyor) uses 8.6.5 and it does, which is a greater version AFAIK. That it why its weird for me, although there may a bug in that newer version... or maybe I'm confusing stuff, first time I used haskell pragmas was on this GSoC haha.

dhall-docs/src/Dhall/Docs/Core.hs
```diff
@@ -0,0 +1,238 @@
...
+import Dhall.Docs.Embedded
+import Dhall.Docs.Html
+import Dhall.Docs.Markdown
+import Dhall.Docs.Store
+import Dhall.Parser        (Header (..), ParseError (..), exprAndHeaderFromText)
+import Path                (Abs, Dir, File, Path, (</>))
+import Text.Megaparsec     (ParseErrorBundle (..))
+
+-- QuasiQuotes requires importing `Rel` on ghc 8.6.5
+#if MIN_VERSION_GLASGOW_HASKELL(8,6,5,0)
```
Gabriel439, line 23: I think this solution is dangerous because we don't know for sure that all GHC versions from 8.6.5 onwards require the import and all GHC versions before that don't require the import.  For example, if this change in behavior were actually introduced in GHC 8.6.2 then this would fail to build for that GHC version.

More generally, I think it's a risky solution because we don't know if one behavior or the other was intended by the GHC developers, so there's a risk that the behavior changes again in a future GHC version.

I think the safer solution was the idea of use `{-# OPTIONS_GHC -Wno-unused-imports #-}` that @sjakobi suggested earlier

german1608: Not a big fan of turning off warnings, but you have a good point though.

german1608:
I'm currently missing the hash generation. I'm still seeing the best approach to do so. If you know of a haskell package that does the trick (i'm still unsure about how to use [cas-store](https://hackage.haskell.org/package/cas-store) suggestion from #1870) please let me know.

Gabriel439:
@german1608: The way I'd do it is to:

* First create the documentation underneath a temporary directory
* Hash the temporary directory
* Move the temporary directory contents to the final destination once you know the hash

german1608:
@Gabriel439 Thanks for your suggestion. I ended up implementing `makeHashForDirectory`. In its haddock you can find more impl details

german1608:
Although the build passes, for some reason the hash generated using both `nix` and `stack` on my local machine differs from the one on hydra.

nix/dhall-docs-gen.nix
```diff
@@ -8,12 +8,15 @@ stdenv.mkDerivation rec {
...
-  buildInputs = [ dhall-docs unzip ];
+  buildInputs = [ dhall-docs unzip tree ];
 
   installPhase = ''
     mkdir -p $out/nix-support
-    dhall-docs --input ./Prelude --output $out/docs > $out/dhall-docs.log
-    echo "report html $out/docs" >> $out/nix-support/hydra-build-products
+    export XDG_DATA_HOME=$out/
+    dhall-docs --input ./Prelude > $out/dhall-docs.log
+    htmlDir=$(dirname $(readlink -f ./docs/index.html))
```
german1608, line 17: I managed to do this to avoid using the hardcoded hash.

dhall-docs/src/Dhall/Docs.hs
```diff
@@ -74,10 +50,10 @@ parseOptions =
...
        <> Options.Applicative.help "Directory of your dhall package" )
     <*> Options.Applicative.strOption
-        ( Options.Applicative.long "output"
-       <> Options.Applicative.metavar "OUTPUT"
-       <> Options.Applicative.help "Directory where your docs will be generated"
-       <> Options.Applicative.value "docs" )
+        ( Options.Applicative.long "output-link"
+       <> Options.Applicative.metavar "OUTPUT-LINK"
+       <> Options.Applicative.help "Path for the link to the generated documentation"
+       <> Options.Applicative.value "./docs/index.html" )
```
Gabriel439, line 56: I would create a symlink to the directory rather than a symlink to the `index.html` file.  That makes it easier for the user to access generated files other than `./docs/index.html` from the command line.  This is also the way Nix tends to work

dhall-docs/src/Dhall/Docs/Store.hs
```diff
@@ -0,0 +1,42 @@
...
+    structure of it and the contents of its files, but not the name of the
+    actual files.
+
+    This is done by computing the hash of each file and sorting them by its
+    absolute file name, and computing the hash of the concatenation of all
+    hashes.
+-}
+makeHashForDirectory :: Path Abs Dir -> IO SHA256Digest
+makeHashForDirectory dir = do
+    -- Builds a map so key order is preserved between several calls
```
Gabriel439, line 35: I believe the hash does not change in some cases if the file name changes because the file names are discarded when computing the hash

However, there might be a simpler approach to do this, which is to create an archive from the directory (e.g. using the `tar` package, which supports doing it in memory) and then hash that.  That's essentially the same way that Nix hash directories (except that Nix has its own custom archive format).

german1608: @Gabriel439 Thanks for your suggestion. I'm implementing it, but I see different hashes on different executions.

_I think_ its because the tarballs include the date of file creation of its files, which change on each execution obviously. A solution for this is to set that date to 0 (it is an epoch time). I tried it and now the hashes preserves, although the tar isn't an exact representation of the generated docs because of that change. Are you ok with that change? I'm pushing it so you can review it.

Gabriel439: Yeah, that's fine to set the date to 0.  This would actually be equivalent to what Nix does as well, too 🙂

dhall-docs/src/Dhall/Docs/Store.hs
```diff
@@ -0,0 +1,44 @@
...
+import qualified Path
+import qualified Path.IO
+
+{-| Fetches the dhall-docs home directory. If @XDG_DATA_HOME@ env var is
+    defined, then @${XDG_DATA_HOME}/dhall-docs@ will be returned. Otherwise,
+    "${HOME}/.local/share/dhall-docs"
+-}
+getDocsHomeDirectory :: IO (Path Abs Dir)
+getDocsHomeDirectory = do
+    dir <- Path.IO.getXdgDir Path.IO.XdgData $ Path.parseRelDir "dhall-docs"
```
sjakobi, line 22: I was confused by the use of `parseRelDir` on a literal here. I now understand hat the result is a `Maybe`:

http://hackage.haskell.org/package/path-io-1.6.0/docs/Path-IO.html#v:getXdgDir

I guess changing this to a quasiquote or TH might again give us trouble with `-Wunused-imports`?

dhall-docs/src/Dhall/Docs.hs
```diff
@@ -74,10 +50,15 @@ parseOptions =
...
-       <> Options.Applicative.help "Directory where your docs will be generated"
-       <> Options.Applicative.value "docs" )
+        ( Options.Applicative.long "output-link"
+       <> Options.Applicative.metavar "OUTPUT-LINK"
+       <> Options.Applicative.help
+            ( "Path to the link targeting the directory with the generated "
+           <> "documentation. If the path exists, is a directory or a symlink "
+           <> "to a directory, its overwritten with the target to the actual "
+           <> "documentation"
+            )
```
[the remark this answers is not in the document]
german1608: I'm changing it so it only removes links to directories. If its just a directory, I'll throw an error

dhall-docs/src/Dhall/Docs/Store.hs
german1608: Probably it will give use those errors. But since that module will be really small, I think the change is worth to do.

dhall-docs/src/Dhall/Docs.hs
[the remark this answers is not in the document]
german1608: It works on both directories and files that are not symlinks.

dhall-docs/src/Dhall/Docs.hs
```diff
@@ -117,202 +100,32 @@ parserInfoOptions =
...
-            Control.Monad.forM_ dataDir $ \(filename, contents) -> do
-                let finalPath = Path.fromAbsFile $ resolvedOutDir </> filename
-                Data.ByteString.writeFile finalPath contents
+        outDirExists <- System.Directory.doesPathExist docLink
+        Control.Monad.when outDirExists $ do
+            isLink <- System.Directory.pathIsSymbolicLink docLink
+            if isLink then System.Directory.removeFile docLink
+            else die $ "The specified --output-link (" <> docLink <> ") already "
+                    <> "exists and its not a symlink. The --output-link flag "
+                    <> "needs to either not exist or to be a symlink"
```
sjakobi, line 116: The wording regarding the existence of the flag is a bit awkward. Maybe just shorten it?!

```suggestion
            else die $ "The specified --output-link (" <> docLink <> ") already "
                    <> "exists and it's not a symlink."
```

dhall-docs/src/Dhall/Docs.hs
```diff
@@ -105,16 +107,25 @@ defaultMain = \case
...
 
         resolvedDocLink <- Path.IO.resolveDir' docLink
-
         let packageName = resolvePackageName resolvedPackageDir
         generateDocs resolvedPackageDir resolvedDocLink packageName
     Version ->
         putStrLn (showVersion Meta.version)
 
+die :: String -> IO a
```
sjakobi, line 124: Nit: I think there are benefits to having a single dominant string type in this code base, and it should probably be `Text .

german1608:
@sjakobi @Gabriel439 thanks for your review & help :smile:

Gabriel439:
@german1608: You're welcome! 🙂

PR #1881, opened 2020-06-24
Move RecordEncoder documentation

toonn:
The documentation for how to use the Divisible instance to create a
RecordEncoder ended up in the "Miscellaneous" section. This made it very
hard to find, after some searching I ended up in the repository and
found out the documentation *was* included on the hackage page I was
trying to find it on.

I moved it to document the definition of the RecordEncoder newtype to
bring it in line with the RecordDecoder documentation.

(Note: The diff is slightly misleading but moving the (>*<) definition up is the same thing as moving the haddock comment down.)

### DIFF
--- a/dhall/src/Dhall.hs
+++ b/dhall/src/Dhall.hs
@@ -2471,6 +2471,12 @@ constructor :: Text -> Decoder a -> UnionDecoder a
 constructor key valueDecoder = UnionDecoder
     ( Data.Functor.Compose.Compose (Dhall.Map.singleton key valueDecoder) )
 
+-- | Infix 'divided'
+(>*<) :: Divisible f => f a -> f b -> f (a, b)
+(>*<) = divided
+
+infixr 5 >*<
+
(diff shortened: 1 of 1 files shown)


dhall/src/Dhall.hs
```diff
@@ -2527,13 +2533,6 @@ injectProject =
 
 -}
 
--- | Infix 'divided'
-(>*<) :: Divisible f => f a -> f b -> f (a, b)
-(>*<) = divided
-
-infixr 5 >*<
-
--- | Intermediate type used for building a `ToDhall` instance for a record
```
german1608, line 2536: Why did you remove this line of doc?

german1608: Oh sorry, I understand what you did. I didn't notice that above `(>*<)` there was the doc for `RecordEncoder`.

toonn:
Yeah, the diff isn't obvious. I *did* remove that one line of documentation because it didn't seem to add much of value and the RecordDecoder docs don't mention the equivalent.

PR #1895, opened 2020-06-29
feat(dhall-docs): add breadcrumbs to improve navigation

german1608:
First attempt, but I think I can do better. Handling the paths is a little bit cumbersome.

I think I can improve it further:

- [x] Add the index of the package link at the beginning of the breadcrumbs
- [x] Add a "click to copy" button at the end of the breadcrumbs to copy the breadcrumbs so user can load the expression in their dhall files as `$HOST/$BREADCRUMBS`


german1608:
Usually, you can check current result at:

https://hydra.dhall-lang.org/job/dhall-haskell/1895/generate-dhall-docs/latest/download/1/docs

when hydra build passes

german1608:
I think I can improve it further:

* Add the index of the package link at the beginning of the breadcrumbs
* Add a "click to copy" button at the end of the breadcrumbs to copy the breadcrumbs so user can load the expression in their dhall files as `$HOST/$BREADCRUMBS`

dhall-docs/src/Dhall/Docs/Html.hs
```diff
@@ -121,17 +127,71 @@ indexToHtml indexDir files dirs params@DocParams{..} = doctypehtml_ $ do
...
         Nothing -> file
         Just (f, _) -> f
 
-    title :: String
-    title = indexDir <> " index"
+    breadcrumbs = relPathToBreadcrumb indexDir
+    htmlTitle = breadCrumbsToText breadcrumbs
+    pageTitle = breadCrumbsToHtml Index breadcrumbs
+
+-- | ADT for handling bread crumbs. This is essentially a backwards list
```
Gabriel439, line 143: Were there any issues with using a non-backwards list?

german1608: Yes, the main issue is how to calculate the `../` path (which ends up being different depending if the file is an `index.html` or a `.dhall.html` file). If it were a non-backwards list, then:

* it might need an extra attribute, reflecting how many times do I have to `replicate` `../`, or;
* Traverse the whole breadcrumbs to calculate this in order to add another function helper to convert it to html with links set-up properly.

I did so and I didn't like how the final code, so I changed to a backwards list:

* I don't need to add any other attribute. The breadcrumbs is just a semantic list-like ADT.
* I can start at either 0 or -1 depending if breadcrumb is for an index.html or `.dhall.html` file, and keep incrementing that number at the same time we traverse it.

Gabriel439: Alright, that sounds reasonable to me

PR #1910, opened 2020-07-10
dhall-docs: filter non dhall file in generateDocs

### LINKED ISSUE #1909: dhall-docs: fail to run when package has binary file
It seems like dhall-docs doesn't filter non `.dhall` files by default and it may failed with this error:
```
.git/index: hGetContents: invalid argument (invalid byte sequence)
```

TristanCacqueray:
This change ensure dhall-docs doesn't halt with this exception
when running on a git project:
 .git/index: hGetContents: invalid argument (invalid byte sequence)

Fixes #1909

### DIFF
--- a/dhall-docs/src/Dhall/Docs/Core.hs
+++ b/dhall-docs/src/Dhall/Docs/Core.hs
@@ -17,6 +17,7 @@
 module Dhall.Docs.Core (generateDocs, generateDocsPure, GeneratedDocs(..)) where
 
 import Control.Monad.Writer.Class (MonadWriter)
+import Data.ByteString            (ByteString)
 import Data.Function              (on)
 import Data.Map.Strict            (Map)
 import Data.Monoid                ((<>))
@@ -54,6 +55,7 @@ import qualified Data.Map.Strict            as Map
 import qualified Data.Maybe
 import qualified Data.Maybe                 as Maybe
 import qualified Data.Text
+import qualified Data.Text.Encoding
 import qualified Data.Text.IO               as Text.IO
 import qualified Data.Text.Lazy             as Text.Lazy
 import qualified Dhall.Core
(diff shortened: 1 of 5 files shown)


dhall-docs/src/Dhall/Docs/Core.hs
```diff
@@ -381,7 +382,7 @@ generateDocs
     -> IO ()
 generateDocs inputDir outLink packageName = do
     (_, absFiles) <- Path.IO.listDirRecur inputDir
-    contents <- mapM (Text.IO.readFile . Path.fromAbsFile) absFiles
+    contents <- mapM (Text.IO.readFile . Path.fromAbsFile) (filter hasDhallExtension absFiles)
```
german1608, line 385: I'd prefer to keep the filtering in just a single place (i.e. `getAllDhallFiles`)

I've tried your patch locally and it fails with binary files that have non-utf8 characters even with files with ".dhall" extension.

I've faced this issue before, and the solution was to:

1. Read files as `ByteString` i.e. using `ByteString.readfile`
2. Encode as `Text` utf-8 i.e. using `Data.Text.Encoding.decodeUtf8'`
3. Filter only the files that don't fail to decode without filtering again by .dhall extension. As I said above, the `.dhall` filtering is done by `getAllDhallfiles`.

german1608: If you don't have time for the patch I can work on it later :smile: I'm currently working on #1908 and it will take some time.

german1608:
Also, I think that we need to improve the test-setup to:

1. Add two new test cases: 
    * A file that doesn't have a .dhall extension
    * A file that has a .dhall extension _but_ has non-utf8 characters.
2. Ensure that the tool doesn't fail and no-documentation was generated for the file without the `.dhall` extension.
    * For the non-utf8 we might need to update the test-suite a little because currently it uses `generateDocsPure` instead of `generateDocs`.
    * For the non `.dhall` extension we can compare the amount of golden files vs the amount of files generated by the tool.

dhall-docs/src/Dhall/Docs/Core.hs
TristanCacqueray: I don't understand, `getAllDhallFiles` is not called by `generateDocs` which seems to be failing when running dhall-docs on binary files. `generateDocs` is already using `Text.IO.readFile`, perhaps we could simply ignore the files that raised the invalid byte sequence exception?

german1608: `generateDocs` calls `generateDocsPure` which calls `getAllDhallFiles`.

That fix fails even with binary files with `.dhall` extension. The steps I pointed you in my comment are how I fixed this error the time I encountered it (weeks ago) and what probably would work to fix the issue.

`Data.Text.IO.readFile` fails reading binary files, `Data.ByteString.readFile` doesn't. You can read the file using the latest and try to decode it as utf8 using `Data.Text.Encoding.decodeUtf8'` to also do the filtering.

TristanCacqueray: I see, thank you very much for the detailed explanation.

dhall-docs/src/Dhall/Docs/Core.hs
```diff
@@ -422,7 +429,7 @@ generateDocs inputDir outLink packageName = do
 -}
 generateDocsPure
     :: Text                    -- ^ Package name
-    -> [(Path Rel File, Text)] -- ^ (Input file, contents)
+    -> [(Path Rel File, ByteString)] -- ^ (Input file, contents)
```
german1608, line 432: Really clever to update this to `ByteString`, that way all filtering is done in a single place.

TristanCacqueray: I just obeyed ghc error message :-)

PR #1915, opened 2020-07-11
Load italic and bold Lato fonts

SiriusStarr:
Currently, `dhall-docs` loads the Fira Code and Lato fonts using [Google fonts' v2 API](https://developers.google.com/fonts/docs/css2), but only the regular Lato font is loaded.  Since `dhall-docs` supports markdown, italic and bold display should be expected but they currently don't display properly.  This PR adds the italic and bold versions of Lato to the fonts loaded.

Before:

After:


SiriusStarr:
Unless I single-handedly broke AWS, I don't think that CI failure was my fault.  XD

german1608:
I'll restart appveyor build, the error looks weird to me

PR #1916, opened 2020-07-12
Alphabetize exported packages in dhall-docs

SiriusStarr:
Due to a `foldl` in `dirToDirsMap` (in `createIndexes`), exported packages were listed in reverse alphabetical order, which was unexpected behavior given that exported files were in alphabetical order.  Switching it to a `foldr` fixes the inconsistency.

### DIFF
--- a/dhall-docs/src/Dhall/Docs/Core.hs
+++ b/dhall-docs/src/Dhall/Docs/Core.hs
@@ -307,7 +307,7 @@ createIndexes packageName files = map toIndex dirToDirsAndFilesMapAssocs
         documentation to get more information.
     -}
     dirToDirsMap :: Map (Path Rel Dir) [Path Rel Dir]
-    dirToDirsMap = Map.map removeHereDir $ foldl go initialMap dirs
+    dirToDirsMap = Map.map removeHereDir $ foldr go initialMap dirs
       where
         -- > removeHeredir [$(mkRelDir "a"), $(mkRelDir ".")]
         --   [$(mkRelDir "a")]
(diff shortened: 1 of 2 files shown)


dhall-docs/src/Dhall/Docs/Core.hs
```diff
@@ -323,8 +323,8 @@ createIndexes packageName files = map toIndex dirToDirsAndFilesMapAssocs
         initialMap :: Map (Path Rel Dir) [Path Rel Dir]
         initialMap = Map.fromList $ map (,[]) dirs
 
-        go :: Map (Path Rel Dir) [Path Rel Dir] -> Path Rel Dir -> Map (Path Rel Dir) [Path Rel Dir]
-        go dirMap d = Map.adjust ([d] <>) (key $ Path.parent d) dirMap
+        go :: Path Rel Dir -> Map (Path Rel Dir) [Path Rel Dir] -> Map (Path Rel Dir) [Path Rel Dir]
+        go d dirMap = Map.adjust ([d] <>) (key $ Path.parent d) dirMap
```
german1608, line 327: IAnother solution would be to switch the adjust function here:

```suggestion
        go d dirMap = Map.adjust (<> [d]) (key $ Path.parent d) dirMap
```

but better stick with `foldr`

SiriusStarr: Guilty as charged.  ;-)

Your call on which solution you prefer, but `foldr` is maybe more "canonical" to my mind for keeping the order.

german1608: Let's stick with `foldr`. In almost all cases it performs better than `foldl`.

PR #1921, opened 2020-07-13
Move man/dhall.1 to data files

vmchale:
This means that after running `cabal install dhall` (from e.g. hackage) the manpage will be installed to `share/...`. This is what `hlint` does, for instance.

### DIFF
--- a/dhall/dhall.cabal
+++ b/dhall/dhall.cabal
@@ -22,11 +22,12 @@ Description:
     .
     Read "Dhall.Tutorial" to learn how to use this library
 Category: Compiler
+Data-Files:
+    man/dhall.1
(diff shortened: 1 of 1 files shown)


vmchale:
Well, it's common enough :p, see e.g. https://github.com/ndmitchell/hlint/blob/master/hlint.cabal#L23

Gabriel439:
@sjakobi: My understanding is that this is an idiomatic use of `data-files`.  The description for the option is misleading, but anything under `data-files` gets installed under a `share` directory, which plays a role analogous to `/usr/local/share`.  So in other words, `./docs` and `./man` subdirectories are normal and expected underneath such a `share` directory.

Issue #1918, opened 2020-07-13
dhall-docs: type annotation is only added to the index when defined in a `let` binding


TristanCacqueray:
Not sure if this is currently expected, but currently it seems like the type annotations are only listed when they are defined in a `let` binding, for example this `Type.dhall` file doesn't have annotation in the index:

```dhall
{ name : Text } : Type
```

And this file do have the annotation in the index:

```dhall
let type : Type = { name : Text }
in type
```

german1608:
@TristanCacqueray: this is the expected behaviour. `dhall-docs` doesn't do any type-checking (yet) because it may complicate the code a little, and it's not necessary to begin.

We extract the type from the let-binding. I believe that the `dhall-docs` README says so:
https://github.com/dhall-lang/dhall-haskell/blob/master/dhall-docs/README.md

Actually, if you have something like this:

```dhall
let b = Text
let a : b = "foo"
in a
```

it will show `b` on the index.

I'm not sure if we will add the actual type soon. Think of a Dhall file like this:

```dhall
let type : ./MyImported.dhall = 1
in type
```
in the rendered HTML it will put the `./MyImported.dhall` which will allow us (soon) to jump to that definition. But if we do actual type-checking, that will be normalized to an actual type.

Below I'll write a pair of improvements to this feature
### Redirect to variable until we find a syntactic valid type

This will solve the `let-bindings` problem that I showed above. Using the same example, if the type of a binding `a` is another binding `b`, the type of `a` is the value of `b` and so on.

### Support a subset of Dhall expressions in addition to `let-bindings`

Maybe `Record` could be also treated? These are some cases that I'm thinking of:

```dhall
{ a : Text }
```
... will show `{ a : Text }` on the index;

```dhall
let b = Text
in { a : b }
```

... will show `{ a : Text }`. (same indirection I suggested above)

The other expressions I'm think we can support are:

* Literals
* Function types

Gabriel439:
@german1608: My understanding is that @TristanCacqueray is requesting that if a file contains a simple expression of the form `e : T` that the type annotation for that file is `T` (correct me if I'm wrong).  That does not require type inference

TristanCacqueray:
@Gabriel439 That is correct as this would be easier to add documentation for simple file without using an extra `let .. in ..`.

SiriusStarr:
I agree; I've been having to go through and move everything to `let` bindings when it just makes it excessively verbose.  If files like:

```dhall
[...]

PR #1928, opened 2020-07-17
Add `dhall-docs` support for non-`let` type annotations

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/1918


Gabriel439:
@german1608: You're welcome! 🙂

PR #1929, opened 2020-07-19
feat(dhall-docs): Specification and implementation of the dhall-docs comments format

german1608:
This PR specifies and will implement the `dhall-docs` comments format.

I haven't started the implementation yet since I'd like to see your thoughts about my proposal, which is documented on the README of the tool on this PR.

Please let me know what you think and let's try to reach to an agreement before starting the implementation. I took the inspiration from the discuss in #1868 

cc: @sjakobi @Gabriel439 @Profpatsch


dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+ --      baz
+```
+
+... `dhall-docs` will extract the following contents:
+
+```
+foo
+bar
+    baz
+```
```
sjakobi, line 230: These rules seem quite different than those for block comments, but after thinking about them for a bit, I haven't come up with anything better. :+1:

dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+--  bar
+
+```
+
+`dhall-docs` will extract:
+
+```
+foo
+bar
+```
```
sjakobi, line 266: I'd rather suggest treating `bar` as a separate non-documentation comment. Otherwise I think we'll find "internal" comments appended to the documentation.

dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+
+... will extract the contents of the comment, stripping the `{` leading whitespace
+from all other lines, returning the following text:
+
+```
+foo
+  bar
+    baz
+```
+
```
sjakobi, line 126: I think "indentation" should be defined more clearly. Can it be a mix of tabs and spaces?

dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+    }
+
+    {
+      -- | and here
+      pun,
+
+      -- here too
+      b = True
+    }
+    ```
```
sjakobi, line 336: I think we should maybe wait with this until we've worked out what `dhall-docs` will do with these annotations.

dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+* Before each key-value pair of a record type and literal:
+
+    ```dhall
+    {
+      --| here
+      a : T,
+    }
+
+    {
+      -- | and here
```
sjakobi, line 330: That's illegal!!! ;)

```suggestion
      --| and here
```

dhall-docs/README.md
```diff
@@ -1,6 +1,6 @@
 # `dhall-docs`
 
-:construction: **This tool is on development phase yet. It is not usable right now.**
+:construction: **This tool is on development phase yet. You can use it but you should expect a lot of breaking changes on each new release until the final release.**
```
sjakobi, line 3:
```suggestion
:construction: **This tool is on development phase yet. You can use it but you should expect a lot of breaking changes on each new release until the 1.0 release.**
```

?

dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+  {-|
+  foo
+      bar
+          baz
+  -}
+  myAwesomeFunction
+}
+```
+
+If a block comment doesn't starts with `{-|` it will be ignored by `dhall-docs`.
```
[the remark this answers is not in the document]
sjakobi: (Actually my suggestion doesn't seem so much clearer… feel free to ignore)

dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+{-| foo
+    bar
+    -}
+```
+
+is invalid.
+
+See that the closing brace (`-}`) can be in a new line or in the last one:
+`dhall-docs` will, in any case, insert a newline character (`\n`) in the last
+line.
```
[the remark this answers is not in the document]
german1608: Now reading it I see that I didn't reflect my thoughts about it.

If the `-}` is on a new line e.g.

```
{-|
a
-}
```

or like this:

```
{-|
a-}
```


the text will be extracted like this `"a\n"`. My thought is that it would be good for the user that he knows he can place the `-}` at any place and the result will be the same.

dhall-docs/README.md
german1608: yeah, they are quite different. The main difference is indentation treatment on both of them, but couldn't think any better that is easy to write and to implement than this.

dhall-docs/README.md
german1608: The Dhall standard says it's valid:

https://github.com/dhall-lang/dhall-lang/blob/5fde8ef1bead6fb4e999d3c1ffe7044cd019d63a/standard/dhall.abnf#L167-L169

dhall-docs/README.md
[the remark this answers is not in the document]
german1608: I'll ignore your suggestion but I found a typo on mine: `starts` -> `start`.

german1608:
> How about inviting feedback from the wider community?

@sjakobi I wrote a post at discourse here: https://discourse.dhall-lang.org/t/dhall-docs-comments-format/302

german1608:
I also tweeted the discourse post mentioning @dhall_lang to get more audience.

dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+
+and
+
+```dhall
+{-|
+foo
+-}
+```
+
+will both extract `"foo\n"`.
```
sjakobi, line 171: Is the trailing `\n` relevant for the generated documentation? Otherwise this seems unnecessarily detailed. Maybe just say that the two examples are equivalent with regard to the generated documentation.

sjakobi, line 171: Also, is there a difference between

```
{-|
foo
-}
```

and

```
{-|
foo

-}
```

?

dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+foo-}
+```
+
+and
+
+```dhall
+{-|
+foo
+-}
+```
```
Gabriel439, line 169: Perhaps not insert an implicit newline for the former case.  The only reason I suggest this is so that the documentation format more closely follows how multi-line strings work in Dhall.

dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+
+A block comment that has a line with less indentation than the `{` character
+column one is not a valid `dhall-docs` comment. The following:
+
+```dhall
+    {-|
+    foo
+  bar
+        baz -}
+```
```
Gabriel439, line 140: What about blank lines without leading whitespace?  Are those lines ignored for the purpose of computing indentation?

The reason I ask is that multi-line strings will ignore such lines, and it might be worth documenting whether documentation comments do the same.  Ideally, the more the documentation format should resemble how multi-line strings work in Dhall

dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+from all other lines, returning the following text:
+
+```
+foo
+  bar
+    baz
+```
+
+The indentation level of a line is the number of whitespace characters (i.e. `' '`) from
+the first column until the column of the `{`. The rest of the whitespace that
```
sjakobi, line 128: If tabs and space character are interchangeable, that means that e.g.

```
	{-|
 foo
 -}
```

is well-formed, right?

I don't know whether this could cause problems, but I think it might be worth spelling out.

dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+--      baz
+```
+
+The `--` prefix defines the indentation for any line. That is, on the above
+example and the following
+
+```dhall
+--| foo
+  --  bar
+ --      baz
```
Gabriel439, line 242: I think the documentation generator is being too lenient with respect to indentation.  I would prefer that the documentation generator enforces the recommended comment layout and rejects everything else, otherwise it will be easy for people to inadvertently commit the wrong docs.

Also, in my experience, it's easier to start off strict and then make things more lenient later on in response to user feedback than to do things the other way around (start off too lenient and then make things more strict)

dhall-docs/README.md
```diff
@@ -1,6 +1,6 @@
 # `dhall-docs`
 
-:construction: **This tool is on development phase yet. It is not usable right now.**
+:construction: **This tool is on development phase yet. You can use it but you may expect breaking changes on each new release until the 1.0 release.**
```
Gabriel439, line 3: I would personally not worry keeping all breaking changes before the 1.0 release.  Any tool that has enough users will require a slow stream of slightly breaking changes over time to keep up with their evolving needs.

The rule of thumb I always use is that the users decide when a given tool is out of beta, not the author.  This is one of the reasons I always make the first published version of my packages 1.0.0, in recognition of the fact that users might put my tool into production regardless of whether or not I warn them not to.

dhall-docs/README.md
Gabriel439: I think a lot of these issues would be fixed by using the exact same rules as multi-line strings in Dhall.  Then there's no confusion

dhall-docs/README.md
german1608: Mmm now that you say it I'm unsure. I'll remove it and add it if it's necessary.

dhall-docs/README.md
german1608: I agree. I'll update the convention to enforce the alignment of `--`

dhall-docs/README.md
german1608: I'll remove that fragment of the documentation. I think it confuses the user.

dhall-docs/README.md
german1608: Good catch. I believe we should ignore them as well.

dhall-docs/README.md
german1608: I'd say they are different. Or at least the tool should not strip the newlines. My reasoning is that Markdown is a little sensitive to whitespaces, and after thinking a little it would be better to do the less whitespace stripping on the text. The only exception would be the prefix whitespace for indentation, that is mostly discussed on the proposal.

dhall-docs/README.md
sjakobi: I think Gabriel has a point here. If we can specify how a doc comment corresponds to a multi-line string, we only need to add that the corresponding string is supposed to be the input for Markdown parser, and we're done with a good part of this spec.

I guess the tricky bit is that the multi-line spec isn't exactly super easy to understand itself.

How familiar are you with it at this stage, @german1608?

german1608: @sjakobi: yes, that example is well formed.

@Gabriel439: I thought about sing something similar to multiline strings for this purpose, but I was thinking about something more strict for the comments and ended up using {-| as base of indentation.

Another idea that I had was to use the first line column as a base of indentation, that is more similar to multiline strings. But it will allow users to write things such as:

```
{-|
                  Foo
                      BAR
-}
```

That doesn't look to good IMO

dhall-docs/README.md
german1608: I understand your suggestion, thanks!

I think I'll leave the warning about the current status of the tool, it may be a good notice to people though:

```suggestion
:construction: **This tool is on development phase yet**
```

german1608:
@sjakobi @Gabriel439: I was offline last 3 hours and after thinking about all your comments I think that I'll update the block comments syntax to be as similar as the multi-line strings spec. Also, we could reuse some code from the multi-line strings from the `dhall` package to get this done faster.

That will mean that indentation won't be relative to the opening `{`, but will be the longest common indent prefix. I'll ping you when I finish it.

One thing: Although I'd like to get feedback from the wide community, I feel I'm a little late on the GSoC schedule to wait a long time before coding the implementation. What about waiting until tomorrow at 16:00 VE i.e. **Monday 19, 20:00 UTC**? OTOH, I think that writing block comments docs as Dhall multi-line strings will be really familiar to users, I think there won't be a lot of rejection to the proposal.

dhall-docs/README.md
```diff
@@ -97,6 +60,281 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+```dhall
+{-|
+# My awesome package
+
+This is the header of my package
+
+* foo
+  - bar
+  - baz
+
```
TristanCacqueray, line 185: It may be useful to include some more common usage such as:

* Bold, italic
* Internal (anchor) link
* Hypertext link
* Table (is the CommonMark parser support them?)

german1608: Since the markup language is based on Markdown (specifically, CommonMark), those features will be supported out-of-the-box. We could write a sample of that on that example, though.

The only thing that we need to discuss is how to link to other files on the package, but we could do that in a separate issue.

TristanCacqueray: If I understand correctly, CommonMark does not yet have a syntax for tables, accoding to this [thread](https://talk.commonmark.org/t/tables-in-pure-markdown/81), and for example when using Pandoc you need to activate the `githubMarkdownExtensions` when parsing CommonMark.

The samples doesn't have to be exhaustive, for example the `foo bar baz` sample could be replaced by:

<pre>
* a list item in *bold*
  - a sub item in _italic_
  - a [link](http://example.com)
</pre>

german1608: Oh sorry!

mmark, out markdown parser, doesn't support them AFAIK, but we could create a custom extension to add that feature if it's needed.

german1608:
@sjakobi @Gabriel439 and @TristanCacqueray (since you've already participated on the PR)

I basically rewrote the whole block comments section to make it similar to how we handle multi-line strings indentation. On the single-line comments section, I wrote other edge cases that I believe will be good to disambiguate right now. I'd say that the spec is in better shape now.

dhall-docs/README.md
```diff
@@ -97,6 +60,377 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+```
+
+the following text will be extracted:
+
+```dhall
+baz
+qux
+```
+
+The set of `--` lines needs to be aligned, meaning that a set of lines like this:
```
sjakobi, line 346: "aligned" is slightly vague. Is this aligned?

```
	--| foo
 -- bar
```

What does it mean when before the `--|` there are [fullwidth characters](https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms)?

For simplicity, I think we can demand that each `--` is preceded by the same number of characters (if that's well-defined?!). Or we could give up the alignment requirement again…

dhall-docs/README.md
```diff
@@ -97,6 +60,377 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+{-|
+ qux -}
+```
+
+the extracted text will be the following:
+
+```
+foo
+bar
+```
```
sjakobi, line 382: As mentioned above, I think we should treat subsequent doc comments as errors.

Can we deduplicate these sections BTW?

dhall-docs/README.md
```diff
@@ -97,6 +60,377 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+
+A `dhall-docs` block comment needs to start with `"{-|\n"` i.e. a `{-|` and a
+newline. This means that the actual documentation _must_ start on the following line.
+
+[CommonMark] supports some features that are sensitive to indentation (e.g.
+[indented-code-blocks]). Indentation on `dhall-docs` block comments is really
+similar to [Dhall multi-line strings]: the indentation is determined by the
+longest common whitespace prefix between the lines on the block comment. The
+main difference is that, unlike the closing pair of double single quotes on multi-line
+strings (`''`), the closing brace (`-}`) doesn't alter indentation.
```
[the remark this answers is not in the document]
german1608: From my POV it would be annoying if me, a user, should care about the position of the closing brace `-}`

dhall-docs/README.md
```diff
@@ -97,6 +60,377 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+```dhall
+{-|
+  foo
+-}
+
+{-|
+-}
+```
+
+the extracted text will only contain `foo`.
```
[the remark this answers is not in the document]
german1608: Ok, but I'd say that we should still keep ignoring non-doc comments after a doc comment.

What about something like this:

```
{-|
foo -}
-- bar
{-|
baz -}
```
should be an error?

dhall-docs/README.md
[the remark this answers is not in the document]
sjakobi: Hmm, I'm not convinced that this is important enough that it would be worth breaking the analogy for. There's also at least one case where the significance of the position of the `-}` would be useful:

```
{-
    indented
    code
    block
-}
```

dhall-docs/README.md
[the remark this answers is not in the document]
sjakobi: Yeah, I think that should be an error too. It's just not clear which doc comment the user intended to be used by `dhall-docs`.

AFAIK haddock concatenates the doc comments in such cases, but I think that's error prone and not really useful in practice.

dhall-docs/README.md
german1608: I agree with that definition of alignment. I'd say that will be enough for almost all cases.

I don't have a problem with no enforcing the alignment, but following @Gabriel439 comment ([here](https://github.com/dhall-lang/dhall-haskell/pull/1929#discussion_r456917946)) we should start with something more strict and make it more flexible depending on user requests.

dhall-docs/README.md
german1608: Sure! I'll add another section on the document

dhall-docs/README.md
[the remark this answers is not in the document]
german1608: Good point :) then a `dhall-docs` block comment will be basically a multi-string :+1:

dhall-docs/README.md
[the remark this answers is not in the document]
german1608: > Yeah, I think that should be an error too. It's just not clear which doc comment the user intended to be used by `dhall-docs`.

Ok, so the rule is: after a `dhall-docs` comment, there can't be any other `dhall-docs` comment following it.

dhall-docs/README.md
```diff
@@ -97,6 +60,377 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+
+```dhall
+{- foo
+     bar
+       baz
+-}
+```
+
+A `dhall-docs` block comment needs to start with `"{-|\n"` i.e. a `{-|` and a
+newline. This means that the actual documentation _must_ start on the following line.
```
sjakobi, line 94: In analogy to multi-line strings, we should also support `\r\n` line-endings:

https://github.com/dhall-lang/dhall-lang/blob/f717c748b3751e5a1b22da8af3db31030e7dcd0d/standard/dhall.abnf#L348

dhall-docs/README.md
german1608: I merged both sections in a single one and added some examples there. The section is `Mixing the two type of comments`

german1608:
I'll improve the wording and grammar of the spec

german1608:
I've made some changes to the structure of the Comment's format section and I think that I'm ready to start the implementation.

What do you think? cc: @sjakobi @Gabriel439

sjakobi:
@german1608 Yeah, the technical aspects look good to me! :)

dhall-docs/README.md
```diff
@@ -97,6 +60,407 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+
+Note that a line that only contains trailing whitespace will be taken into
+account when calculating the common indentation. The line between `foo` and `bar`
+on the following example has leading whitespace (two (2) whitespaces):
+
+```dhall
+  {-|
+    foo
+  
+    bar -}
```
Gabriel439, line 192: It might help to use `␠` or `␣` to show the spaces:

```suggestion
␠␠{-|
␠␠␠␠foo
␠␠
␠␠␠␠bar -}
```

dhall-docs/README.md
```diff
@@ -97,6 +60,407 @@ dhall-docs --input . --output-link ${OTHER_LINK}
...
+
+```dhall
+--| # My header
+--
+--  * 1
+--  * 2
+
+let a --| foo
+      --  bar
+        = True
```
Gabriel439, line 388:
```suggestion
let --| foo
    --  bar
    a = True
```

I mainly suggest this since that's the commenting format we've been recommending so far for `let` bindings

TristanCacqueray:
That's looking great, I'm looking forward using this new format :-)

german1608:
I've made some progress on this.

I have to admit that one of my weaknesses when programming is manipulating text, I tend to write not the best code for it, but I managed to write an implementation of the specification. The overview of the algorithm is:

1. Parse all comments in the header to a `[DhallComment]`. In `Dhall.Docs.Comment` you can see the code for that, is really similar to the code in `Dhall.Parser.Token` for handling comments
2. Filter the comments that have the marker for both block and single line
3. Validate each marked comment
4. Throw warnings on each step depending if a comment was valid or not
5. Returning the only `dhall-docs` comment's text (if there was a valid one) using the procedure described in the specification.


I believe that the code can be improved:

1. The `Dhall.Docs.Comment` module could do more context-sensitive parsing, but logging warnings ended-up easier to handle on `Dhall.Docs.Core`
2. Use more functions from `base` that do what I want to
3. Rethought the whole algorithm, although I spent all day and this is the best I got

I think that block comment's handling is correct, but there is a thing on single-line comments that I'd like to ask you first. For a set of single-line comments like this:

```dhall
--| foo
--bar
--  baz
```

What should `dhall-docs` do? I _believe_ that single-line comments treatment would be easier if we enforce all single-line comments to have the `|`, although that's onerous to type.

Last thing: do not be afraid of being rude on the code-review :) It's completely ok if you have a lot of comments with corrections, as I said above: Every time I try to do complicated stuff with `Data.Text` my brain explodes :exploding_head:

dhall-docs/src/Dhall/Docs/Core.hs
```diff
@@ -230,6 +252,109 @@ getAllDhallFiles = emitErrors . map toDhallFile . foldr validFiles [] . filter h
...
         fromAssertion _ = Nothing
 
+    {-
+    Validates the comments from source code against the `dhall-docs`
+    comment's specification
+
+    Non `dhall-docs` comments will be ignored, but invalid `dhall-docs`,
+    in addition, will log a warning.
+    -}
+    stripCommentSyntax :: Path Rel File -> Text -> GeneratedDocs (Maybe Text)
```
sjakobi, line 262: Could the return type alternatively be `Either DocsGenWarning Text`?

I was thinking about this very nice blog post: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

I'm not yet sure how to best apply its advice here though…

sjakobi:
I'd eventually like to see more tests for line comments BTW.

sjakobi:
…maybe even unit tests.

dhall-docs/src/Dhall/Docs/Core.hs
german1608: I've just read that, really nice! I chose `GeneratedDocs (Maybe Text)` to log `DocsGenWarning` easier. Also, I used `Maybe Text` on `FileComments` since not all files have headers.

I believe that the best application of that blog post's advice is to change the return type of `Dhall.Docs.Comment.parseComment` be `Either CustomParseError DhallComment`. I can give it a try.

german1608:
I believe this is now ready for review :)

german1608:
@Gabriel439 I forgot to tell you: I've updated your suggestion, using the `␣` character on the examples where it's most useful to.

Have you (and @sjakobi) had time to review this?

dhall/src/Dhall/Parser/Combinators.hs
```diff
@@ -1,6 +1,10 @@
 {-# LANGUAGE CPP                   #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings     #-}
+
+{-| This module contains Dhall's parser combinators
+-}
+
 module Dhall.Parser.Combinators where
```
sjakobi, line 8: This module should probably have an export list.

dhall-docs/tasty/data/golden/InvalidConsecutiveComments.dhall.html
```diff
@@ -0,0 +1,45 @@
...
+      /><h3
+      >Source</h3
+      ><div class="source-code"
+        ><pre
+          ><span class="dhall-syntax"
+          >{</span
+          ><span class="dhall-syntax"
+          >=</span
+          ><span class="dhall-syntax"
+          >}</span></pre></div></div></body></html>
```
TristanCacqueray, line 45: nit: would it be possible to ensure that the generated files end with a carriage return?

dhall-docs/src/Dhall/Docs/Comment.hs
```diff
@@ -0,0 +1,305 @@
...
+    whitespace from both the start and begin of the string
+
+    * If @a = 'DhallDocsComment'@ then comment is valid to extract its contents
+    and be rendered on documentation
+    * If @a = 'MarkedComment'@ then comment has the @|@ marker that
+    @dhall-docs@ will be aware of, but this comment may not be a @dhall-docs@
+    comment
+    * If @a = `RawComment`@ then the comment is a raw comment
+-}
+data DhallComment a
```
Gabriel439, line 51: If you enable the `DataKinds` extension, you can do this:

```haskell
data CommentType = DhallDocsComment | MarkedComment | RawComment

data DhallComment (a :: CommentType)
```

... which ensures that `a` can only be one of those three constructors

dhall-docs/src/Dhall/Docs/Comment.hs
```diff
@@ -0,0 +1,305 @@
...
+-- | Extracted text from a valid @dhall-docs@ comment
+newtype DhallDocsText = DhallDocsText Text
+    deriving Show
+
+unDhallDocsText :: DhallDocsText -> Text
+unDhallDocsText (DhallDocsText t) = t
+
+-- | A mirror of "Dhall.Parser.Token".'Dhall.Parser.Token.lineComment' but
+--   returning a 'DhallComment'
+lineCommentParser :: Parser (NonEmpty (DhallComment RawComment))
```
Gabriel439, line 76: Would it be possible to upstream this into the `dhall` package?  That way we don't have to worry about the two implementations of this diverging

german1608:
> Unless I've overlooked it, could you add a test for a valid doc comment with indentation, like
> 
> ```
>   --| foo
>   --  bar
> ```

Done!

dhall-docs/tasty/data/golden/InvalidConsecutiveComments.dhall.html
german1608: These files are not directly generated by the tool, they are the result of generating the HTML Text documentation from the package and passing that result to `HaXml` parser and pretty printer. I guess that a solution would be to append a newline to the result of the prettyprinter.

dhall-docs/src/Dhall/Docs/Comment.hs
```diff
@@ -0,0 +1,305 @@
...
+--   checks if a 'MarkedComment' is a 'DhallDocsComment'. For 'SingleLineComments'
+--   this also removes the prefix lines before the first marked comment
+parseDhallDocsComment :: DhallComment MarkedComment -> Either CommentParseError (DhallComment DhallDocsComment)
+parseDhallDocsComment (BlockComment comment) =
+    BlockComment <$> checkNewlineAfterMarker comment
+  where
+    checkNewlineAfterMarker :: Text -> Either CommentParseError Text
+    checkNewlineAfterMarker t =
+        if any (`Data.Text.isPrefixOf` t) ["{-|\n", "{-|\r\n"] then pure t
+        else Left MissingNewlineOnBlockComment
```
Gabriel439, line 212: This can be simplified a little bit:

```suggestion
    if any (`Data.Text.isPrefixOf` t) ["{-|\n", "{-|\r\n"]
    then BlockComment t
    else Left MissingNewlineOnBlockComment
```

dhall-docs/src/Dhall/Docs/Comment.hs
```diff
@@ -0,0 +1,305 @@
...
+parseMarkedComment :: DhallComment RawComment -> Maybe (DhallComment MarkedComment)
+parseMarkedComment (BlockComment comment)
+    | "{-|" `Data.Text.isPrefixOf` comment = Just $ BlockComment comment
+    | otherwise = Nothing
+
+parseMarkedComment (SingleLineComments ls)
+    | any (("--|" `Data.Text.isPrefixOf`) . snd) ls = Just (SingleLineComments ls)
+    | otherwise = Nothing
+
+-- | Knowning that there is a @dhall-docs@ marker inside the comment, this
```
Gabriel439, line 202:
```suggestion
-- | Knowing that there is a @dhall-docs@ marker inside the comment, this
```

dhall-docs/src/Dhall/Docs/Comment.hs
```diff
@@ -0,0 +1,305 @@
...
+parseDhallDocsComment (SingleLineComments lineComments) =
+    fmap SingleLineComments $ checkAlignment lineComments >>= checkAmountOfMarkers >>= checkPrefixes
+  where
+    sourceCol = Text.Megaparsec.unPos . Text.Megaparsec.sourceColumn
+
+    checkAmountOfMarkers :: ListOfSingleLineComments -> Either CommentParseError ListOfSingleLineComments
+    checkAmountOfMarkers ls =
+        if numberOfMarkers > 1 then Left SeveralSubseqDhallDocsComments
+        else case newLines of
+            [] -> fileAnIssue "checkAmmountOfMarkers failed with newLines = []"
```
Gabriel439, line 223:
```suggestion
            [] -> fileAnIssue "checkAmountOfMarkers failed with newLines = []"
```

dhall-docs/src/Dhall/Docs/Comment.hs
```diff
@@ -0,0 +1,305 @@
...
+            l : remainder -> Right $ l :| remainder
+      where
+        commentLInes = NonEmpty.toList ls
+        numberOfMarkers = length $ filter (Data.Text.isPrefixOf "--|" . snd) commentLInes
+        (_, newLines) = break (Data.Text.isPrefixOf "--|" . snd) commentLInes
```
Gabriel439, line 228:
```suggestion
        commentLines = NonEmpty.toList ls
        numberOfMarkers = length $ filter (Data.Text.isPrefixOf "--|" . snd) commentLines
        (_, newLines) = break (Data.Text.isPrefixOf "--|" . snd) commentLines
```

dhall-docs/src/Dhall/Docs/Comment.hs
german1608: I believe we could do it. Would you mind if I do that in a follow-up PR?

german1608: OTOH: should we move only the parsing logic returning `Text` and keeping all data-types here? I say this because almost all defined data-types here are specific for dhall-docs comments validation

Gabriel439: Yeah, it's fine to do that in a follow-up PR.  I think this can be done in such a way that it does not depend on `dhall-docs`-specific types

dhall-docs/src/Dhall/Docs/Comment.hs
german1608: I had to also add `KindSignatures` to properly use the type restriction on `DhallComment

PR #1945, opened 2020-07-25
feat(parsing): centralize comments parsing extraction on Dhall.Parser.Token

german1608:
...as @Gabriel439 suggested [(comment)](https://github.com/dhall-lang/dhall-haskell/pull/1929#discussion_r459529799)

I didn't create any `newtype` over Comment's `Text` because I don't feel it's really necessary. If you believe it could be better by doing so let me know in your review :+1:

### DIFF
--- a/dhall/src/Dhall/Parser/Token.hs
+++ b/dhall/src/Dhall/Parser/Token.hs
@@ -4,6 +4,7 @@
 -- | Parse Dhall tokens. Even though we don't have a tokenizer per-se this
 ---  module is useful for keeping some small parsing utilities.
 module Dhall.Parser.Token (
+    getSourcePos,
     validCodepoint,
     whitespace,
     nonemptyWhitespace,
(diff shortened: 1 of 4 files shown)


dhall/src/Dhall/Parser/Token.hs
```diff
@@ -349,29 +352,32 @@ hexNumber = choice [ hexDigit, hexUpper, hexLower ]
...
 
     endOfLine
 
-    return ()
+    return ("--" <> commentText)
   where
     endOfLine =
         (   void (Text.Parser.Char.char '\n'  )
         <|> void (Text.Parser.Char.text "\r\n")
+        <|> Text.Megaparsec.eof
```
german1608, line 370: This line is added because of this: https://github.com/dhall-lang/dhall-lang/issues/1046, we can rollback it if necessary

PR #1974, opened 2020-08-04
feat(dhall-text): add output file command line argument

TristanCacqueray:
This change enables setting a file output for the dhall text command
so that it is consistent with the other command that generate an
output.

### DIFF
--- a/dhall/src/Dhall/Main.hs
+++ b/dhall/src/Dhall/Main.hs
@@ -161,7 +161,7 @@ data Mode
           }
     | Encode { file :: Input, json :: Bool }
     | Decode { file :: Input, json :: Bool, quiet :: Bool }
-    | Text { file :: Input, outputFile :: Output }
+    | Text { file :: Input, output :: Output }
     | DirectoryTree { file :: Input, path :: FilePath }
     | Schemas { file :: Input, outputMode :: OutputMode, schemas :: Text }
     | SyntaxTree { file :: Input, noted :: Bool }
@@ -902,7 +902,7 @@ command (Options {..}) = do
 
             case normalizedExpression of
                 Dhall.Core.TextLit (Dhall.Core.Chunks [] text) ->
-                    let write = case outputFile of
+                    let write = case output of
                           StandardOutput -> Data.Text.IO.putStr
                           OutputFile file_ -> Data.Text.IO.writeFile file_
                     in write text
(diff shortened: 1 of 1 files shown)


german1608:
Is this change relevant to `dhall-docs`? I believe it's just for the `dhall` executable. I'm asking because the title of the PR starts with `feat(dhall-docs)...`

TristanCacqueray:
Oops, I meant dhall-text, sorry :-)

dhall/src/Dhall/Main.hs
```diff
@@ -161,7 +161,7 @@ data Mode
           }
     | Encode { file :: Input, json :: Bool }
     | Decode { file :: Input, json :: Bool, quiet :: Bool }
-    | Text { file :: Input }
+    | Text { file :: Input, outputFile :: Maybe FilePath }
```
german1608, line 164: Instead of using `Maybe FilePath` you could use `Output` that is defined on `Dhall.Util`. There are some other flags that use that for this purpose.

Check here for instance:
https://github.com/dhall-lang/dhall-haskell/blob/760d98536d3c7e6a5cbf26d9c7a900758add91b8/dhall/src/Dhall/Main.hs#L339-L348

german1608: `Output` is a homomorphism for `Maybe FilePath`, but it's better to use that for consistency on this specific purpose though

TristanCacqueray: Indeed, thanks!

dhall/src/Dhall/Main.hs
```diff
@@ -161,7 +161,7 @@ data Mode
           }
     | Encode { file :: Input, json :: Bool }
     | Decode { file :: Input, json :: Bool, quiet :: Bool }
-    | Text { file :: Input }
+    | Text { file :: Input, outputFile :: Output }
```
german1608, line 164: It is not necessary a file!
```suggestion
    | Text { file :: Input, output :: Output }
```

TristanCacqueray: Indeed indeed.

TristanCacqueray:
Thanks for the review @german1608

german1608:
> Thanks for the review @german1608

@TristanCacqueray: you're welcome!

PR #1982, opened 2020-08-07
feat(dhall-docs): add jump to definition on lam parameters

german1608:
`Pi`-introduced variables are missing, but I'll introduce them at another time.

I also added a _small_ underline on variables to make this feature easier to discover.


german1608:
These are good demos:
* https://hydra.dhall-lang.org/build/69424/download/1/docs/List/fold.dhall.html
* https://hydra.dhall-lang.org/build/69424/download/1/docs/JSON/renderAs.dhall.html

Remember that links to another file on the prelude are broken (needs the `.dhall` extension)

PR #1990, opened 2020-08-12
Remove unneeded semigroups dependency

felixonmars:
It's not needed for GHC 8+ and we only support those.


Gabriel439:
@felixonmars: Thank you! :slightly_smiling_face:

PR #2001, opened 2020-08-16
Update static Nix build

Gabriel439:
This is a revamp of our use of
[fully static Haskell executables](https://github.com/NixOS/nixpkgs/issues/43795).
The main changes are:

* We can now use a newer revision of Nixpkgs

  The current revision we're using corresponds to Nixpkgs 20.03

* We can now more easily update Nixpkgs

  The logic in this repository more closely tracks how `static-haskell-nix`
  works, so whenever that repository updates Nixpkgs we can pick up the change
  here, too.

  Note that we cannot currently use Nixpkgs `master` due to
  [this issue](https://github.com/NixOS/nixpkgs/issues/85924), but once that is
  fixed when we can track Nixpkgs `master` more freely.

* We no longer need to keep track of separate Nixpkgs revisions for
  static vs. non-static builds

  The two builds can also share dependencies, too.  Only the top-level
  derivations differ now.

The reason that this doesn't use `static-haskell-nix` directly is because of
[`cabal2nix` evaluation failures](https://github.com/nh2/static-haskell-nix/issues/73),
so this change works by lifting just the minimum logic that we need from
the `static-haskell-nix` repository until we can finally use that
repository (or Nixpkgs directly when it's upstreamed).


release.nix
```diff
@@ -80,6 +58,10 @@ in
       tarball-dhall-nixpkgs
       tarball-dhall-openapi
       tarball-dhall-yaml
+
+      prelude-dhall-docs
+      test-dhall-docs
+#     kubernetes-dhall-docs
```
Gabriel439, line 64: Note that I plan on fixing this before merging.  I commented it out temporarily because it failed due to too many open file handles when building.

german1608: I had that error, and the solution was to rebuild when there are no executing Jobs on Hydra

PR #2005, opened 2020-08-20
feat(type-on-hover): Add type-on-hover on boolean and natural literals

german1608:
This is the first implementation of the type-on-hover feature. It is currently a draft and it is intended to be really small (excluding the golden file updates).

The reason for this PR to be that small is because this idea will be expanded on further PRs related to type-on-hover. So please let me know what you think and what can be improved before working on the next elements that will have this feature.

The tooltip library that I'm using is https://atomiks.github.io/tippyjs/, which is really lightweight.

I wanted to add jump-to-definition on imports before this, but I got blocked on that... didn't found the best way to do it. My problem is with implementation, I got an idea which I described [here (comment)](https://github.com/dhall-lang/dhall-haskell/issues/1994#issuecomment-675622903). Path manipulation was really messy and I felt that working on this feature first might be better because I already had a clearer idea about how to do this and I prefer to deliver as much work before ending the program as I can.

To end, I feel that the `dhall-docs` package structure is not the best... do you have any advice about organizing Haskell packages? I feel that the entire tool can fit roughly in a single file :laughing:

### DIFF
--- a/dhall-docs/src/Dhall/Docs/CodeRenderer.hs
+++ b/dhall-docs/src/Dhall/Docs/CodeRenderer.hs
@@ -52,6 +52,7 @@ import Dhall.Core
 import Dhall.Docs.Util
 import Dhall.Src                         (Src (..))
 import Lucid
+import Data.Text.Prettyprint.Doc (Pretty)
 import Text.Megaparsec.Pos               (SourcePos (..))
 
 import qualified Control.Monad.Trans.Writer.Strict     as Writer
@@ -66,6 +67,7 @@ import qualified Dhall.Core                            as Core
 import qualified Dhall.Map                             as Map
 import qualified Dhall.Parser
 import qualified Dhall.Pretty
+import qualified Dhall.TypeCheck as TypeCheck
 import qualified Lens.Family                           as Lens
 import qualified Text.Megaparsec.Pos                   as SourcePos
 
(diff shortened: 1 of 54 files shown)


german1608:
Please check out this whenever you have time and let me know what you think @Gabriel439 @sjakobi @Profpatsch

german1608:
https://hydra.dhall-lang.org/build/70598/download/1/docs/TypeOnHover.dhall.html Checkout this simple example

dhall-docs/src/Dhall/Docs/CodeRenderer.hs
```diff
@@ -239,6 +239,28 @@ fragments = Data.List.sortBy sorter . removeUnusedDecls . Writer.execWriter . in
...
+          where
+            _type = TypeFromSource $ case TypeCheck.typeOf (BoolLit b) of
+                Left _ -> fileAnIssue "typeOfStatic"
+                Right e -> e
+
+        Note src (NaturalLit n) -> Writer.tell [SourceCodeFragment src $ StaticExpression _type] >> return NoInfo
+          where
+            _type = TypeFromSource $ case TypeCheck.typeOf (NaturalLit n) of
+                Left _ -> fileAnIssue "typeOfStatic"
+                Right e -> e
```
Gabriel439, line 262: I recommend studying how `dhall-lsp-server` does this:

https://github.com/dhall-lang/dhall-haskell/blob/master/dhall-lsp-server/src/Dhall/LSP/Backend/Typing.hs

For example, one trick you can reuse from that code is to use the `subExpressions` `Traversal` to detect if a term has children or not when deciding whether or not to create a type tooltip

Gabriel439:
I'll close this for now since the work has stalled

PR #2010, opened 2020-08-25
Add preliminary dhall-docs manpage

vmchale:
This adds a `dhall-docs` manpage - more examples & feedback welcome :)

It's bundled the same way the `dhall` cli manpage is.


german1608:
@vmchale the [last](https://hydra.dhall-lang.org/build/70837/nixlog/2) error on the hydra pipeline is because `dhall-docs` read all contents from the `data` folder for static assets. That logic is on `Embedded.hs`.

A simple solution would be to move all assets to a dedicated folder, say, `data/assets`

vmchale:
Makes sense! Will do...

PR #2016, opened 2020-09-01
Version 1.34.0 → 1.35.0

Gabriel439:


dhall-bash/dhall-bash.cabal
```diff
@@ -31,7 +31,7 @@ Library
         base                      >= 4.11.0.0 && < 5   ,
         bytestring                               < 0.11,
         containers                               < 0.7 ,
-        dhall                     >= 1.34.0   && < 1.35,
+        dhall                     >= 1.34.0   && < 1.36,
```
german1608, line 34: Shouldn't the lower bound be `1.35.0`?

dhall-nixpkgs/dhall-nixpkgs.cabal
```diff
@@ -19,7 +19,7 @@ Executable dhall-to-nixpkgs
   Build-Depends:       base                 >= 4.11     && < 5
                      , aeson                >= 1.0.0.0  && < 1.5
                      , data-fix
-                     , dhall                >= 1.32.0   && < 1.35
+                     , dhall                >= 1.32.0   && < 1.36
```
german1608, line 22: Same comment here. Shouldn't the lower bound be `1.35.0`?

Gabriel439:
@german1608: The reason why not all packages change the lower bound is that some of them were unaffected by the breaking changes introduced in version 1.35.0, so as far as I can tell those packages still build against version 1.34.0.

german1608:
@Gabriel439 alright, makes sense!

Gabriel439:
Sorry for the delay in cutting this.  I just needed to fix https://github.com/dhall-lang/dhall-haskell/pull/2030 first since it's sufficiently disruptive

sjakobi (approved):

sjakobi (approved):

PR #2019, opened 2020-09-02
Don't require `apiVersion`/`type` for `PersistentVolumeClaim`

Gabriel439:
Related to https://github.com/dhall-lang/dhall-kubernetes/issues/136

In the long-run I plan to fix all of the `dhall-kubernetes` special-case
logic by undoing this logic:

https://github.com/dhall-lang/dhall-haskell/blob/9dca2b8129bfec47afa069d5d1008448dcd2f873/dhall-openapi/src/Dhall/Kubernetes/Convert.hs#L44

… which would make `apiVersion`/`kind` `Optional` across the board.
However, until then, this change makes them selectively `Optional`
only for `PersistentVolumeClaim`.

### DIFF
--- a/dhall-openapi/src/Dhall/Kubernetes/Convert.hs
+++ b/dhall-openapi/src/Dhall/Kubernetes/Convert.hs
@@ -57,9 +57,14 @@ requiredFields maybeName required
     --   in the Swagger spec but just in the docs
     notRequiredConstraints = Data.Map.fromList
       [ ( ModelName "io.k8s.api.core.v1.ObjectFieldSelector"
-        , Set.fromList [FieldName "apiVersion"])
+        , Set.fromList [ FieldName "apiVersion" ]
+        )
       , ( ModelName "io.k8s.apimachinery.pkg.apis.meta.v1.StatusDetails"
-        , Set.fromList [FieldName "kind"])
+        , Set.fromList [ FieldName "kind" ]
+        )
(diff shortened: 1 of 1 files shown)


PierreR:
@Gabriel439 I am wondering if the fix is complete. It looks like the generated default file for `io.k8s.api.core.v1.PersistentVolumeClaim.dhall`  is:

```
{ apiVersion = "v1"
, kind = "PersistentVolumeClaim"
, spec = None ./../types/io.k8s.api.core.v1.PersistentVolumeClaimSpec.dhall
, status = None ./../types/io.k8s.api.core.v1.PersistentVolumeClaimStatus.dhall
}
```
and it does not match the type:
```
{ metadata : ./io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.dhall
, apiVersion : Optional Text
, kind : Optional Text
, spec : Optional ./io.k8s.api.core.v1.PersistentVolumeClaimSpec.dhall
, status : Optional ./io.k8s.api.core.v1.PersistentVolumeClaimStatus.dhall
}
```

Gabriel439:
@PierreR: Yeah, it is a bug.  I'm currently on vacation this week, but I'll fix it afterwards

PierreR:
@Gabriel439 Thanks. It is not urgent at all.

Gabriel439:
@PierreR: The fix is up here: https://github.com/dhall-lang/dhall-haskell/pull/2053

PR #2027, opened 2020-09-03
make the HTTP Manager configurable

### LINKED ISSUE #2022: Swap out the HTTP Manager
Right now, dhall-haskell uses the `Manager` from http-client-tls to resolve remote imports by default. It can be disabled, but I don't see an easy way to provide a different `Manager` (maybe an `IO Manager` is better for laziness), especially without depending on `http-client-tls` (maybe a new flag is necessary). This might be desirable if
 - one only needs plain HTTP imports (probably rare),
 - one does already depend on another dependency providing a TLS-enabled `Manager` (http-client-openssl), or
 - one has security considerations (see e.g. [here](https://github.com/vincenthz/hs-tls/issues/117) or [here](https://github.com/haskell-crypto/cryptonite/issues/330)).

I am happy to create a PR if the direction is clear, though it might take some time.

amesgen:
Closes #2022.

This description is partly outdated (see discussion below).

---

A LOT of functions now take `IO Dhall.Import.Manager`, where `Manager = ()` on GHCJS or if `with-http` is disabled. The `Manager` is configurable via `EvaluateSettings`, and defaults to the previous default (default TLS `Manager` from http-client-tls with 30s timeout) when the `use-http-client-tls` flag is enabled, and the default plain `Manager` otherwise. This default is available in `Dhall.Import.defaultNewManager`.

Some remarks:
 1. The remote import tests fail (as expected) with `-fwith-http -f-use-http-client-tls` due to TLS failures. Do you know any plain HTTP servers without HTTPS redirections for this use case?
 2. When testing with `-f-with-http`, [this url](https://raw.githubusercontent.com/Nadrieril/dhall-rust/f7d8c64a9799f139ad65427c2518376adb9e2e2f/dhall/tests/import/success/unit/asLocation/EnvA.dhall) from `dhall-lang/tests/import/success/unit/asLocation/RemoteChainEnvA.dhall` does not seem to have been mocked.
 3. cryptonite is a rather big dependency, and it is only used for SHA256. It is a transitive dep of http-client-tls, so maybe now (not necessarily in this PR) it would be a good idea to use [cryptohash-sha256](https://hackage.haskell.org/package/cryptohash-sha256) instead.

In addition to bloat and security concerns, this PR enables to selectively disallow remote imports (e.g. in `Dhall.Format`).

Feel free to close this PR if the changes are too invasive or you see a different way to make the HTTP Manager configurable.


amesgen:
I don't want to DOS the CI, how do I mirror a CI run locally (with cabal)?

Gabriel439:
@amesgen: The most accurate way is using Nix:

```bash
$ nix build --file ./release.nix
```

… but the closest thing using plain Cabal is:

```bash
$ cabal configure -f-use-http-client-tls
$ cabal test
```

amesgen:
dhall-nixpkgs seems to be missing from the root cabal.project. Should I add it in this PR or is this out of scope?

Gabriel439:
@amesgen: Feel free to add them! 🙂

amesgen:
Rebased on master. I removed dhall-try again as it requires GHCJS (and thus, a naive `cabal build all` fails), and most people interested in GHCJS probably already use nix.

Gabriel439:
@amesgen: Instead of parametrizing each API function on a new `Manager` argument, would it be possible for the new `cabal configure` flag to bake in the insecure HTTP manager?  That would avoid having to change all of the type signatures.

amesgen:
If we did this, one could choose between these options:

 1. disable `with-http`: no HTTP imports at all
 2 . enable `with-http`, but disable  `use-http-client-tls`: only plain HTTP imports possible
 3. enable `with-http` and `use-http-client-tls`: HTTP and HTTPS imports possible

1. and 3. are already possible. In my [original comment](https://github.com/dhall-lang/dhall-haskell/issues/2022#issue-691983797), I mentioned (first bullet point) that it is probably rare that 2. is actually useful (how often does one need HTTP imports, but *not* HTTPS imports?).

The second and third bullet point are only possible with changed type signatures, and they are much more useful:
If one does already depend on e.g. http-client-openssl, one can disable the http-client-tls dependency of dhall. And if one has e.g. security reservations, but wants HTTPS imports, one needs to supply a different `Manager`, which is only possible with changed type signatures.

Gabriel439:
@amesgen: If the goal is to depend on `http-client-openssl` instead of `http-client-tls`, why not make a `cabal configure` flag for that instead?  Switching out a dependency seems like a reasonable thing to do at package configuration time.

amesgen:
That would have the disadvantage that possible future `Manager` backends can't be used/have to given special treatment in dhall (and disables the (rare) use case of plain HTTP-only `Manager`s).

Also, a configurable `Manager` has many other benefits apart from the http-client-tls vs. http-client-openssl thing (which was my original motivation indeed), e.g. one can configure proxy settings, connection timeouts (hardcoded right now), retry conditions, and one can share a `Manager` between different invocations of Dhall functions. 

I understand that these type signature changes are not ideal. If you feel that the added flexibility of this PR is not worth the cost, I can create a new one with your suggestion to create a flag to switch between `http-client-openssl` and `http-client-tls` (which I also like, even though it is somewhat ad-hoc).

Gabriel439:
@amesgen: How about preserving the original type signatures for the default functions, but add variations on them that accept a `Manager` as an argument?

amesgen:
Good idea, what suffix should we use? `'` , or `withMgr`, or `withManager`,...?

Gabriel439:
@amesgen: `…WithManager` works fine for me

amesgen:
I reverted the changes to all subprojects except `dhall`, as `dhall-json` etc. are probably not often used as libraries. Also, I did not create `...withManager` variants for the functions in `dhall` which only serve as the implementation of specific subcommands for the same reason.

It might be a good idea to add a test that the `Manager` passed to `EvaluateSettings` is actually used in the import resolution, as it is imaginable that in a refactor, the `Manager` is not passed on due to the usage of a non-`withManager`-function.

Gabriel439 (approved):
Looks great!  I'm just waiting on CI

Gabriel439:
@amesgen: Thank you for doing this! 🙂

PR #2029, opened 2020-09-05
feat: separate lib/exe in dhall-openapi.cabal 

PierreR:


Gabriel439:
@PierreR: Thank you! 🙂

Issue #2033, opened 2020-09-10
Command line:  file path completion


PierreR:
This might looks like a detail but it comes to a surprise that `dhall` command supports completions (which is 🎉) but on the other hand file name are not completed (for instance dhall type --file TABTAB).

The later behaviour is really uncommon on a linux terminal.

Is this something that could be fixed easily ?

Gabriel439:
@PierreR: Yeah, I'm fairly confident that this can be fixed.  The main thing I need to know is how you obtained/install `dhall`, since that can affect how the completions are provided

ggilmore:
I install dhall via https://github.com/aaaaninja/asdf-dhall

Gabriel439:
Fix is up here: https://github.com/dhall-lang/dhall-haskell/pull/2034

PR #2034, opened 2020-09-11
Fix installed completions

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/2033


sjakobi (approved):
Neat! I wasn't aware of these `action`s!

PR #2054, opened 2020-09-22
Pack up manpages alongside binaries (for linux + mac)

vmchale:
`dhall-docs` and `dhall` themselves have manpages, so these could be packed up in the binary distributions. This adds some travis machinery to do so.


Gabriel439:
@vmchale: Thank you for doing this! 🙂

Issue #2066, opened 2020-10-11
"Multiple declarations" errors in Haskell using makeHaskellTypes with MultipleConstructors


joefiorini:
I'm working on a project where I have unions that reference other types in the same file. Here is an example of some Dhall types that will exhibit the problem I'm having:

```dhall
let Bar = { baz : Integer }
let Foo = { foo : Integer, bar : Bar }
let Qux = < Foo : Foo | Bar : Bar >
in { Foo, Bar, Qux }
```
When I try to convert these to Haskell types using `Dhall.TH` I get a bunch of "Multiple declarations" errors from Haskell because (I'm assuming) it's replacing `Qux`s reference to `Bar` with the actual type of `Bar` (probably some sort of normalization logic?) Here is the Haskell code I'm using to bring them in:

```haskell
#!/usr/bin/env nix-shell
#! nix-shell -I nixpkgs="channel:nixpkgs-unstable" --keep LANG --keep LC_ALL --keep NIX_CONF_DIR
#! nix-shell --pure -i "runghc" --packages ghcid "haskellPackages.ghcWithPackages (pkgs: [ pkgs.turtle pkgs.dhall ])" -p git -p glibcLocales

{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE QuasiQuotes #-}
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE StandaloneDeriving #-}
{-# LANGUAGE DerivingStrategies #-}
{-# LANGUAGE DeriveAnyClass #-}
{-# LANGUAGE DeriveGeneric #-}

import Dhall
import Dhall.TH
import Data.Text (Text)
import qualified Data.Text as T
--import DhallCode


makeHaskellTypes
  [ SingleConstructor "Bar" "Bar" "(./test-types.dhall).Bar"
  , SingleConstructor "Foo" "Foo" "(./test-types.dhall).Foo"
  , MultipleConstructors "Qux" "(./test-types.dhall).Qux"
  ]

foo' = Foo {
  foo = 1
, bar = Bar {
    baz = 2
  }
}

deriving instance Show Foo
deriving instance Show Bar

main = do
  putStrLn $ show foo'
```
When I run it I get the following errors from Haskell:

```
test-interp.hs:21:1: error:
    Multiple declarations of ‘bar’
    Declared at: test-interp.hs:21:1
                 test-interp.hs:21:1
   |
21 | makeHaskellTypes
[...]

PR #2070, opened 2020-10-13
Improve template Haskell support for record constructors

Gabriel439:
Now the code generation will detect if a record constructor is
identical in its entirety to another type declaration and replace
the field with the reference to the other type.

Fixes https://github.com/dhall-lang/dhall-haskell/issues/2066


sjakobi (approved):
:+1:

PR #2085, opened 2020-11-01
read 'explain' and 'ascii' from config file

mheiber:
related: https://github.com/dhall-lang/dhall-lang/issues/1086

search for dhall-config.dhall in:
- cwd if reading from stdout
- current file if targeting on a single file
- first file if using --transitive

There are only two configuration options, corresponding to their respective command line flags:

```dhall
{ ascii: Maybe Bool, explain: Maybe Bool }
```

**TESTING**

See https://github.com/dhall-lang/dhall-haskell/pull/2085#issuecomment-723504279


dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -337,10 +337,10 @@ documentFormattingHandler request = do
     _ -> throwE (Warning, "Failed to format dhall code; parse error.")
 
   ServerConfig {..} <- getServerConfig
-  let charSet | asciiOnly = ASCII
+  let characterSet | asciiOnly = ASCII
               | otherwise = Unicode
```
Gabriel439, line 341:
```suggestion
  let characterSet
          | asciiOnly = ASCII
          | otherwise = Unicode
```

Same comment for the other two occurrences of this pattern

dhall/src/Dhall/Main.hs
```diff
@@ -210,6 +213,14 @@ parseOptions =
             <>  Options.Applicative.help description
             )
 
+    flag name description =
+        Options.Applicative.flag
+            Nothing
```
Gabriel439, line 218:
```suggestion
            (Just True)
```

dhall/src/Dhall/Main.hs
```diff
@@ -540,19 +551,51 @@ noHeaders
...
+
+                configToIsAscii Nothing = Just False
+                configToIsAscii (Just config) = ascii config
+
+                decode (Just s) = fmap Just $ Dhall.input Dhall.auto $ Data.Text.pack s
+                decode Nothing = return Nothing
+
+                findConfig startDir =
+                    System.Directory.makeAbsolute startDir >>= \case
+                            "/" -> return Nothing
```
Gabriel439, line 588: It seems like we should technically allow the configuration file to reside in `/`, just for consistency

dhall/src/Dhall/Main.hs
```diff
@@ -540,19 +551,51 @@ noHeaders
 noHeaders i =
     i
 
+data Config = Config
+    { ascii  :: Maybe Bool } deriving Dhall.Generic
```
Gabriel439, line 555: We should probably also add a configuration option for `explain`, since this configuration is likely to remain Haskell-specific

dhall/src/Dhall/Main.hs
```diff
@@ -540,19 +551,51 @@ noHeaders
...
+                configPath <- findConfig $ rootDirectory input
+                config <- decode configPath
+                let isAscii = fromMaybe False $ configToIsAscii config in
+                    return . toCharSet $ isAscii
+            where
+                toCharSet True = ASCII
+                toCharSet False = Unicode
+
+                configToIsAscii Nothing = Just False
+                configToIsAscii (Just config) = ascii config
```
Gabriel439, line 581: It seems like the `configToIsAscii` can be omitted:

```suggestion
                let isAscii = fromMaybe False config
                return . toCharSet $ isAscii
            where
                toCharSet True = ASCII
                toCharSet False = Unicode
```

dhall/src/Dhall/Main.hs
```diff
@@ -540,19 +551,51 @@ noHeaders
...
+                            "/" -> return Nothing
+                            dir ->
+                                let f = System.FilePath.joinPath [dir, "dhall-config.dhall"] in
+                                    System.Directory.doesFileExist f >>= \case
+                                        True -> return $ Just f
+                                        False -> findConfig $ System.FilePath.takeDirectory dir
+
+    let getCharacterSetFromPossiblyTransitive = \case
+            NonTransitiveStandardInput -> getCharacterSet StandardInput
+            (PossiblyTransitiveInputFile filePath _) -> getCharacterSet (InputFile filePath)
```
Gabriel439, line 597:
```suggestion
            PossiblyTransitiveInputFile filePath _ -> getCharacterSet (InputFile filePath)
```

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
mheiber: is there an autoformatter?

dhall/src/Dhall/Main.hs
mheiber: who, nice catch

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
Gabriel439: We do not currently use an autoformatter

dhall/src/Dhall/Main.hs
```diff
@@ -210,6 +213,14 @@ parseOptions =
             <>  Options.Applicative.help description
             )
 
+    flag name description =
+        Options.Applicative.flag
+            (Just True)
```
mheiber, line 218: @Gabriel439 I'm pretty sure we want this one to default to `Just False`, since the defaults for both `--ascii` and `--explain` are `False`

dhall/src/Dhall/Main.hs
```diff
@@ -107,19 +109,28 @@ import qualified System.AtomicWrite.Writer.LazyText        as AtomicWrite.LazyTe
...
-    { mode    :: Mode
-    , explain :: Bool
-    , plain   :: Bool
-    , ascii   :: Bool
-    , censor  :: Censor
+    { mode             :: Mode
+    , optionsExplain   :: Maybe Bool
+    , plain            :: Bool
+    , optionsAscii     :: Maybe Bool
+    , censor           :: Censor
```
mheiber, line 123: will fix

dhall/src/Dhall/Main.hs
mheiber: I added support for `explain`. Still can't run the tests until I figure out how to get Nix working on Catalina.

mheiber:
Regarding testing, how about I set up fixture directories with config files and invoking `main`–does that sound good?

dhall/src/Dhall/Main.hs
```diff
@@ -113,13 +116,21 @@ import qualified Text.Pretty.Simple
 
 -- | Top-level program options
 data Options = Options
-    { mode    :: Mode
-    , explain :: Bool
-    , plain   :: Bool
-    , ascii   :: Bool
-    , censor  :: Censor
+    { mode             :: Mode
+    , optionsExplain   :: Maybe Bool
```
Gabriel439, line 120: If you enable the `DuplicateRecordFields` extension then you can keep the original field names

dhall/src/Dhall/Main.hs
```diff
@@ -210,6 +221,14 @@ parseOptions =
             <>  Options.Applicative.help description
             )
 
+    flag name description =
+        Options.Applicative.flag
+            (Just False)
+            Nothing
```
Gabriel439, line 227: I think the issue we're running into here is that we don't have a good way via the command line to specify the following three options:

* Use the default (possibly obtained from the configuration file)
* Override the default to prefer `False`
* Override the default to prefer `True`

… because if we use `flag` then the user can only indicate at most two of those options.

One way handle this is to add a new flag for the opposite of `--explain` and `--ascii`.  I'm thinking something like:

* `--unicode` (the opposite of `--ascii`)
* `--concise` (the opposite of `--explain`)

… and if the user doesn't specify either alternative then it uses whatever is in the default or the configuration file.

You can implement that behavior in `optparse-applicative` by doing something like this (for the case of ASCII vs. Unicode):

```haskell
    Options.Applicative.flag' (Just ASCII) …
<|> Options.Applicative.flag' (Just Unicode) …
<|> pure Nothing
```

Gabriel439 (commented):
Looking great so far!

dhall/src/Dhall/Main.hs
mheiber: That's a nice catch for what would have been a UX bug!

In your (very helpful) code example, you used ASCII/Unicode instead of True/False. Currently, the options `ascii` constructor is `Bool`. I think it may actually be easier to keep it close to as-is (Maybe Bool) because then we can use the same `resolve` logic for both ascii and explain:

```hs
                -- resolve :: Maybe Bool -> (t -> Maybe Bool) -> Maybe t -> Bool
                resolve (Just option) _configKey _config = option
                resolve Nothing _configKey Nothing = False
                resolve Nothing configKey (Just config) = fromMaybe False $ configKey config
```

mheiber (commented):
I updated the flags and used duplicate record fields.
Thanks again for the code suggestion for parsing command flags, I think it would have taken me a long time to figure that out.

Do you have any suggestions for how to test?

dhall/src/Dhall/Main.hs
Gabriel439: @vmchale : Yeah, it's fine if you use `True`/`False` instead of `ASCII`/`Unicode`

dhall/src/Dhall/Main.hs
```diff
@@ -113,13 +117,21 @@ import qualified Text.Pretty.Simple
...
-    { mode    :: Mode
-    , explain :: Bool
-    , plain   :: Bool
-    , ascii   :: Bool
-    , censor  :: Censor
+    { mode      :: Mode
+    , explain   :: Maybe Bool
+    , plain     :: Bool
+    , ascii     :: Maybe Bool
+    , censor    :: Censor
```
Gabriel439, line 124:
```suggestion
    { mode    :: Mode
    , explain :: Maybe Bool
    , plain   :: Bool
    , ascii   :: Maybe Bool
    , censor  :: Censor
```

Gabriel439 (commented):
I think the main way to test this is to export the logic for resolving the configuration file (i.e. `resolvedConfig`) into a top-level function in `Dhall.Main` that is parametrized on the starting directory:

```haskell
resolveConfig
    :: FilePath
    -- ^ Starting directory
    -> IO Config
```

… and testing it against a test config file within a test directory underneath `./tests/`

Gabriel439:
@mheiber: Before continuing, you might want to review @PierreR's comment about the idea of using `direnv` to do the same:

https://github.com/dhall-lang/dhall-lang/issues/1086#issuecomment-727617496

My understanding is that he's suggesting that a project could use `direnv` to locally alias `dhall` to `dhall --ascii` within a project

dhall/tests/Dhall/Test/Dhall.hs
```diff
@@ -489,3 +497,83 @@ shouldShowCorrectErrorForInvalidDecoderInUnion =
...
+            ascii = ascii
+            , censor = Dhall.Util.Censor
+            , explain = explain
+            , mode = Dhall.Main.Format{
+                possiblyTransitiveInput = Dhall.Util.NonTransitiveStandardInput
+                , outputMode = Dhall.Util.Write
+            }
+            , plain = False
+        }
+
```
mheiber, line 572: I think it would be more maintainable here to use Dhall's built-in ability to turn Haskell values into Dhall values, but I couldn't figure out how to do it.

dhall/src/Dhall/Main.hs
```diff
@@ -15,16 +17,20 @@ module Dhall.Main
     , ResolveMode(..)
     , parseOptions
     , parserInfoOptions
+    , findConfig
+    , resolveSettings
 
       -- * Execution
     , Dhall.Main.command
     , main
+    , CharacterSet(..)
```
mheiber, line 26: I exported this so I could refer to it in the tests – a bit awkward, since CharacterSet is defined in Internal.hs

dhall/tests/Dhall/Test/Dhall.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase #-}
```
mheiber, line 1: is Dhall.hs the right file for these tests?

Gabriel439:
@mheiber: I'd still like to consider @PierreR's suggestion to use `direnv`, because then we wouldn't need any changes to `dhall`.  A user can use it selectively alias `dhall` to `dhall --ascii` within a project.  See: https://github.com/dhall-lang/dhall-lang/issues/1086#issuecomment-727617496

mheiber:
> @mheiber: I'd still like to consider @PierreR's suggestion to use `direnv`, because then we wouldn't need any changes to `dhall`. A user can use it selectively alias `dhall` to `dhall --ascii` within a project. See: [dhall-lang/dhall-lang#1086 (comment)](https://github.com/dhall-lang/dhall-lang/issues/1086#issuecomment-727617496)

thanks, I missed [the comment](https://github.com/dhall-lang/dhall-lang/issues/1086#issuecomment-727617496)

mheiber:
CI is failing because of "missing haddocks" - is this because I added more top-level stuff to the `Main` module?

Gabriel439:
I'll close this since I believe it is mostly superseded by https://github.com/dhall-lang/dhall-haskell/pull/2108 which now detects the character set from the input file

PR #2097, opened 2020-11-15
Improve error message for duplicate projection label

Gabriel439:
… by correctly adhering to the standard, which treats this as a type
error instead of a parse error.

Fixes https://github.com/dhall-lang/dhall-haskell/issues/2094


sjakobi (approved):
:+1: 

I think `Dhall.Parser.Combinators.noDuplicates` could be removed now.

Issue #2096, opened 2020-11-13
Haddocks for RecordField are confusing


ocharles:
The Haddocks for `RecordField` talk about comments `A`, `B` and `C`, but these aren't visible. On inspection of the source, I think the problem is they need to be escaped.

Compare https://hackage.haskell.org/package/dhall-1.36.0/docs/src/Dhall.Syntax.html#RecordField against https://hackage.haskell.org/package/dhall-1.36.0/docs/Dhall-Core.html#t:RecordField

german1608:
This issue is not specific for `RecordField`, but for `Binding` and `FieldSelection` as well:

* https://hackage.haskell.org/package/dhall-1.36.0/docs/src/Dhall.Syntax.html#Binding
    https://hackage.haskell.org/package/dhall-1.36.0/docs/Dhall-Core.html#t:Binding
* https://hackage.haskell.org/package/dhall-1.36.0/docs/src/Dhall.Syntax.html#FieldSelection
    https://hackage.haskell.org/package/dhall-1.36.0/docs/Dhall-Core.html#t:FieldSelection

The reason is that (could be great if someone else confirms this too) Haskell's block comments can be nested, and the syntax for Dhall's block comments it's the same, so these in-between comments for documentations are erased on haddock docs.  I tried to escape those comments in the past, but no luck from my side :(

Gabriel439:
I know how to fix this.  The trick is that if you want haddocks to render block comments in code blocks then the haddock comment itself has to use `-- ` for the comment

Gabriel439:
Fix is up here: https://github.com/dhall-lang/dhall-haskell/pull/2098

PR #2098, opened 2020-11-15
Fix haddocks for `RecordField`/`Binding`/`FieldSelection`

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/2096

`haddock` doesn't correctly handle block comments in code blocks if
the haddock comment itself is a block comment.  However, if the haddock
comment uses line comments then it works just fine.


Gabriel439:
@sjakobi: Thank you for reporting that! 🙂

PR #2100, opened 2020-11-21
Fix example code in the tutorial

jeremyschlatter:


jeremyschlatter:
The changes here allowed me to compile and run the example Haskell code. The code does not seem to work as described, though. In particular this claim

> Now we can omit the definition of Result in our example.dhall file. While this will work perfectly...

seems to be false. When I change example.dhall to be

    -- example.dhall
    
    Result.Failure +1

and then parse it with

    Dhall.input (Dhall.auto :: Dhall.Decoder Result) ("let Result = " <> Data.Text.pack resultTypeString <> " in ./example.dhall")

I get this failure:

    main: 
    ↳ ./example.dhall
    
    Error: Unbound variable: Result

Gabriel439 (approved):
@jeremyschlatter: Yeah, that part of the documentation is incorrect.  Bound variables are not forwarded to transitive imports.  Would you like to also delete that part as part of this pull request?

PR #2108, opened 2020-12-03
dhall-format: Detect character set from input

basile-henry:
This PR implements the suggestion from https://github.com/dhall-lang/dhall-lang/issues/1086

`dhall format` can now detect which character set is used in the input instead of defaulting to Unicode.

- Only Dhall syntax is considered for the detection (strings and comments don't affect the decision)
- If no Unicode is used in expression default to ASCII
- Add CLI switch to specify Unicode explicitly


dhall/src/Dhall/Format.hs
```diff
@@ -117,3 +131,100 @@ format (Format { input = input0, ..}) = go input0
                         let modified = "formatted"
 
                         Control.Exception.throwIO CheckFailed{..}
+
+--
```
basile-henry, line 135: I will add a better comment :sweat_smile:

dhall/src/Dhall/Format.hs
```diff
@@ -117,3 +131,100 @@ format (Format { input = input0, ..}) = go input0
...
+                      "→" `Text.isInfixOf` srcText -> Unicode
+                Prefer _ _ _
+                    | "⫽" `Text.isInfixOf` srcText -> Unicode
+                Combine _ _ _
+                    | "∧" `Text.isInfixOf` srcText -> Unicode
+                CombineTypes _ _
+                    | "⩓" `Text.isInfixOf` srcText -> Unicode
+                _ -> recurse expr
+
+        -- Recurse down the AST
```
basile-henry, line 160: Do we have a better way to recurse on the rest of the AST?

dhall/src/Dhall/Format.hs
```diff
@@ -117,3 +131,100 @@ format (Format { input = input0, ..}) = go input0
...
+                Pi _ _ _
+                    | "∀" `Text.isInfixOf` srcText ||
+                      "→" `Text.isInfixOf` srcText -> Unicode
+                Prefer _ _ _
+                    | "⫽" `Text.isInfixOf` srcText -> Unicode
+                Combine _ _ _
+                    | "∧" `Text.isInfixOf` srcText -> Unicode
+                CombineTypes _ _
+                    | "⩓" `Text.isInfixOf` srcText -> Unicode
+                _ -> recurse expr
```
basile-henry, line 158: I think we might still get some false positives, for example if one of these Unicode characters appears in a string or comment within these specific source spans.

I'm not sure how to improve this though. The only other approach I have in mind involve changing the `Parser` type, which seems a bit heavy handed :sweat_smile:

basile-henry: Well I guess I could add an extra field for `CharacterSet` to `Lam`, `Pi`, `Prefer`, `Combine`, and `CombineTypes`. That might actually be the cleanest option.

dhall/src/Dhall/Format.hs
Gabriel439: We do!  You can use the `subExpressions` traversal in conjunction with one of the utilities from `Control.Lens.Plated`.  For example, I think this would work:

```haskell
detectCharacterSet :: Expr s a -> CharacterSet
detectCharacterSet = foldOf (cosmosOf subExpressions . to characterSet)
  where
    characterSet :: Expr s a -> CharacterSet
    characterSet (Note Src{ srcText } expression) = case expression of
        Lam{} | … -> Unicode
        Pi{} | … -> Unicode
        …
        CombineTypes{} | …-> Unicode
        _ -> mempty
```

That way you don't need to implement the recursion.  The `characterSet` function only needs to analyze things one level deep.

Note that this might require vendoring the implementation of `cosmosOf` in the `Dhall.Optics` module to avoid the `lens` dependency

dhall/src/Dhall/Format.hs
Gabriel439: Yeah, I think the reliable solution is to extend those constructors to store which character set was parsed

dhall/src/Dhall/Format.hs
basile-henry: Woah that looks like magic 😍

I'll give it a go! 👍

basile-henry: It works very well, thank you! I need to learn these advanced lenses one day :sweat_smile:

dhall/src/Dhall/Format.hs
basile-henry: I have now implemented to reliable solution. It makes the PR quite a bit bigger, but at least we won't get any weird behaviour around formatting.

I hope this won't affect parsing performance. Do we have any benchmarks in place?

dhall-openapi/openapi-to-dhall/Main.hs
```diff
@@ -66,7 +66,7 @@ writeDhall path expr = do
   putStrLn $ "Writing file '" <> path <> "'"
   Text.writeFile path $ pretty expr <> "\n"
 
-  let characterSet = Dhall.Pretty.ASCII
+  let chosenCharacterSet = Just Dhall.Pretty.ASCII
```
basile-henry, line 69: I'm not entirely sure why `openapi-to-dhall` was defaulting to `ASCII` (and not `Unicode` like `dhall format`). Now that we have the option to infer it from the input, should we do that?

dhall/src/Dhall/Pretty/Internal.hs-boot
```diff
@@ -1,13 +1,27 @@
...
+data CharacterSet = ASCII | Unicode
+
+instance Eq CharacterSet
+instance Ord CharacterSet
+instance Show CharacterSet
+instance Data CharacterSet
+instance Lift CharacterSet
+instance NFData CharacterSet
+instance Semigroup CharacterSet
+instance Monoid CharacterSet
```
basile-henry, line 23: I am not 100% sure how `hs-boot` files work. I had write the constructors and the instances otherwise they wouldn't propagate properly. Is there another way?

dhall/src/Dhall/Format.hs
basile-henry: I found some benchmarks! :smile:

basile-henry: So, this more complete solution kind of makes equality for `Expr` a bit more annoying. I am currently not propagating the character set into the evaluation types, which means I lose that information when not dealing with actual source code. I thought that would be a fine approach except it breaks all the tests that rely on equality not caring about character set (which is many of them :sweat_smile: )

Would a hacky `Eq` instance for `CharacterSet` where `_ == _ = True` be acceptable? :stuck_out_tongue_closed_eyes:

basile-henry: Or maybe I should take a similar approach to what we do for the imports. I could have an extra type level argument for `Expr` and then `Void` out that field when the `Expr` doesn't represent something that was directly inputted by text.

Gabriel439: Could you give an example of a test that relies on equality on `Expr`?

basile-henry: There's quite a few of them (around 200): https://hydra.dhall-lang.org/build/79114/nixlog/2

I gave the extra parameter to `Expr` a go. It unfortunately involves a lot of code changes. And I don't think it can work. Maybe of the traversal code would require the extra parameter to be a `Monoid` and `Void` is only a `Semigroup`.

One other alternative to fix this is to actually properly propagate `CharacterSet` through the evaluation (in types like `Val`). I will try this approach.

basile-henry: > One other alternative to fix this is to actually properly propagate CharacterSet through the evaluation (in types like Val). I will try this approach.

This does fix many of the failing tests but not all. Some of them seem unfixable with this approach. For example binary serialisation loses that information (as it should). And typechecking now fails to unify if it inferred different `CharacterSet`s, which is obviously very silly :sweat_smile:

At the moment I think either I need to revert to the `Src` based solution which wasn't very robust since it could falsely detect unicode when it wasn't actually used in syntax (badly placed comments), or I need to add that hacky `Eq` instance. I could potentially but that instance on a newtype to make it more explicit:

```haskell
newtype AlwaysEq a = AlwaysEq a

instance Eq (AlwaysEq a) where
    _ == _ = True
```

And then use `AlwaysEq CharacterSet` in the `Expr` AST. It's still a bit ugly but at least there shouldn't be too many surprises.

Gabriel439: Actually, what I think you can do is to make the field a `Maybe` value, and parsing provides a `Just` value and `denote` turns it into `Nothing`.  This would be consistent with how we handle other source spans in the code.

basile-henry: Right! That actually makes more sense. I'm not sure why I didn't think that Src would also be an issue for Eq.

I will do that thanks! 👍

dhall/src/Dhall/Format.hs
basile-henry: I think I missed the importance of the guards in your suggestion. My current version probably doesn't handle nested Lambdas and only uses the character set from the outer one.

I will fix that and I will also try to add some tests to cover the new behaviour of dhall format.

dhall/src/Dhall/Format.hs
basile-henry: I implemented your `Maybe CharacterSet` solution, much nicer than what I was attempting! Thanks for the suggestion :smile:

dhall-openapi/openapi-to-dhall/Main.hs
Gabriel439: @basile-henry: The only reason it was defaulting to ASCII is because it was factored out from `dhall-kubernetes`, which was using ASCII.  However, I think it's fine to change the behavior to infer it from the input

dhall/src/Dhall/Pretty/Internal.hs-boot
Gabriel439: What you did is fine, but to answer your question: another solution is to factor out the `CharacterSet` type into its own module (e.g. `Dhall.CharacterSet`) to break the cyclic dependency

dhall/src/Dhall/Main.hs
```diff
@@ -542,9 +545,14 @@ noHeaders i =
...
 command (Options {..}) = do
-    let characterSet = case ascii of
-            True  -> ASCII
-            False -> Unicode
+    let chosenCharacterSet =
+            case (ascii, unicode) of
+                -- If both flags are set, default to ASCII
+                (True, _) -> Just ASCII
+                (_, True) -> Just Unicode
+                _         -> Nothing
```
Gabriel439, line 553: It's actually possible to ensure that the flags are mutually exclusive.  The trick is to replace this:

```haskell
    <*> switch "ascii" "Format code using only ASCII syntax"
    <*> switch "unicode" "Format code using only Unicode syntax" 
````

… with this:

```haskell
  <*> parseCharacterSet
  …
  where
    parseCharacterSet =
            flag' (Just Unicode) (long "unicode" <> help "Format code using only Unicode syntax")
        <|> flag' (Just ASCII) (long "ascii" <> help "Format code using only ASCII syntax")
        <|> pure Nothing
```

Gabriel439 (approved):
Looks great!  Just a few small suggestions

dhall/src/Dhall/Main.hs
basile-henry: That's a good idea! :smile: 

Implementing this change made me realise that we are still using the old `--ascii` switch for other things than just `dhall format`. We seems to be using it for pretty much every command. Should we also infer from the input in these other cases? Or should we limit this to `dhall format`? I should have thought about this earlier... :sweat_smile:

Gabriel439: The behavior should probably be the same for all of the commands, not just `dhall format`

basile-henry: I believe I've now made it infer the CharacterSet in all commands where we could already specify which one we wanted. I think there are some remaining commands where it wasn't configurable before like `dhall diff` (where it always uses Unicode). Should I look into making them configurable as well?

Gabriel439: @basile-henry: You can if you would like to

Gabriel439:
@basile-henry: Note that I'd like to hold off on merging this until we first cut a small 1.37.1 release to push out the fix in https://github.com/dhall-lang/dhall-haskell/pull/2112

basile-henry:
@Gabriel439 No worries, it can wait. I was also wondering if we should get feedback/approval from others in the https://github.com/dhall-lang/dhall-lang/issues/1086 thread.

Someone in that thread wanted to be able to configure the language server to use `Unicode` specifically. Is this something I can do in this PR/codebase or do I need to synchronize with the VS code LSP plugin somewhere else as well?

Unrelated: About cutting fixes for previous releases, do you know that Mergify can backports PR automatically? https://docs.mergify.io/commands.html#backport :smile: I think we just need to keep around a branch for the latest release (or however many releases we want to keep supporting) and then we can easily backport a PR to that branch.

dhall-lsp-server/src/Dhall/LSP/State.hs
```diff
@@ -40,21 +41,19 @@ data Severity = Error
               -- ^ Log message, not displayed by default.
 
 data ServerConfig = ServerConfig
-  { asciiOnly :: Bool
-  -- ^ Use ASCII symbols rather than fancy unicode when formatting and linting
-  -- code.
+  { chosenCharacterSet :: Maybe CharacterSet
   } deriving Show
```
basile-henry, line 45: @PanAeon Do you know if I need to modify something else to make this new config format work? Maybe somewhere in https://github.com/PanAeon/vscode-dhall-lsp-server ? I haven't seen anything that sets the LSP config there though.

timbertson:
> If no Unicode is used in expression default to ASCII

Is that accurate? I am extremely vocally in favour of ASCII mode so I'd love it. But I haven't seen discussion of it, and dhall has historically defaulted to unicode, so I don't want to get my hopes up ;)

If it is accurate, are there eventual plans to drop unicode, or would it be supported as opt-in forever?

basile-henry:
> Is that accurate?

Yes! The main way people use the Unicode syntax is by letting the formatter change their operators automatically (I don't think anyone types in Unicode, but maybe I'm wrong).
We don't want to completely lose that, so the logic used for detection is: if there is at least one Unicode operator used, then use Unicode.
However that does mean that the default has to change, otherwise ASCII would never be detected.

> dhall has historically defaulted to unicode

Yes, and that's one of the reasons I've been so slow at merging this PR. Ideally I'd like more feedback on this new way of handling character sets. I know some people have become attached to the Unicode character set and now they have one extra step to use it (either type a Unicode operator once or use `dhall format --unicode` once).

> are there eventual plans to drop unicode, or would it be supported as opt-in forever

As far as I know there are no plans to drop Unicode.

Gabriel439:
@basile-henry: I'm fine merging this now.  There's no need to hold off merging any longer, in my opinion

@timbertson: This comment explains why the default is Unicode if the file has a mixture of ASCII and unicode: https://github.com/dhall-lang/dhall-lang/issues/1086#issuecomment-738121239.  If the file is all ASCII then it will default to ASCII

dhall-lsp-server/src/Dhall/LSP/State.hs
basile-henry: It looks like it works just fine without modifying the VS code extension, and this config can still be set from `settings.json` manually. I might try to make this new option accessible from the extension's UI, but it's not necessary for now.

basile-henry:
> I'm fine merging this now. There's no need to hold off merging any longer, in my opinion

Alright, let's do it! We can always revert/fixup as needed in future PRs :+1:

Gabriel439:
@basile-henry: Thank you for contributing this! 🙂

Issue #2105, opened 2020-11-30
possible `with` performance regression in 1.36.0 release


ggilmore:
Related https://github.com/dhall-lang/dhall-haskell/issues/1960

When I run `dhall` against a file with a lot of chained `with` statements, dhall `1.36.0` seems to choke while dhall `1.35.0` executes the file quickly. 

I've created a reproduction repository here: https://github.com/ggilmore/dhall-performance-regression-example-with

Run `dhall 1.35.0` with `dhall --file=src/fast-1.35.0/fast.dhall` - see that the command executes quickly

Run `dhall 1.36.0` with `dhall --file=src/slow-1.36.0/slow.dhall` - see that the command never terminates

Gabriel439:
Yeah, I'm pretty sure I know what went wrong here.  The regression is likely due to https://github.com/dhall-lang/dhall-haskell/pull/2055 and I believe I can fix it by using an optimization similar to the one we used back when `with` was syntactic sugar.

Gabriel439:
Fix is up here: https://github.com/dhall-lang/dhall-haskell/pull/2112

PR #2112, opened 2020-12-08
Fix performance regression for `with` expressions

Gabriel439:
Fixes: https://github.com/dhall-lang/dhall-haskell/issues/2105


ggilmore:
Thank you! Would it be possible to roll this into a patch release, or should we wait for `1.38.0`?

Gabriel439:
@ggilmore: I can cut an early 1.37.1 release to get this fix out

sjakobi (approved):

ggilmore:
Awesome thank you! Also I don't know if this is the appropriate place to ask this, but is there a test that can be added to the suite that would catch this in the future? It might be tricky if the only indicator of an issue is the length of the execution time though.

Gabriel439:
@ggilmore: Probably not.  Anything that would causes tests to fail would be problematic when run on different machines (e.g. as part of Nixpkgs CI, for example)

PR #2118, opened 2020-12-31
Fix links to vscode-language-dhall repository

basile-henry:
While the HTTP links properly redirect to the new repo, the git clone
command would checkout an old `master` branch. It took me a while to
figure out what went wrong 😅


basile-henry:
Ah, it's only 1 commit off, I think I got confused by `vscode-language-dhall` vs `vscode-dhall-lsp-server`. But I still think we should change the link if future commits are to go to the `dhall-lang` version of either repo.

Gabriel439:
Thanks for fixing that!

PR #2119, opened 2020-12-31
Fix nested required fields

muff1nman:
For third level definitions and below, the required fields were looked up
from the top level definition rather than the parent (who isn't a top
level definition) causing these fields to be marked optional rather than
required.


Gabriel439 (approved):

Gabriel439:
@muff1nman: Thank you for fixing this! 🙂

PR #2121, opened 2021-01-02
Fix CORS compliance check

Gabriel439:
The CORS compliance check was previously not rejecting anything,
because the parent and child import were the same import, so it
would always succeed


sjakobi (approved):
Looks like something we ought to have a test for. No idea how to setup such a test though. Maybe record this on the `dhall-lang` issue tracker?!

Gabriel439:
@sjakobi: I created an issue to track that: https://github.com/dhall-lang/dhall-lang/issues/1130

PR #2124, opened 2021-01-04
Add support for explicit splitting of definitions

muff1nman:
The motivation behind this is better user experience when dealing
with Kubernetes CRDs. CRDs in an openapi definition are expressed
as one massive definition rather than the Kubernetes builtin types
which rarely nest and instead make use of references within openapi.
These references are desirable as it makes it more manageable to
sparsely define resources utilizing the nested defaults as you go down
the structure.

Unfortunately it looks like support for references wrt CRDs is unlikely
to be supported in the near future:

https://github.com/kubernetes/kubernetes/issues/62872

As a workaround, this adds a new option `--splitPaths` that takes a path
and an optional new model name. The path roughly emulates the idea
behind `kubectl explain` but I ended up using the `~` character instead
of `.` such that the full parent model name could still be used without
ambiguities. During type conversion any matches of path will cause a
Dhall reference to be injected with the nested definition being pushed
back on the stack of definitions to convert. The top level model name
can either be specified on the command line via `=com.some.ModelName` or
be guessed in the case the CRD author follows the best practice of using
the field name as the first word in the description.

Some future work that might need exploring related to this are:
1. Allow specifying `splitPaths` from a file
2. Clean up which types get accumulated into the `typesUnion`. I noticed
   while doing this that all? top level definitions are being dumped
   into this union where really it should only be top level definitions
   that have an `apiVersion` and `kind`
3. See if the `mergeNoConflicts` function needs to be improved such that
   it takes semantic equality (traversing imports) into account.

### DIFF
--- a/dhall-openapi/openapi-to-dhall/Main.hs
+++ b/dhall-openapi/openapi-to-dhall/Main.hs
@@ -17,7 +17,9 @@ import Text.Megaparsec
     , errorBundlePretty
     , parse
     , some
+    , optional
     , (<|>)
+    , eof
     )
 import Text.Megaparsec.Char            (alphaNumChar, char)
 
@@ -25,6 +27,7 @@ import Dhall.Kubernetes.Data           (patchCyclicImports)
 import Dhall.Kubernetes.Types
     ( DuplicateHandler
     , ModelName (..)
+    , ModelHierarchy
     , Prefix
     , Swagger (..)
     )
@@ -55,6 +58,7 @@ import qualified Text.Megaparsec.Char.Lexer            as Megaparsec.Lexer
 data Options = Options
     { skipDuplicates :: Bool
     , prefixMap :: Data.Map.Map Prefix Dhall.Import
+    , splits :: Data.Map.Map ModelHierarchy (Maybe ModelName)
     , filename :: String
     , crd :: Bool
     }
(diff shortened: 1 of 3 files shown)


dhall-openapi/openapi-to-dhall/Main.hs
```diff
@@ -177,8 +181,23 @@ parsePrefixMap =
...
     result = parse (Dhall.Parser.unParser parser `sepBy1` char ',') "MAPPING"
 
+parseSplits :: Options.Applicative.ReadM (Data.Map.Map ModelPath (Maybe ModelName))
+parseSplits =
+  Options.Applicative.eitherReader $ \s ->
+    bimap errorBundlePretty Data.Map.fromList $ result (pack s)
+  where
+    parser = do
+      path <- some (alphaNumChar <|> char '.' <|> char '-' <|> char '~')
+      model <- optional . try $ do
```
Gabriel439, line 191: I believe the `try` is not necessary here:

```suggestion
      model <- optional $ do
```

dhall-openapi/openapi-to-dhall/Main.hs
```diff
@@ -191,6 +210,16 @@ parseOptions = Options <$> parseSkip <*> parsePrefixMap' <*> fileArg <*> crdArg
         <> Options.Applicative.help "Specify prefix mappings as 'prefix1=importBase1,prefix2=importBase2,...'"
         <> Options.Applicative.metavar "MAPPING"
         )
+    parseSplits' =
+      option Data.Map.empty $ Options.Applicative.option parseSplits
+        (  Options.Applicative.long "splitPaths"
+        <> Options.Applicative.help
+          "Specifiy path and model name pairs with paths being delimited by '~' and pairs separated by '=' for which \ 
+          \definitions should be aritifically split with a ref: \n\
+          \'com.example.v1.Certificate~spec=com.example.v1.CertificateSpec'\n\
```
Gabriel439, line 219: Perhaps instead of `com.example.v1.Certificate~spec` we could use something like `(com.example.v1.Certificate).spec`, that way we can keep the `.` syntax for the path

dhall-openapi/src/Dhall/Kubernetes/Convert.hs
```diff
@@ -102,63 +109,121 @@ namespacedObjectFromImport _ = Nothing
...
+                     if   left == right 
+                     then left 
+                     else error ("Cannot merge differing values " ++ show left ++ " and " ++ show right ++ " for key " ++ show key))
+
+{- | Extract the 'ModelName' to be used when splitting a definition.
+
+    This is considered a guess as it does not work with all types. Currently it uses the first word from the description
+    appended to the largest prefix before the last @.@ of the parent.
+-}
+guessModelNameForSplit :: ModelHierarchy -> Definition -> Maybe ModelName
```
Gabriel439, line 125: I'm a bit concerned about guessing the model name from the first word of the description, since I'm still new to this domain and I don't know how much we can rely on users following this convention.  The main thing I'm trying to avoid is having to help users with debugging their CRD if things go wrong.  The other reason I bring this up is because it seems like we don't need to worry about the splits being particularly ergonomic since they only need to be specified once.

One intermediate alternative is perhaps to allow the user to omit the prefix from the model name, which would default to using the parent's prefix.  In other words, the user could specify:

```
com.example.v1.Certificate~spec=CertificateSpec
```

dhall-openapi/openapi-to-dhall/Main.hs
muff1nman: I like that idea.

dhall-openapi/src/Dhall/Kubernetes/Convert.hs
muff1nman: Here is my sample of CRDs that led me to think this was a good idea:
| CRD | Follows Convention |
| --- | --- |
| kubevirt | ✓ |
| cert-manager | ✗ |
| calico | ✓ |
| contour | ✓ |
| rook | ✗  |
| non public from my team | ✓ |

I would have to think on making the parents prefix optional. The one use case I wanted to support in the future is:
```
(com.example.ResourceSpec).podSelector=io.k8s.apimachinery.pkg.apis.meta.v1.LabelSelector
```
(Right now the only thing preventing that is `LabelSelector` has imports that a naiive `==` on `Dhall.Expr` cant see through.)

Gabriel439: Alright, then we can keep it the way it is

Gabriel439 (approved):

Gabriel439:
Feel free to merge after resolving the merge conflicts

PR #2126, opened 2021-01-04
dhall-openapi: support signed openapi integers

WillSewell:
OpenAPI integer types are signed ([see `integer`](http://spec.openapis.org/oas/v3.0.3#data-types)), but they are currently decoded into the Dhall Natural type, which is unsigned. This PR updates the behaviour to decode into Dhall Integers.

### DIFF
--- a/dhall-openapi/dhall-openapi.cabal
+++ b/dhall-openapi/dhall-openapi.cabal
@@ -51,6 +51,7 @@ library
     containers              >= 0.5.8.0   && < 0.7  ,
     dhall                   >= 1.35.0    && < 1.38 ,
     prettyprinter           >= 1.7.0     && < 1.8  ,
+    scientific              >= 0.3.0.0   && < 0.4  ,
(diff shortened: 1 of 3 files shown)


Gabriel439:
@WillSewell: Would it be possible to still use `Natural` when the OpenAPI type specifies a `minimum` of `0`?  I haven't checked how much work it would entail, so if it's prohibitive then don't worry about it

WillSewell:
Good point - I've pushed a PoC. I still need to test it out though.

WillSewell:
@Gabriel439 I think the most recent commit is correct. Unfortunately the kubernetes OpenAPI spec uses `minimum`, so all the integers in dhall-kubernetes are assumed to be `Integers`.

Gabriel439:
@WillSewell: Thank you!  Yeah, it's fine if the `Natural` support doesn't yet benefit `dhall-kubernetes` specifically.  We can always ask upstream to fix their OpenAPI specification if people want `Natural` numbers

PR #2128, opened 2021-01-06
dhall-openapi: make `metadata` optional for "inner" k8s types

WillSewell:
The `metadata` field is required for all top level resources
according to
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#metadata.

However for "inner" types this is not the case. For example
`io.k8s.api.batch.v1beta1.JobTemplateSpec` has a `metadata` field,
but it is not required. The example in the official k8s docs does
not include this field:
https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#example.

I searched the kubernetes 1.17 OpenAPI spec for "inner" types
that include a `metadata` field with:

```
curl -s https://raw.githubusercontent.com/kubernetes/kubernetes/release-1.17/api/openapi-spec/swagger.json | jq '.definitions | with_entries(select(.value.properties.apiVersion == null and .value.properties.metadata != null))'
```

It came up with:

* io.k8s.api.batch.v1beta1.JobTemplateSpec
* io.k8s.api.batch.v2alpha1.JobTemplateSpec
* io.k8s.api.core.v1.PodTemplateSpec

The pod spec template example in
https://kubernetes.io/docs/concepts/workloads/pods/
also does not include a `metadata` field.

This PR adds special cases to these types to make these fields
optional.

In general I think a nicer solution would be
https://github.com/dhall-lang/dhall-kubernetes/issues/8#issuecomment-755213541.

### DIFF
--- a/dhall-openapi/src/Dhall/Kubernetes/Convert.hs
+++ b/dhall-openapi/src/Dhall/Kubernetes/Convert.hs
@@ -65,6 +65,15 @@ requiredFields maybeName required
       , ( ModelName "io.k8s.api.core.v1.PersistentVolumeClaim"
         , Set.fromList [ FieldName "apiVersion", FieldName "kind" ]
         )
+      , ( ModelName "io.k8s.api.batch.v1beta1.JobTemplateSpec"
+        , Set.fromList [ FieldName "metadata" ]
+        )
+      , ( ModelName "io.k8s.api.batch.v2alpha1.JobTemplateSpec"
+        , Set.fromList [ FieldName "metadata" ]
+        )
+      , ( ModelName "io.k8s.api.core.v1.PodTemplateSpec"
+        , Set.fromList [ FieldName "metadata" ]
+        )
       ]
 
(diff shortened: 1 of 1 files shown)


muff1nman (approved):

WillSewell:
Thanks for reviewing @muff1nman. I'm conscious that this will lead to a breaking change in the dhall generated in dhall-kubernetes. Do you see this being a problem given release branches are not being used in that repository?

Gabriel439:
@WillSewell: We don't have release branches in `dhall-kubernetes`, but we do have release tags.  If you want, I can cut a new release tag before we incorporate this change

Either way, I think this is totally fine to merge

Gabriel439 (approved):
I just cut a [5.0.0 release for `dhall-kubernetes`](https://github.com/dhall-lang/dhall-kubernetes/releases/tag/v5.0.0) so this should be good to go

PR #2134, opened 2021-01-12
Only put models with GVK in types union (with some refactoring)

muff1nman:
The types unions main/only purpose was to allow creating lists of objects to
output to a single yaml document for sending to kubernetes. These
objects must have a group/version and kind. Therefore any objects
without would not be used in this manner and should be removed from the
typesUnion.

Let me know if you'd like these commits/prs split/merged differently

### DIFF
--- a/dhall-openapi/openapi-to-dhall/Main.hs
+++ b/dhall-openapi/openapi-to-dhall/Main.hs
@@ -7,6 +7,7 @@ import Control.Applicative.Combinators (option, sepBy1)
 import Data.Aeson                      (decodeFileStrict, eitherDecodeFileStrict)
 import Data.Bifunctor                  (bimap)
 import Data.Foldable                   (for_)
+import Data.Maybe                      (isJust)
 import Data.Text                       (Text, pack, unpack)
 import Data.Void                       (Void)
 import Dhall.Core                      (Expr (..))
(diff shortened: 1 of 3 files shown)


dhall-openapi/src/Dhall/Kubernetes/Convert.hs
```diff
@@ -315,46 +315,27 @@ toDefault prefixMap definitions modelName = go
...
       <$> modelNames
-      where
-        getKind (ModelName name) =
-          let elems = Text.split (== '.') name
-          in elems List.!! (length elems - 1)
+
+    toSimpleModelName :: ModelName -> SimpleModelName
+    toSimpleModelName (ModelName name) =
+      let elems = Text.split (== '.') name
+      in elems List.!! (length elems - 1)
```
Gabriel439, line 331: This can be made safer by defining a wrapper around `Text.split` that returns a `NonEmpty Text`:

```haskell
split :: (Char -> Bool) -> Text -> NonEmpty Text
split predicate text =
    case Text.split predicate text of
        []     -> "" :| []
        x : xs ->  x :| xs
```

… which is really what the type of `Text.split` should be anyway (since it never returns an empty list).  Then you can simplify this to:

```haskell
toSimpleModelName (ModelName name) = NonEmpty.last (split (== '.') name)
```

dhall-openapi/openapi-to-dhall/Main.hs
```diff
@@ -329,22 +331,25 @@ main = do
...
-      typesMap = getImportsMap "types" $ Data.Map.keys types
-      defaultsMap = getImportsMap "defaults" $ Data.Map.keys defaults
-      schemasMap = getImportsMap "schemas" $ Data.Map.keys schemas
+  let isStandalone model = and $ (\ def -> isJust $ Types.baseData def) <$> Data.Map.lookup model defs
 
-      typesRecordPath = "./types.dhall"
+  let makeRecord = Dhall.RecordLit . fmap Dhall.makeRecordField
+      typesMap = Dhall.Map.fromList $ List.sortOn snd $ Data.Map.toList $ Convert.groupBySimpleModelName duplicateHandler $ Data.Map.keys types
+      defaultsMap = Dhall.Map.filter (`elem` (Data.Map.keys defaults)) typesMap
+      schemasMap = Dhall.Map.filter (`elem` (Data.Map.keys schemas)) typesMap
```
Gabriel439, line 339:
```suggestion
      defaultsMap = Dhall.Map.filter (`elem` Data.Map.keys defaults) typesMap
      schemasMap = Dhall.Map.filter (`elem` Data.Map.keys schemas) typesMap
```

dhall-openapi/src/Dhall/Kubernetes/Convert.hs
```diff
@@ -315,46 +315,27 @@ toDefault prefixMap definitions modelName = go
...
 
     -- | There will be more than one namespaced object for a single object name
     --   (because different API versions, and objects move around packages but k8s
     --   cannot break compatibility so we have all of them), so we have to select one
     --   (and we error out if it's not so after the filtering)
-    selectObject :: (Text, [ModelName]) -> Maybe (ModelName, Text)
-    selectObject (kind, namespacedNames) = fmap (,kind) namespaced
+    selectObject :: ([ModelName]) -> Maybe ModelName
+    selectObject (namespacedNames) = namespaced
```
Gabriel439, line 338:
```suggestion
    selectObject namespacedNames = namespaced
```

Gabriel439 (commented):
Looks good!  Just a few comments

PR #2135, opened 2021-01-12
Add `full` duplicates option

muff1nman:
Sometimes even within the same kubernetes cluster you'd like to work
with different api versions of a given resource. This is currently not
easy considering that only a single "SimpleModelName" can be exposed in
the top level package.

This adds two --duplicates options.
- `full` which esckews the short names
  used previously in favor of the full model name from the openapi
  definition.
- `nested` which is similar to `full` but also splits on '.' making nested records.
  The exception is the typeUnion which keeps identical behavior to
  `full`.

Other instances where this is necessary include CRDs using the same kind
as native kubernetes type. I.e. Calico NetworkPolicy.


muff1nman:
Requires #2134

Issue #2142, opened 2021-01-28
Tighten base bounds on latest Dhall (1.38.0)


vmchale:
Has a silly little error with GHC 8.4.4:

```
Failed to build dhall-1.38.0.
Build log (
/home/vanessa/.cabal/logs/ghc-8.4.4/dhall-1.38.0-5f6c290ce61d2413104cf7b27d787d394acc9f059b32d25462e4185ed5f1af3c.log
):
Configuring library for dhall-1.38.0..
Preprocessing library for dhall-1.38.0..
Building library for dhall-1.38.0..

src/Dhall/Syntax.hs:7:14: error:
    Unsupported extension: GeneralisedNewtypeDeriving
    Perhaps you meant ‘GeneralizedNewtypeDeriving’ or ‘NoGeneralizedNewtypeDeriving’
  |
7 | {-# LANGUAGE GeneralisedNewtypeDeriving #-}
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^
cabal: Failed to build dhall-1.38.0 (which is required by
test:shake-dhall-test from shake-dhall-0.1.1.3). See the build log above for
details.
```

Gabriel439:
@vmchale: Alternatively, could we fix it to use `GeneralizedNewtypeDeriving` (with a `z`)?

Gabriel439:
Fix is up here: https://github.com/dhall-lang/dhall-haskell/pull/2143

vmchale:
Nice!

`dhall-1.38.0` needs a revision still, right?

1.38.1 could support GHC 8.4.4 ^

Gabriel439:
@vmchale: I just pushed a revision to amend the lower bound on `base` to `4.12.0.0`

vmchale:
Thanks!

PR #2143, opened 2021-01-28
Use `GeneralizedNewtypeDeriving` (with a 'z')

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/2142

The extension with an 's' is not supported on GHC 8.4.4


sjakobi (approved):
I'm surprised this wasn't caught in CI. If we want to continue to support GHC 8.4 we should probably use 8.4 in CI too.

Issue #1917, opened 2020-07-13
dhall-docs: directory without any dhall files are missing index.html


TristanCacqueray:
It seems like when a directory doesn't have a dhall file, then no index are generated, for example such tree:

* Zuul/Job/Type.dhall
* Zuul/Project/Type.dhall
* package.dhall

Results in:
```ShellSession
$ dhall-docs --input .
$ ls docs/Zuul/index.html
```

TristanCacqueray:
The `dhall-docs` command also get stuck when the `--input` path points at the directory without dhall file, e.g.:

```ShellSession
$ dhall-docs --input Zuul/
[command doesn't return]
```

german1608:
I can take a look at this after finishing #1908 but it looks like these are the guilty lines:

https://github.com/dhall-lang/dhall-haskell/blob/7a00835ea23372b993a14e6be81f5db091f199d7/dhall-docs/src/Dhall/Docs/Core.hs#L309-L330

What I could say now is that a folder without Dhall files won't have an index, so in your first example that is the expected behaviour. The reason for this is subjectively personal: I believe that navigation is improved if we don't have that index.html that only lists sub-folders. If we keep this logic, we will need to do an exception for the root index.

TristanCacqueray:
I forgot to mention the actual issue being that in this situation, the breadcrum link is a 404

SiriusStarr:
> The `dhall-docs` command also get stuck when the `--input` path points at the directory without dhall file, e.g.:
> 
> ```
> $ dhall-docs --input Zuul/
> [command doesn't return]
> ```

This bug still exists and caused me quite some issues today.  If you point `dhall-docs` at something like the following structure:
```
folder/
  subfolder/
    file.dhall

> dhall-docs --input folder
```
the command will hang forever (to the point that `SIGINT` with Ctrl-C won't close it and you have to `kill` it).

Even if we don't want to create `index.html`s for empty folders, this should probably be resolved to at least gracefully exist with a message to point it at a folder that actually contains `.dhall` files, along the lines of the

> No documentation was generated because no file with .dhall extension was found

message when you point it at a completely empty folder.

Gabriel439:
I have a PR up that fixes both issues: https://github.com/dhall-lang/dhall-haskell/pull/2150

PR #2150, opened 2021-02-20
Fix dhall-docs index generation

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/1917

This fixes two separate issues raised in #1917:

* Indices not being created for empty directories
* `dhall-docs` hanging when the top-level directory had no index


TristanCacqueray (approved):
index are created for directories without dhall files using this PR. Thanks!

sjakobi (approved):
LGTM! Fixing the tests might be as simple as running them with `--accept`. I hope this will also produce some regression tests.

Gabriel439:
Yeah, `cabal test tasty --test-option=--accept` did the trick

PR #2163, opened 2021-03-20
Upgrade Nixpkgs

Gabriel439:
This is necessary to support development on Big Sur


sjakobi (approved):

Gabriel439:
Hmmm.  It looks like this revision of Nixpkgs fails on Linux (and I can reproduce this on a Linux desktop).  I'll try again later with a different revision.

PR #2169, opened 2021-04-05
Provide multiple files to various CLI subcommands

basile-henry:
This PR makes it possible to provide multiple files as argument to the `format`, `freeze`, and `lint` subcommands.
These subcommands already had to deal with potentially multiple files via the use of the `transitive` flag.

# Motivation

It is common in unix tools for the CLI to accept multiple file arguments as input, so it would be great if `dhall` could do the same where possible/it makes sense.
A more concrete motivation is that I would like `dhall format` to be easily usable with formatting tools/managers such as https://github.com/numtide/treefmt. `treefmt` has a specification of sorts for what basic API formatters should expose as part of their CLI: https://numtide.github.io/treefmt/docs/formatters-spec.html. This specification requires the ability for a formatter to take multiple file arguments.

Providing multiple files as argument as opposed to calling these subcommands in a loop in `bash` (or with `xargs -n1`) has some performance benefits since we don't have to start a new process for every file.

# Changes

This PR make `inplace` implicit for file arguments (so that flag is removed). That was already kind of the case for these 3 subcommands as the only way to provide a file argument was `inplace` or `transitive` which implied `inplace`.

This PR changed the `transitive` flag from a flag that takes a file path to a switch that applies to all the input files.

Stdin is still supported, in 2 ways now: no file arguments, or a file argument `"-"`. This should enable applying the subcommands to both stdin and input files at the same time.
Note: stdin is still a non-transitive input even when the `transitive` switch is on.

---

Other subcommands might benefit from taking multiple files as input as well, I simply started with the ones I thought needed it the most.

### DIFF
--- a/dhall/src/Dhall/Util.hs
+++ b/dhall/src/Dhall/Util.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE NamedFieldPuns    #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards   #-}
 {-# LANGUAGE TypeApplications  #-}
@@ -19,7 +20,8 @@ module Dhall.Util
     , getExpressionAndHeaderFromStdinText
     , Header(..)
     , CheckFailed(..)
-    , mapMThrowCheckFailed
+    , MultipleCheckFailed(..)
+    , handleMultipleChecksFailed
     ) where
 
 import Control.Exception         (Exception (..))
@@ -167,15 +169,22 @@ data Output = StandardOutput | OutputFile FilePath
 -}
 data OutputMode = Write | Check
 
+newtype CheckFailed = CheckFailed { input :: Input }
+
 -- | Exception thrown when the @--check@ flag to a command-line subcommand fails
-data CheckFailed = CheckFailed { command :: Text, modified :: Text, input :: Input }
+data MultipleCheckFailed = MultipleCheckFailed
+  { command :: Text
+  , modified :: Text
+  , inputs :: NonEmpty Input
+  }
 
-instance Exception CheckFailed
+instance Exception MultipleCheckFailed
 
-instance Show CheckFailed where
-    show CheckFailed{..} =
-         _ERROR <> ": ❰dhall " <> command_ <> " --check❱ failed on " <> input_ <> "\n\
-        \\n\
+instance Show MultipleCheckFailed where
+    show MultipleCheckFailed{..} =
+         _ERROR <> ": ❰dhall " <> command_ <> " --check❱ failed on:\n\
+        \\n" <> files <>
+        "\n\
         \You ran ❰dhall " <> command_ <> " --check❱, but the input appears to have not\n\
         \been " <> modified_ <> " before, or was changed since the last time the input\n\
         \was " <> modified_ <> ".\n"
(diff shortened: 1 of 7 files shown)


dhall/src/Dhall/Util.hs
```diff
@@ -171,13 +168,13 @@ data Output = StandardOutput | OutputFile FilePath
 data OutputMode = Write | Check
 
 -- | Exception thrown when the @--check@ flag to a command-line subcommand fails
-data CheckFailed = CheckFailed { command :: Text, modified :: Text }
+data CheckFailed = CheckFailed { command :: Text, modified :: Text, input :: Input }
```
Gabriel439, line 171: Perhaps move the `command` and `modified` fields out of the `CheckFailed` type and into the `MultipleCheckFailed` type, since all the checks are going to share the same `command` / `modified` values.

This would also imply consolidating their error messages by moving the "You ran …" error header into `MultipleChecksFailed` so that it's only displayed once for all of the checks.

Gabriel439 (commented):

basile-henry:
Thanks for the review @Gabriel439 :smile:

TristanCacqueray (approved):
The change seems to work as expected, thanks! Would it be ok to keep a deprecated `--inplace` argument to ease the transition to avoid breaking CI?

basile-henry:
> Would it be ok to keep a deprecated `--inplace` argument to ease the transition

I am already breaking backwards compatibility with `--transitive` in this PR by making it a switch instead of an flag that takes an argument. So I'm not sure keeping a backward compatible `--inplace` (with argument I assume) makes that much sense. In general I think it's sensible to make breaking API changes in one go.

> to avoid breaking CI

In this PR? I thought everything passed at the moment, or are there other tests that are only run on `master` that I need to update?

I still need to have a closer look at documentation to make sure the CLI flags are up to date there as well.

I should also add a changelog entry for this change. :smile:

TristanCacqueray:
I agree it's better to make breaking API changes in one go. Though this one is likely to affect integration jobs, e.g. https://github.com/search?q=%22dhall+format+--inplace%22&type=Code .

Gabriel439:
One thing that can smooth the transition is to make `--inplace` a switch that does nothing but emits a warning (so that old usages of `dhall format --inplace someFile.dhall` still work during the migration period)

TristanCacqueray:
If the switch does nothing, then it might be better to make it fail though. Note that this will also break the vim plugin according to https://github.com/vmchale/dhall-vim/blob/master/ftplugin/dhall.vim#L32

basile-henry:
I have added `--inplace` back as a switch this time. It doesn't do anything other than outputting a notice of deprecation on stderr.
It is also hidden from `--help` which should hopefully discourage further use.

We can revert commit https://github.com/dhall-lang/dhall-haskell/pull/2169/commits/2d6035a62c81460fb7657aa1209679667c6f9224 at a later time if when we want to remove the flag entirely.

basile-henry:
It looks like [`dhall-lang`](https://github.com/dhall-lang/dhall-lang) will need to be updated after this change as well: https://github.com/dhall-lang/dhall-lang/search?q=%22--inplace%22

But at least with this deprecation approach we shouldn't break too many workflows/scripts. Do you think something similar is needed for `--transitive`? I expect that to be used much less often

Gabriel439 (approved):
Sorry, I missed that you moved this out of Draft status.  This looks great!

basile-henry:
No worries, I don't work on this as often as I'd like, so slow PR progress works for me. As long as it's not blocking others.

But I think it's ready now, so I will merge it, and hopefully not too many people will have issues with the new CLI API 😁

Issue #2109, opened 2020-12-04
treat comment before let the same as comment between let and binding when formatting


adamse:
It feels very natural to put comments before a let binding:

```
let b = 1
-- a comment
let value = something complicated
in value b
```

This currently is formatted as, losing the comment :(,
```
let b = 1 let value = something complicated in value b
```

But maybe we can treat it the same as 
```
let b = 1
let 
  -- a comment
  value = something complicated
in value b
```

PR #2173, opened 2021-04-19
Strict comment parsing

basile-henry:
This PR adds a way to control if comments should be parsed as whitespace.

I believe many of the issues we are seeing where we lose comments when formatting Dhall (#2172, #2109, and probably others) stem from the fact that at the core of the parser comments are treated as whitespace.
We have since added some logic to recover comments from whitespace source spans and keep track of them in the AST with some success. Unfortunately it is very easy to leave out places in the AST where comments could be and silently lose them. So I think the way forward is instead to parse whitespace and comments separately when doing formatting (or anything that pretty prints the original source code). This is not necessary and probably redundant work for the parser when typechecking/normalizing or doing anything that doesn't need to preserve comments for pretty printing (i.e. it is okay to lose comments when normalising Dhall).

My approach is to make what we consider whitespace configurable. When we care about comments we should not consider them to be whitespace.
I have set the default for this configuration to parsing whitespace/comments the same as it was before so that everything works as usual. But now `dhall` has an extra flag `--strict-comment-parsing` used by `format`, `freeze`, `lint`, and `schema` which explicitly doesn't consider comments to be whitespace. Currently it basically errors out (while parsing) on any comment, which is not immediately useful. My intention is to slowly move the logic we have for parsing comments via whitespace to explicitly parsing comments instead.

@Profpatsch  on [twitter](https://twitter.com/Profpatsch/status/1383759730762342404?s=20):
> I’d much rather it errored out than throwing away comments tbh.

I agree with him in that until we are certain that we don't lose any comments, no matter their location, we should provide a way to make `dhall format` error out loudly to make it obvious that a comment is being lost. I believe this `--strict-comment-parsing` flag could be a solution to this (at least once I get all the comments we currently do keep to not trigger an error :sweat_smile: ). This will help us keep track of which comment locations we do support in the meantime (both manually using the flag and in test suites).

The main reason I am opening this PR as a draft in the such an early state of progress is because I suspect others have thought about this issue as well and maybe have better ideas on how to tackle it. Are there any issues I missed discussing something like this?

I don't know if this approach is great, I expect there might be some performance regression when parsing since I have added an extra level of `ReaderT` to the parsing `Monad`. I would need to run some of the benchmarks to make sure either way. The environment provided as part of `ReaderT` is rather static (a comment is considered whitespace or it is not), so maybe a different solution would be more appropriate (`ImplicitParams`, manually threading the config).

What do you think? Should I carry on with this approach? :smile:


Gabriel439:
@basile-henry: Yeah, I think this is a good idea.  My only suggestion is to make the strict behavior the default, with a flag to opt out and recover the old behavior

basile-henry:
> My only suggestion is to make the strict behavior the default, with a flag to opt out and recover the old behavior

Should the strict behaviour only be used for `format`, `freeze`, or anything that modifies the source code? I am not sure we want strict parsing for evaluation for example, dropping comments when evaluating is not really an issue :sweat_smile:

Gabriel439:
@basile-henry: Yeah, I meant to say just for the modification commands like `dhall {format,freeze,lint}`

dhall/src/Dhall/Parser/Expression.hs
```diff
@@ -954,6 +954,26 @@ parsers embedded = Parsers {..}
...
+
+-- | Parse a multi comment or whitespace
+commentOrWhitespace :: Parser (Maybe MultiComment)
+commentOrWhitespace = setCommentControl CommentIsNeeded $ -- Strict whitespace
+    do whitespace; optional multiComment
+
+-- | Parse a multi comment or non-empty whitespace
+commentOrNonEmptyWhitespace :: Parser (Maybe MultiComment)
+commentOrNonEmptyWhitespace = setCommentControl CommentIsNeeded $ -- Strict whitespace
+        try (Just <$ whitespace <*> multiComment)
```
Gabriel439, line 974: Generally you want to minimize use of `try`, both for performance reasons and better error messages.  What I'd suggest here to remove the dependence on `try` is something like:

```haskell
commentOrNonEmptyWhitespace = setCommentControl CommentIsNeeded $ do
    m <- optional whitespaceChunk

    case m of
        Nothing -> Just <$> multiComment
        Just () -> whitespace *> optional multiComment 
```

dhall/src/Dhall/Syntax.hs
```diff
@@ -196,33 +217,34 @@ instance Pretty Var where
 --
 -- … will be instantiated as follows:
 --
--- * @bindingSrc0@ corresponds to the @A@ comment.
+-- * @bindingComment0@ corresponds to the @A@ comment.
 -- * @variable@ is @"x"@
--- * @bindingSrc1@ corresponds to the @B@ comment.
+-- * @bindingComment1@ corresponds to the @B@ comment.
```
Gabriel439, line 222: Don't forget to also document what `variableSrc` would correspond to in this example

dhall/src/Dhall/Parser/Token.hs
```diff
@@ -318,13 +317,16 @@ identifier = do
...
+whitespaceChunk = do
+    commentControl <- askCommentControl
+
+    choice (concat
+        [ [ void (Dhall.Parser.Combinators.takeWhile1 predicate)
+          , void (Text.Parser.Char.text "\r\n" <?> "newline")
+          ]
+        , [ void lineComment | commentControl == CommentIsWhitespace ]
+        , [ void blockComment | commentControl == CommentIsWhitespace ]
+        ]) <?> "whitespace"
```
Gabriel439, line 329:
```suggestion
    let descriptor = case commentControl of
            CommentIsWhiteSpace -> "whitespace"
            CommentIsNeeded     -> "comment-free whitespace"

    choice (concat
        [ [ void (Dhall.Parser.Combinators.takeWhile1 predicate)
          , void (Text.Parser.Char.text "\r\n" <?> "newline")
          ]
        , [ void lineComment | commentControl == CommentIsWhitespace ]
        , [ void blockComment | commentControl == CommentIsWhitespace ]
        ]) <?> descriptor
```

dhall/src/Dhall/Parser/Expression.hs
basile-henry: Great, thanks for the explanation! :smile:

dhall/src/Dhall/Parser.hs
```diff
@@ -82,7 +84,7 @@ exprFromText
             --   used in parsing error messages
   -> Text   -- ^ Input expression to parse
   -> Either ParseError (Expr Src Import)
-exprFromText delta text = fmap snd (exprAndHeaderFromText delta text)
+exprFromText delta text = fmap snd (exprAndHeaderFromText CommentIsWhitespace delta text)
```
basile-henry, line 87: Todo: take this as an argument just like in `exprAndHeaderFromText`. Caller's responsibility to set comment control.

dhall/src/Dhall/Parser/Expression.hs
Gabriel439: You're welcome! 🙂

dhall/src/Dhall/Syntax.hs
```diff
@@ -188,6 +192,24 @@ instance IsString Var where
 instance Pretty Var where
     pretty = Pretty.unAnnotate . prettyVar
 
+-- | Keep track of potential multiple comments as part of Dhall syntax. This is
+-- useful for source preserving transformations such as /format/.
+newtype MultiComment = MultiComment { getMultiComment :: NonEmpty Comment }
+  deriving stock (Data, Generic, Eq, Ord, Show, Lift)
```
Gabriel439, line 198: It will probably be more efficient to derive `newtype` for `Eq` and `Ord`

dhall/src/Dhall/Syntax.hs
```diff
@@ -188,6 +192,24 @@ instance IsString Var where
...
+  deriving newtype (Semigroup)
+  deriving anyclass NFData
+
+-- | Keep track of a comment as part of Dhall syntax. This is useful for source
+-- preserving transformations such as /format/.
+data Comment
+  = LineComment Text
+  -- ^ Corresponds to a single line comment: `--<text>$` where <text> does not
```
Gabriel439, line 206: `haddock` will try to link anything within backticks:

```suggestion
  -- ^ Corresponds to a single line comment: @--<text>$@ where @<text>@ does not
```

dhall/src/Dhall/Syntax.hs
```diff
@@ -188,6 +192,24 @@ instance IsString Var where
...
+  -- contain newline characters
+  | BlockComment Text
+  -- ^ Corresponds to a multi line comment: `{-<text>-}` where <text> can
```
Gabriel439, line 209:
```suggestion
  -- ^ Corresponds to a multi line comment: @{-<text>-}@ where <text> can
```

dhall/src/Dhall/Main.hs
```diff
@@ -205,6 +208,13 @@ parseOptions =
             <>  Options.Applicative.help description
             )
 
+    parseCommentControl = fmap f $ switch
+        "loose-comment-parsing"
+        "Ignore parse error for unhandled comments, but they will get dropped (please submit an issue for such cases)"
```
Gabriel439, line 213:
```suggestion
        "discard-unsupported-comments"
        "Discard comments that cannot be auto-formatted instead of failing (but continue to submit issues for any comments the auto-formatter rejects)"
```

dhall/src/Dhall/Parser.hs
Gabriel439: Do you still plan to do this?

dhall-openapi/openapi-to-dhall/Main.hs
```diff
@@ -182,7 +185,10 @@ parsePrefixMap =
       e <- Dhall.Parser.expr
       imp <- parseImport prefix e
       return (pack prefix, imp)
-    result = parse ((Dhall.Parser.unParser parser `sepBy1` char ',') <* eof) "MAPPING"
+
+    parser' = runReaderT (Dhall.Parser.unParser parser) Dhall.Util.CommentIsWhitespace
```
Gabriel439, line 189: Should this use `runParser`?

dhall-openapi/openapi-to-dhall/Main.hs
```diff
@@ -198,7 +204,10 @@ parseSplits =
         mo <- parseModel
         return mo
       return (path, model)
-    result = parse ((Dhall.Parser.unParser parser `sepBy1` char ',') <* eof) "MAPPING"
+
+    parser' = runReaderT (Dhall.Parser.unParser parser) Dhall.Util.CommentIsWhitespace
```
Gabriel439, line 208: Same question here: should this use `runParser`?

dhall/src/Dhall/Parser/Expression.hs
```diff
@@ -963,6 +963,29 @@ parsers embedded = Parsers {..}
...
+    MultiComment <$> Combinators.NonEmpty.sepEndBy1 comment whitespace
+
+-- | Parse a multi comment or whitespace
+commentOrWhitespace :: Parser (Maybe MultiComment)
+commentOrWhitespace = setCommentControl CommentIsNeeded $ -- Strict whitespace
+    do whitespace; optional multiComment
+
+-- | Parse a multi comment or non-empty whitespace
+commentOrNonEmptyWhitespace :: Parser (Maybe MultiComment)
+commentOrNonEmptyWhitespace = setCommentControl CommentIsNeeded $ do -- Strict whitespace
```
Gabriel439, line 982: I suggest moving the `--Strict whitespace` comment to the documentation for the `CommentIsNeeded` constructor

```suggestion
-- | Parse a multi comment (which contains at least one comment)
multiComment :: Parser MultiComment
multiComment = setCommentControl CommentIsNeeded $
    MultiComment <$> Combinators.NonEmpty.sepEndBy1 comment whitespace

-- | Parse a multi comment or whitespace
commentOrWhitespace :: Parser (Maybe MultiComment)
commentOrWhitespace = setCommentControl CommentIsNeeded $
    do whitespace; optional multiComment

-- | Parse a multi comment or non-empty whitespace
commentOrNonEmptyWhitespace :: Parser (Maybe MultiComment)
commentOrNonEmptyWhitespace = setCommentControl CommentIsNeeded $ do
```

dhall/src/Dhall/Parser/Combinators.hs
```diff
@@ -69,11 +74,42 @@ laxSrcEq (Src p q _) (Src p' q' _) = eq p p' && eq q q'
...
 {-# INLINE laxSrcEq #-}
 
+-- | A way to control how comments get parsed
+--
+-- Parsing comment as whitespace is fast and convenient when we don't care
+-- about the comments. We often need to keep comments thoughout
+-- transformations.
+data CommentControl
+  = CommentIsWhitespace
+  | CommentIsNeeded
```
Gabriel439, line 84: I think I would suggest these names and documentation:

```suggestion
data WhitespaceControl
  = UnsupportedCommentsForbidden
  -- ^ The @whitespace@ parser rejects comments, meaning that comments need to be
  -- parsed explicitly in order to avoid parse errors.  We use this to detect comments that are
  -- not yet preserved and formatted.
  | UnsupportedCommentsPermitted
  -- ^ The @whitespace@ parser accepts comments, meaning that comments are accepted
  -- anywhere within the source code, even if they are not preserved or formatted.
```

dhall/src/Dhall/Pretty/Internal.hs
```diff
@@ -171,107 +171,76 @@ prettySrcExpr = prettyCharacterSet Unicode
...
+
+    toLines = \case
+        BlockComment txt
+          | Just rest <- Text.stripPrefix "{-" txt
+          , Just body <- Text.stripSuffix "-}" rest
+          -> Text.strip <$> Text.lines body
+        LineComment txt
+          | Just body <- Text.stripPrefix "--" txt
+          -> [ Text.strip body ]
+        _ -> error "Dhall.Pretty.Internal.normalizeMultiComment: Unexpected comment"
```
Gabriel439, line 206: I suggest using `Dhall.Core.internalError` here and including the text of the comment in the error message

dhall/src/Dhall/Pretty/Internal.hs
```diff
@@ -171,107 +171,76 @@ prettySrcExpr = prettyCharacterSet Unicode
...
+        LineComment _ -> False
+
+    stripStartEmptyLines = dropWhile Text.null
+
+    -- Remove leading and terminating empty lines
+    stripEmptyLines =
+          reverse
+        . stripStartEmptyLines
+        . reverse
+        . stripStartEmptyLines
```
Gabriel439, line 191: You can inline the `dropWhileEnd'` function from here:

https://hackage.haskell.org/package/extra-1.7.9/docs/src/Data.List.Extra.html#dropWhileEnd%27

… which will be more efficient (and lazy) than reversing the list

dhall/src/Dhall/Pretty/Internal.hs
```diff
@@ -171,107 +171,76 @@ prettySrcExpr = prettyCharacterSet Unicode
...
-renderSrcMaybe _ = Nothing
-
-{-| @
-    'containsComment' mSrc ≡ 'Data.Maybe.isJust' ('renderSrcMaybe' mSrc)
-    @
--}
-containsComment :: Maybe Src -> Bool
-containsComment Nothing        = False
-containsComment (Just Src{..}) = not (Text.all isWhitespace srcText)
+    f x y= x <> Pretty.hardline <> y
```
Gabriel439, line 243:
```suggestion
    f x y = x <> Pretty.hardline <> y
```

Gabriel439 (commented):

Also, I think you will want to update the `dhall-lang` submodule to the latest `master`

dhall/tests/format/recordLitComments1B.dhall
```diff
@@ -1,3 +1,3 @@
-{ --| a comment for y
+{ -- | a comment for y
```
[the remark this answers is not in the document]
basile-henry: Good point! :smile:

dhall/tests/Dhall/Test/QuickCheck.hs
```diff
@@ -763,6 +802,17 @@ tests =
           , Test.QuickCheck.property idempotenceTest
           , adjustQuickCheckTests 10000
           )
+        , ( "Any whitespace could be a comment and still parse"
+          , Test.QuickCheck.property commentAsWhitespace
+          , -- adjustQuickCheckTests 10000
```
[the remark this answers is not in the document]
basile-henry: I will remove it. Ultimately when we can parse comments anywhere it'll probably end up looking like that, but we can worry about that when we get there :smile:

dhall/src/Dhall/Main.hs
basile-henry: Nice! Much better wording :+1:

dhall/src/Dhall/Parser.hs
basile-henry: I am actually not sure any more. This `exprFromText` function seems to be used in many places. It would be convenient to not have to specify the `CommentControl`/`WhitespaceControl` when using it. I don't think this control is really necessary outside of `format`, `freeze`, and `lint` anyway.

dhall-openapi/openapi-to-dhall/Main.hs
basile-henry: Yes! I must have thought applying `unParser` before adding other combinators (`sepBy1...`) originally mattered which is why I applied the reader manually here.

But it looks like the other parser combinators apply perfectly well to our custom parser :+1:

dhall/src/Dhall/Pretty/Internal.hs
basile-henry: We don't currently depend on `extra` as far as I can tell. Is the version from [`base`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-List.html#v:dropWhileEnd) sufficient?

basile-henry:
Thanks for the review! 😄


It looks like my latest changes to the comment rendering logic broke the idempotence tests.
I will have another go at fixing this, maybe by going back to stripping leading whitespace in comments.
But I won't have the time today or tomorrow.

basile-henry:
It looks like the way I am rendering the `MultiComment`s (by merging them) is incompatible with some of the assumptions around how comments are kept after formatting in `dhall-docs`.

I will remove this merging behaviour. Hopefully that should be enough to get `dhall-docs` to keep on working as expected.
In a future PR it might make sense to unify the comment parsing logic between `dhall` the library and `dhall-docs`. I believe a bit more info will need to be kept in order to keep groups of single line comments as groups, something I haven't considered in this PR.

Gabriel439:
@basile-henry: Another option might be to change `dhall-docs` to permit those comments.  It might actually be a feature to support multiple block comments for documenting something and merging them in the same way as line comments

basile-henry:
@Gabriel439 

You're right, it might be nice to merge consecutive doc comments as well.

I think the issue is merging doc comments with plain comments. That is probably unexpected behaviour as it adds documentation that was intended as a plain comment. I will need to handle that specifically.

Slow progress, I thought this PR was done 😅
I should have some time tomorrow to finish this new bit.

Profpatsch:
Chiming in to give some cheers :1st_place_medal:

basile-henry:
Thanks! Cheers appreciated :smile:

I will fix the remaining `dhall-docs` test failures ASAP so we can merge this soon. :crossed_fingers: 

Once this PR goes in, the main difference to the users of `dhall format` will be that fewer programs will parse by default. This doesn't sound like a great feature (:sweat_smile:)  but hopefully this should enable us to go and explicitly parse comments everywhere they can be.

PR #2175, opened 2021-05-01
Add support for Unix shebangs

Gabriel439:
… as standardized in https://github.com/dhall-lang/dhall-lang/pull/1158

### DIFF
--- a/dhall/src/Dhall/Parser/Expression.hs
+++ b/dhall/src/Dhall/Parser/Expression.hs
@@ -10,7 +10,6 @@ module Dhall.Parser.Expression where
 import Control.Applicative     (Alternative (..), liftA2, optional)
 import Data.ByteArray.Encoding (Base (..))
 import Data.Foldable           (foldl')
-import Data.Functor            (void)
 import Data.List.NonEmpty      (NonEmpty (..))
 import Data.Text               (Text)
 import Dhall.Src               (Src (..))
@@ -113,7 +112,17 @@ data Parsers a = Parsers
 parsers :: forall a. Parser a -> Parsers a
 parsers embedded = Parsers {..}
   where
-    completeExpression_ = whitespace *> expression <* whitespace
+    completeExpression_ =
+        optional shebang *> whitespace *> expression <* whitespace
+
+    shebang = do
+        _ <- text "#!"
+
+        let predicate c = ('\x20' <= c && c <= '\x10FFFF') || c == '\t'
+
+        _ <- Dhall.Parser.Combinators.takeWhile predicate
+
+        endOfLine
 
     expression =
         noted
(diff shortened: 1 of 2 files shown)


dhall/src/Dhall/Parser/Expression.hs
```diff
@@ -757,7 +766,7 @@ parsers embedded = Parsers {..}
                   where
                     predicate c = c == '$' || c == '\''
 
-                endOfLine = do
+                endOfLine_ = do
                     a <- "\n" <|> "\r\n"
```
TristanCacqueray, line 770: nit: could this be replaced by `a <- endOfLine` ?

basile-henry:
It would be nice to support multiple lines of `shebang` so that we could use the `nix-shell` shebangs.

I have just tested this branch with the following (from the root of the `dhall-haskell` repo):

```dhall
#! /usr/bin/env nix-shell
#! nix-shell -i "dhall --file" -p "(import ./.).dhall"

42
```

Which currently gives me:

```bash
➜ ./test.dhall
dhall:
Error: Invalid input

./test.dhall:2:1:
  |
2 | #! nix-shell -i "dhall --file" -p "(import ./.).dhall"
  | ^^^^^^^^
unexpected "#! nix-s"
expecting expression or whitespace
```

I saw the `dhall-lang` PR for this, but didn't think about mentioning the `nix-shell` shebang use case at the time :sweat_smile:

Gabriel439:
@basile-henry: Yeah, that would first require an upstream change to the standard, which we can still do afterwards

TristanCacqueray:
It seems like `nix-shell` shebang is relying on the fact that `#` is a comment in the target language, but that does not always work. Though that seems like a good thing to support.

Alternatively, perhaps nix-shell could be modified to also consider `--! nix-shell -i "dhall --file" -p dhall` as a valid extra shebang line, but that would not look good.

TristanCacqueray:
Also it is unfortunate that `#!/usr/bin/env -S nix-shell -p dhall -i "dhall --file"` does not work.

basile-henry:
> `nix-shell` shebang is relying on the fact that `#` is a comment in the target language

That's probably true for `bash` and `python`. However I have seen multi-line shebangs with `nix-shell` for Haskell too, usually with GHC's `runhaskell`. That probably means that GHC takes this edge case specifically into account?

basile-henry (approved):

TristanCacqueray:
`runhaskell` does something special where it ignores any lines starting with '#!', not just the first ones, according to: https://gitlab.haskell.org/ghc/ghc/-/blob/e61d2d47c4942c829ec98016be5aa5ae36982524/compiler/GHC/Parser/Lexer.x#L279

PR #2187, opened 2021-05-12
dhall-nix{,pkgs}: cabal: dep: allow hnix 0.13

Anton-Latukha:
Looked into the use of the HNix modules in both projects.

Despite a huge [changelog](https://hackage.haskell.org/package/hnix/changelog) of the HNix release, there were no changes in `Shorthands` or `Pretty` that Dhall relies on. Therefore the straight-forward update.


Anton-Latukha:
There is update already: #2186

Closing.

PR #2188, opened 2021-05-15
Fix compatibility with tasty-hspec 1.1.7

felixonmars:
tasty-hspec no longer re-exports Test.Hspec since 1.1.7. Let's import
it ourselves.


Gabriel439:
@felixonmars: Thank you! 🙂

PR #2195, opened 2021-05-25
Report input name in parsing errors

### LINKED ISSUE #2194: Errors do not specify which file they occur in when multiple files are passed to a subcommand
#2169 was great and allows one to pass multiple files at once to e.g. `dhall freeze`.  However, currently no information about *which* file fails is output to the user, making it very difficult to track down the problem if many files are passed.

For example:

```bash
echo 0 > a.dhall 

echo 0f > b.dhall

dhall freeze a.dhall b.dhall
dhall: 
Error: Invalid input

(input):1:2:
  |
1 | 0f
  |  ^
unexpected 'f'
expecting "→", "∧", "⩓", "⫽", ->, //, //\\, /\, :, end of input, keyword, operator, or whitespace
```

If instead the error were output as:

```bash
dhall: 
Error: Invalid input

(./b.dhall):1:2:
  |
1 | 0f
  |  ^
unexpected 'f'
expecting "→", "∧", "⩓", "⫽", ->, //, //\\, /\, :, end of input, keyword, operator, or whitespace
```

or something instead of just `(input)` it would be much more clear.

This also occurs with e.g. `dhall lint`

As such, I've been unable to use the new multi-file commands in my CI and am stuck doing something like:

```bash
for n; do echo "$n"; dhall lint --check "$n" || exit 1; done
```

which of course still works but it would be great to use the new option to process multiple files at once, since it's currently a pain to track down what the issue is if you're linting 500 files at once.  :smile:

basile-henry:
Fixes #2194

Parser errors in `format`, `freeze`, `lint` didn't report which input they came from. This wasn't an issue until we started supporting multiple inputs.

### DIFF
--- a/dhall/src/Dhall/Util.hs
+++ b/dhall/src/Dhall/Util.hs
@@ -118,13 +118,13 @@ get parser censor input = do
         case input of
             Input_ (InputFile file) -> Data.Text.IO.readFile file
             Input_ StandardInput    -> Data.Text.IO.getContents
-            StdinText text          -> pure text
+            StdinText _ text        -> pure text
 
     let name =
             case input of
                 Input_ (InputFile file) -> file
                 Input_ StandardInput    -> "(input)"
-                StdinText _             -> "(input)"
+                StdinText inputName _   -> inputName
 
     let result = parser name inText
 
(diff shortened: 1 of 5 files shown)


dhall/src/Dhall/Util.hs
```diff
@@ -149,7 +149,11 @@ data Censor = NoCensor | Censor
 data Input = StandardInput | InputFile FilePath deriving (Eq)
 
 -- | Path to input or raw input text, necessary since we can't read STDIN twice
-data InputOrTextFromStdin = Input_ Input | StdinText Text
+data InputOrTextFromStdin
+    = Input_ Input
+    | StdinText String Text
+    -- ^ StdinText name text where name is a use-friendly name describing the
```
basile-henry, line 155:
```suggestion
    -- ^ @StdinText name text@ where name is a user-friendly name describing the
```

TristanCacqueray (approved):
Thanks, that is looking good to me.

Gabriel439:
@basile-henry: Thank you for fixing that! 🙂

SiriusStarr (commented):
LGTM!  Thanks for the quick fix.  :-)

PR #2198, opened 2021-05-26
Dhall quasi-quoter

### LINKED ISSUE #869: Dhall quasiquoter?
It would be super useful to have quasiquotation support ala http://hackage.haskell.org/package/shakespeare  for code generators written in Haskell. It would make writing things like dhall-kubernetes really easy whilst guaranteeing well-formed code.

mmhat:
This PR adds a simple quasi-quoter name `dhall` to Dhall.TH.
It can be used to insert Dhall expressions in Haskell expression contexts at compile time.
It is based on the `staticDhallExpression` function residing in the same module.

Fixes #869


dhall/src/Dhall/TH.hs
```diff
@@ -83,6 +84,23 @@ staticDhallExpression text = do
...
+    This quoter is meant to be used in expression context only; Other contexts
+    like pattern contexts or declaration contexts are not supported and will
+    result in an error.
+-}
+dhall :: QuasiQuoter
+dhall = QuasiQuoter
+    { quoteExp = staticDhallExpression . Text.pack
+    , quotePat = const $ error "dhall quasi-quoter: Quoting pattern is not supported !"
+    , quoteType = const $ error "dhall quasi-quoter: Quoting types is not supported !"
+    , quoteDec = const $ error "dhall quasi-quoter: Quoting declarations is not supported !"
```
Gabriel439, line 101:
```suggestion
    , quotePat = const $ error "dhall quasi-quoter: Quoting patterns is not supported!"
    , quoteType = const $ error "dhall quasi-quoter: Quoting types is not supported!"
    , quoteDec = const $ error "dhall quasi-quoter: Quoting declarations is not supported!"
```

Gabriel439 (approved):
Great idea!  Just one minor suggestion

PR #2232, opened 2021-06-26
Exclude dhall-nix on Ubuntu from CI

mmhat:
The dhall-nix build on Ubuntu [fails](https://github.com/dhall-lang/dhall-haskell/runs/2920388542) with LTS 18.0. The reason is that `libsodium`, which is an external dependency of `saltine` is not found. This PR removes dhall-nix/Ubuntu from the build matrix. While it is in general a Bad Idea to "fix" failing CI jobs by removing them it is IMHO not a big deal in this case: We already cover the Linux builds with Hydra which handles external dependencies correctly. The inclusion of Ubuntu in the build matrix was therefore highly optional in the first place.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,6 +58,15 @@ jobs:
               bin/dhall-to-yaml-ng
               bin/yaml-to-dhall
         exclude:
+          # Temporarily exclude dhall-nix on Ubuntu due to build failure
+          # See: https://github.com/dhall-lang/dhall-haskell/runs/2920388542
+          - os:
+              runner: ubuntu-latest
+              package: null
+            package:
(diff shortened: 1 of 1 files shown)


mmhat:
@Gabriel439 I noticed that every now and then a build step hangs indefinitely. See for example [this](https://github.com/dhall-lang/dhall-haskell/runs/2920388606) job. Do you know if we are running into any limits here? According to the [docs](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration) we shouldn't.

Gabriel439:
@mmhat: My guess is probably a property test is hitting a case where the code hangs

Gabriel439 (approved):

PR #2236, opened 2021-06-29
Support for headers.dhall configuration

timbertson:
This PR implements support for per-host header configuration, as standardised in https://github.com/dhall-lang/dhall-lang/pull/1192.

The basic idea is that instead of headers being inline with an import (`URL with toMap { Authorization = "TOKEN" }`), you will be able to specify per-origin headers in a config file, e.g:

```dhall
-- ~/.config/dhall/headers.dhall
toMap {
  `raw.githubusercontent.com:443` = toMap { Authorization = "TOKEN" }
}
```

TODO:

 - [x] standardisation in dhall-lang (ready to merge)
 - [x] load dhall instead of JSON (needs to be a reduced subset that doesn't support remote imports)
 - [x] tests
 - [x] merge https://github.com/dhall-lang/dhall-haskell/pull/2268
 - [x] cache config contents (only load at most once)
 - [x] error reporting


Gabriel439:
@timbertson: Sorry for the delayed review!

My first general comment is that I usually do the standardization in dhall-lang first before I begin work on the Haskell implementation, for two main reasons:

* Forcing things to go through the standard acts like a simplicity filter

  The natural deduction pseudo-code we use usually keeps things from growing too complex, because it is so restricted in what it can express

* Working on the standard first helps me to design the feature and solicit feedback at a higher level

That said, I don't foresee any issue standardizing this (it seems pretty easy to specify using our existing notation)

timbertson:
I think this is getting close to ready. I'm pretty happy with how everything hangs together now.

The one thing I'm not quite satisfied with is the error reporting. When loading the headers, we need a fresh Status (because e.g. it shouldn't attempt remote imports, or loading headers recursively). I copy the `_stack` from the real import though in order to represent a useful call stack (i.e. "why was the header file being loaded").

But it doesn't seem to work quite right. In some cases it reports a good looking stack, but then also reports an additional error:

```
$ env DHALL_HEADERS=https://example.com ./dist/build/Dhall/dhall --file dhall-lang/tests/import/success/userHeadersA.dhall
dhall:
↳ https://httpbin.org/user-agent as Text
  ↳ env:DHALL_HEADERS
    ↳ https://example.com/

Error: Cannot import a remote URL from the headers configuration expression.


1│ https://example.com

1:1
```

The first half is good, but I think it's trying to display some dhall code after `1:1` and struggling.


In other cases, it doesn't show a stack at all:

```
$ env 'DHALL_HEADERS={ foo }' ./dist/build/Dhall/dhall --file dhall-lang/tests/import/success/userHeadersA.dhall

Error: Unbound variable: foo

1│ { foo }

1:1
```

Is this because of the type of the error?


A parse error seems to be reported well, at least:

```
$ env 'DHALL_HEADERS={ foo' ./dist/build/Dhall/dhall --file dhall-lang/tests/import/success/userHeadersA.dhall
dhall:
↳ https://httpbin.org/user-agent as Text
  ↳ env:DHALL_HEADERS

Error: Invalid input

1:6:
  |
1 | { foo
  |      ^
unexpected end of input
expecting ',', =, operator, whitespace, or }
```

timbertson:
(the error handling code in question is in Import.hs `siteHeadersLoader`)

Gabriel439:
@timbertson: So I think the solution to this problem is to stick to https://github.com/dhall-lang/dhall-lang/pull/1192 as written.  In particular, this section of the import resolution logic:

```
Γ(env:DHALL_HEADERS ? "${XDG_CONFIG_HOME}/dhall/headers.dhall" ? ~/.config/dhall/headers.dhall ? []) = userHeaders
getKey(userHeaders, origin, []) = headers  ; Extract the first `mapValue` from `userHeaders`
```

The precise interpretation of that is that the `userHeaders` fetched from one of the standard headers locations is not further transitively resolved (or even interpreted at all).  In other words, it needs to be a record.  So based on that interpretation then you don't need to use the `_stack` at all to transitively resolve the headers expression; just require the imported expression to be a record.  Specifically, the `siteHeadersLoader` function would use `Dhall.Import.assertNoImports` instead of `loadWithStatus`

timbertson:
Thanks, I'll give that a go. Strangely, I was suddenly struck by https://github.com/dhall-lang/dhall-haskell/issues/2024 so I'm a bit stuck for now :/

timbertson:
Hmm, I like that it's simpler, but it gives worse error messages:

```
$ env DHALL_HEADERS=https://example.com ./dist/build/Dhall/dhall --file dhall-lang/tests/import/success/userHeadersA.dhall
dhall:
Import resolution is disabled
```

I'd really like for any errors to be reported with an evaluation stack containing either `env:DHALL_HEADERS` or `$HOME/.config/dhall/headers.dhall`, because if I got this error message with no further detail, I'd assume that the original import was inexplicably disallowed, rather than realising that the error is about the headers expression.

Should I give up the notion of merging this with a stack of imports, and just wrap all exceptions from this function with a new SiteHeadersLoadException or something, which explicitly tracks the source (envvar or file)?

Gabriel439:
@timbertson: What you could do is create a custom exception just for the headers file.  The idea is that instead of using `assertNoImports`, you write a function which uses the same logic except with a different exception (since it's implementation is very simple).  The custom exception could just say something like:

```
Import resolution is disallowed in the headers file located at:

↳ ${PATH of headers.dhall file}
```

It still won't have the stack of how it got there, but it's probably enough information for the user to figure out how to fix the immediate problem.

timbertson:
So it turns out disabling imports is unfortunately not what I want. The primary use case in mind that I have for this feature would use the following headers file:

```
toMap {
  , `raw.githubusercontent.com:443` = toMap {
    , Authorization = "token ${env:GITHUB_TOKEN as Text}"
    , User-Agent = "dhall"
    , Accept = "application/vnd.github.4.raw"
    }
}
```

If I'm sharing dhall expressions it's much easier to have folks set `GITHUB_TOKEN` and use this standard file, instead of making a copy with their own token inlined. But `env:GITHUB_TOKEN` is an import.

So I definitely would like to support that, does this mean I need to amend the standard? I'd be OK with it supporting arbitrary _local_ imports in the headers expression, although `env:*` is the only one I'd actually care about.

Gabriel439:
@timbertson: I would be okay with that if the transitive import behavior were behind a flag, mainly because it's technically deviating from the standard.

timbertson:
Do you mean a command line flag? How do you feel about altering the
standard to require this support instead?

Most authentication use cases seem likely to involve environment variables,
so I'd prefer to have that supported everywhere.

Gabriel439:
The main thing I want to avoid is the Haskell implementation making changes ahead of the standard, because I want to ensure that no Dhall implementation is privileged over others.

What makes this particular situation a little complicated is that if we interpret the standard as written as broadly as possible to permit transitive import resolution then it could potentially lead to infinite loops for remote imports.  That's one of the reasons I'm going in the other direction and reading the standard as narrowly as possible in order to be conservative and not implement behavior that other implementations haven't agreed to via the standardization process.

That said, there are multiple possible solutions to that which I would be fine with:

* We guard the transitive import resolution behind a command-line flag (the suggestion I just made)

* We postpone merging this feature until we've amended the standard to support transitive import resolution in `headers.dhall`

  That would mean cutting a 1.40 Haskell release that only supports standard version 21.0.0 except for `headers.dhall` support

* We merge the `headers.dhall` support without transitive import resolution before cutting Haskell release 1.40 and then add transitive import resolution in a future release when the standard is amended

* We merge your suggestion to only transitively resolve non-remote imports in `headers.dhall`

  … and declare that this was close enough to our original intent in standardizing the change and treat the upstream change to the standard as more of a clarification pull request.

timbertson:
Now I feel bad for merging the standard PR right before the 21.0.0 cut 😬 .

I'd be happiest with either options 2 or 4. I'd rather not ship this until it does what I intended. For both of those, we probably want to work on this PR and a standard update PR in parallel, and whichever is ready first determines which path we take?

> That would mean cutting a 1.40 Haskell release that only supports standard version 21.0.0 except for headers.dhall support

I'd be fine with that, but also: if the standard change is small / uncontroversial, could v1.40 simply support that version of the standard (v21.1.0 or v22.0.0)? Or do you want to release a version of dhall-haskell for each major standard version?

Gabriel439:
@timbertson: I think the easiest route would be to cut a 1.40.0 release without merging this and then cut a small 1.40.1 release right after adding support for `headers.dhall`

timbertson:
Sounds good to me 👍

timbertson:
I've raised a PR to update the standard: https://github.com/dhall-lang/dhall-lang/pull/1206

timbertson:
OK, the standard is merged, so getting back into this 🥳 

I'm still stuck with the error messaging, so if you have ideas for fixing that'd it'd be great. Alternatively, we merge it and hopefully the initial error messages are good enough because not that many people write header files without evaluating them to check that they're valid 🤞 😅

dhall/ghc-src/Dhall/Import/HTTP.hs
```diff
@@ -241,20 +259,34 @@ corsCompliant (Remote parentURL) childURL responseHeaders = liftIO $ do
...
+addHeaders :: SiteHeaders -> Maybe [HTTPHeader] -> HTTP.Request -> HTTP.Request
+addHeaders siteHeaders urlHeaders request =
+    request { HTTP.requestHeaders = (filterHeaders urlHeaders) <> originHeaders }
+      where
+        origin = decodeUtf8 (HTTP.host request) <> ":" <> Text.pack (show (HTTP.port request))
+      
+        originHeaders = HashMap.lookupDefault [] origin siteHeaders
+
+        filterHeaders Nothing = []
+        filterHeaders (Just headers) = filter (not . overridden) headers
```
Gabriel439, line 271: Minor suggestion (I don't feel too strongly about this):

```suggestion
        filterHeaders = foldMap (filter (not . overriden))
```

dhall/ghc-src/Dhall/Import/UserHeaders.hs
```diff
@@ -0,0 +1,82 @@
...
+import qualified Data.Text      as Text
+import qualified Data.Text.IO   as IO
+import qualified Dhall.Core     as Core
+
+siteHeadersFile :: FilePath -> ImportType -> Text -> SiteHeadersFile
+siteHeadersFile parentDirectory source fileContents =
+    SiteHeadersFile { parentDirectory , source, fileContents }
+
+-- lift 'siteHeadersFile' to work on IO (Maybe Text)
+siteHeadersFile' :: FilePath -> ImportType -> IO (Maybe Text) -> IO (Maybe SiteHeadersFile)
```
Gabriel439, line 36: I believe if you replace all occurrences of `IO (Maybe a)` with `MaybeT IO a` it will simplify several things you're doing here

For example, you would no longer need the `siteHeadersFile'` function because it's (basically) just `(>>=)` for `MaybeT`.  That means that `envOnlyUserHeaders` would simplify to:

```haskell
envOnlyUserHeaders :: MaybeT IO SiteHeadersFile
envOnlyUserHeaders = do
    string <- liftIO (lookupEnv key)

    siteHeadersFile "." (Core.Env (Text.pack key)) (Text.pack string)
  where
    key = "DHALL_HEADERS"
```

Similarly, `configFileOnlyUserHeaders` would simplify to:

```haskell
configFileOnlyUserHeaders = do
    directory <- liftIO (getXdgDirectory XdgConfig "dhall")

    text <- liftIO (tryReadFile (directory </> (Text.unpack filename)))

    siteHeadersFile directory (makeSource directory) text
```

… and `defaultUserHeaders` would simplify to:

```haskell
defaultUserHeaders = envOnlyUserHeaders <|> configOnlyUserHeaders
```

dhall/src/Dhall/Import.hs
```diff
@@ -786,16 +800,20 @@ fetchFresh (Env env) = do
 
 fetchFresh Missing = throwM (MissingImports [])
 
+fetchDisabledForHeaders :: URL -> StateT Status IO Data.Text.Text
```
Gabriel439, line 803: Correct me if I'm wrong, but I believe you no longer need to disable fetch for headers since we can now lean on cyclic import detection

However, that would entail a matching change to how we fetch the headers, which is to have the `loadSiteHeaders` subroutine reuse the ambient `Status` (see my comment below)

dhall/src/Dhall/Import.hs
```diff
@@ -1019,62 +1014,75 @@ getCacheBaseDirectory = alternative₀ <|> alternative₁ <|> alternative₂
...
+-- | A no-op user headers loader used for remote contexts
+--   (and loading user headers themselves)
+noopUserHeaders :: IO (Maybe SiteHeadersFile)
+noopUserHeaders = return Nothing
+
+-- | Given a SiteHeadersFile loader, return a SiteHeaders loader.
+--   The loader uses the caller's import stack, despite not using the
+--   same status (in particular, remote imports are disallowed)
+siteHeadersLoader :: IO (Maybe SiteHeadersFile) -> NonEmpty Chained -> IO SiteHeaders
+siteHeadersLoader loadSideHeadersFile importStack = loadSideHeadersFile >>= \case
```
Gabriel439, line 1036:
```suggestion
siteHeadersLoader loadSiteHeadersFile importStack = loadSiteHeadersFile >>= \case
```

dhall/ghc-src/Dhall/Import/HTTP.hs
```diff
@@ -166,6 +169,21 @@ newManager = do
         Just manager ->
             return manager
 
+getSiteHeaders :: StateT Status IO SiteHeaders
+getSiteHeaders = do
+    Status { _siteHeaders = oldSiteHeaders, _stack, ..} <- State.get
+
+    case oldSiteHeaders of
+        Nothing -> do
+            siteHeaders <- liftIO (_loadSiteHeaders _stack)
```
Gabriel439, line 178: I think you want `_loadSiteHeaders` to reuse the ambient `StateT Status` here.  Then you don't need to pass the `_stack` to it explicitly.

Also, then you wouldn't need to extend the `Status` to add a new `_loadSiteHeaders` field

timbertson: I think I still need the _loadSiteHeaders field to the status in order to support creating a status based on `envOnlyUserHeaders` (in the tests), `defaultUserHeaders` normally, and `noopUserHeaders` (for remoteStatus).

Is there another way to do that without keeping it in the status? And when you say "reuse the ambient StateT Status", do you mean to make `_loadSiteHeaders` be type `StateT Status IO SiteHeaders` (My eyes still fog over when I read monad transformer types 😅 )?

timbertson:
OK, I think this is looking much better now. Thanks for your feedback, while implementing it I realised how to make this much conceptually simpler. I've now removed the UserHeaders module entirely, the loading of user headers (from the env / file) is now just pure dhall (except for the xdgDirectory part), using ImportAlt to implement the precedence. Just like in the spec 🎉 .

Getting fallback to work was tough, I eventually figure out the fallback handling code only triggers if you have a `Note` around your `Embed`. I added a fake note, and now all the error messages are better too 😀 

One thing I'm not sure about is the **new import mode** I added: `OriginHeaders` (alongside Code / Text / Location). This feels a bit inelegant, but seemed the most reliable way to avoid tripping the "referentially opaque" error, since remote expressions should be allowed to depend on headers.

On second thoughts, I don't _think_ this should be necessary. Since I'm caching the headers, only the first remote import should actually trigger the dhall `loadWith` codepath. And I'm pretty sure the first remote import must always come from a local source, so that should mean we don't need to special-case this. I think I'm performing the load slightly too late on the stack, i.e. the headers look like they're being loaded _from_ the remote import, not as a precursor to the remote import. I'll see what I can do about this.

dhall/src/Dhall/Syntax.hs
```diff
@@ -1097,7 +1097,10 @@ instance Pretty ImportType where
     pretty Missing = "missing"
 
 -- | How to interpret the import's contents (i.e. as Dhall code or raw text)
-data ImportMode = Code | RawText | Location
+--   OriginHeaders is identical to Code, except local imports with this mode
+--   are allowed from any source (local or remote). This type is only used when
+--   loading Origin headers, it can't be set by the user.
+data ImportMode = OriginHeaders | Code | RawText | Location
```
Gabriel439, line 1103: I don't think we need a new `ImportMode` if we implement this in the same way as the standard specifies.  I'll comment elsewhere on how I think we can make this work without the `OriginHeaders` mode

dhall/ghc-src/Dhall/Import/HTTP.hs
```diff
@@ -238,20 +250,35 @@ corsCompliant (Remote parentURL) childURL responseHeaders = liftIO $ do
...
+        overridden (key, _value) = any (matchesKey key) perOriginHeaders
+
+        matchesKey :: CI ByteString -> HTTPHeader -> Bool
+        matchesKey key (candidate, _value) = key == candidate
 
 fetchFromHttpUrl :: URL -> Maybe [HTTPHeader] -> StateT Status IO Text.Text
 fetchFromHttpUrl childURL mheaders = do
+    Status { _loadOriginHeaders } <- State.get
+
+    originHeaders <- _loadOriginHeaders
```
Gabriel439, line 273: I made one mistake when I previously suggested using the ambient `StateT` for `_loadOriginHeaders`.  We do still want `_loadOriginHeaders` to use `StateT`, but we want the `_stack` to be reset to contain just one import, which is the path to origin headers.  This is because of this line from the standard:

```
(ε, headersPath) × Γ₀ ⊢ userHeadersExpr ⇒ userHeaders ⊢ Γ₁ ; Resolve userHeadersExpr with an empty import context
```

The key bit is the `(ε, headersPath)` part, which says that when resolving the `userHeadersExpr` you want the stack to contain only the headers path and nothing else.

However, we don't have to reset the stack here within the `fetchFromHttpUrl` function.  Instead, I think the best place to do that would be inside of the `_loadOriginHeaders` function.  I'll comment below with some more notes about how to do that.

dhall/src/Dhall/Import.hs
```diff
@@ -1019,62 +1005,83 @@ getCacheBaseDirectory = alternative₀ <|> alternative₁ <|> alternative₂
...
-    return url { headers = Just (fmap absurd headersExpression') }
-
-normalizeHeaders url = return url
+    loaded <- loadWith (ImportAlt partialExpr emptyOriginHeaders)
+    headers <- liftIO (toOriginHeaders loaded)
+
+    -- return cached headers next time
+    _ <- State.modify (\state -> state { _loadOriginHeaders = return headers })
+
+    return headers
```
Gabriel439, line 1065: This is where you will need to reset the stack to contain just the headers file.  However, it will require restructuring things a little bit because the type of `originHeadersLoader` doesn't permit it to know what path the `headersExpr` came from.  You might need to change it so that the `headersExpr` returns the original `Import` that the `Expr` came from so that `originHeadersLoader` can set the `stack` to that `Import`

Gabriel439 (commented):
I think this is getting pretty close.  Just one more change I can think of

dhall/src/Dhall/Import.hs
timbertson: > You might need to change it so that the headersExpr returns the original Import that the Expr came from so that originHeadersLoader can set the stack to that Import

Hmm, I think I removed my ability to do this 🤦 

`headersExpr` in the normal case is `env:DHALL_HEADERS ? ~/.config/dhall/headers.dhall`. So you can't know which branch was successful until _after_ `loadWith` is called, but you need to give `loadWith` a stack.

I've pushed an alternative solution: when loading headers, we just start with the root import. I believe this is generally either `.` (for an expression) or the path to the main file (if using --file). So this should always be local.

The trick is I had to ensure reentrant calls to load the headers didn't also reset the stack, since that's how we'd detect the `Cycle` error. Unfortunately that didn't work in practice (see [this comment](https://github.com/dhall-lang/dhall-haskell/blob/dd85fbad9edb130d32b8f3ce08279c8562d67fe3/dhall/src/Dhall/Import.hs#L1071-L1077)), but conceptually that's how it's working.

I think I could remove that hack by _also_ popping the most recent item off the stack in reentrant calls (so that it doesn't look like the headers are being loaded _from_ the remote import), but it didn't seem worth that extra complexity to achieve the same result.

Gabriel439: Actually, I think that indicates that there is a bug in the standardized behavior.  The bug is that, as you noted, resetting the stack as the standard specifies interferes with cyclic import detection.

What if we were to change the standard to not reset the stack (like you just implemented in your most recent changes) and instead change the logic to:

```
(Δ, parent, headersPath) × Γ₀ ⊢ userHeadersExpr ⇒ userHeaders ⊢ Γ₁
```

In other words, we append the headers path to the stack (not including the current URL import that the headers will be used for) when resolving the headers file.

dhall/src/Dhall/Import.hs
```diff
@@ -1019,62 +1000,105 @@ getCacheBaseDirectory = alternative₀ <|> alternative₁ <|> alternative₂
...
+originHeadersLoader :: IO (Expr Src Import) -> StateT Status IO OriginHeaders
+originHeadersLoader headersExpr = do
+
+    -- Load the headers using a parallel state with an empty impport chain.
+    -- We also set _loadOriginHeaders to prevent reentrant loads.
+
+    status <- State.get
+
+    let headerLoadStatus = status {
+        _stack = pure (NonEmpty.last (_stack status)),
```
Gabriel439, line 1059: If we change this to `_stack = headerImport :| NonEmpty.tail (_stack status)` then it will do the right thing and not trigger the referential sanity check.  Then I think you wouldn't need the `reentrantLoad` work-around.

That does, however, mean that you would need to get the `headerImport` which is currently not in scope, but I'll leave a separate comment for that.

dhall/src/Dhall/Import.hs
```diff
@@ -1019,62 +1000,105 @@ getCacheBaseDirectory = alternative₀ <|> alternative₁ <|> alternative₂
...
+
+-- | Load headers only from the environment (used in tests)
+envOriginHeaders :: Expr Src Import
+envOriginHeaders = Note headersSrc (Embed (Import (ImportHashed Nothing (Env "DHALL_HEADERS")) Code))
+
+-- | Load headers in env, falling back to config file
+defaultOriginHeaders :: IO (Expr Src Import)
+defaultOriginHeaders = do
+    fromFile <- originHeadersFileExpr
+    return (Note headersSrc (ImportAlt envOriginHeaders (Note headersSrc fromFile)))
```
Gabriel439, line 1047: I think we won't be able to use `ImportAlt` here since, as you noted, it won't let us remember which import succeeded.  i think you'd need to change the type from `IO (Expr Src Import)` to `IO [Import]` to get this to work.  It would also improve the error messages, too, I think

timbertson: I went down this path, but I couldn't see how to make it work. The fallback code is implemented by `loadWith`. I considered extracting the logic from the `ImportAlt` branch, as a function with signature `StateT Status IO (Expr Src Void) -> StateT Status IO (Expr Src Void) -> StateT Status IO (Expr Src Void)`. I couldn't just take two `Expr Src Void`, since we need to provide a different _stack in each branch (based on the header we're importing).

But it seems we need to use the full `loadWith`, since that's also the code which does the `Cycle` detection. In particular, if we manually construct `_stack = headerImport :| NonEmpty.tail (_stack status)` then that would mean we'd need to also implement the cycle detection here too. That seems messy and repetitive.

Am I missing something?

I did try an alternate, which is to just literally use the parent stack. It removed the need to special-case reentrant loads, which is nice. It requires a runtime assertion (because we can't prove the parent stack is nonempty with types). It also shouldn't ever hit a referentially opaque import error, although that's an implementation detail (because we cache on the first load). I've made that change in c77f5a7efbdd7995fa5a5a2f1816a928f18690b7

Gabriel439: @timbertson: Oh, I just figured out a (hacky) way to do this, but at least it doesn't require the partial function: instead of trying to figure out which `headers.dhall` we imported, just add both of them to the stack

timbertson: Hmm, would errors then report that e.g. ~/.config/dhall/headers.dhall was imported from `env:DHALL_HEADERS` (or vice versa)? The partial function shouldn't ever fail (despite being a bit yuck), right? 🤞

Gabriel439: Yeah, you're right.  Then we'll stick with what you have for now

Gabriel439 (approved):

timbertson:
🥳  hooray, thanks for all the assistance getting this in shape.

I'm fixing CI issues, currently stuck with this hydra error:

```
haddockPhase
Error: Incomplete haddocks
```

I'm not sure what to run to reproduce locally. I ran `cabal v1-haddock` and I get a few warnings, but nothing obvious. I thought it was referring to the <100% coverage in Import.hs, but https://github.com/dhall-lang/dhall-haskell/pull/2236/commits/dd515d9cab97f1d69c43b0205cbd44005b079410 doesn't seem to have fixed it.

Gabriel439:
CI checks for either "Missing documentation for" or "Warning: … is out of scope"

https://github.com/dhall-lang/dhall-haskell/blob/8f901d2a427fa60f58752c0e23573193a81357aa/nix/shared.nix#L84

timbertson:
Ah OK. I think I've fixed those, but now [one hydra job](https://hydra.dhall-lang.org/build/95088/nixlog/1) is dying with:

```
/nix/store/p61072rqa772qnwqhbn9xy9nbpcxw7cb-stdenv-linux/setup: line 1316: 3437713 Killed                  ./Setup build
builder for '/nix/store/q87zqrqwnizj0zpvr25jd2cjw353613l-dhall-1.40.1.drv' failed with exit code 137
```

Which I think means out of memory. Is that a known problem? I couldn't see any way to retry the job, nix is too good at caching ;)

Gabriel439:
It's a known problem.  The machine sometimes runs out of memory.  I can retry the job

Gabriel439:
@timbertson: Thank you for doing this! 🙂

timbertson:
I'm _slightly_ reluctant to ask 😉 , but does the final approach imply any necessary changes to the standard? IIRC the standard loads headers with an empty context (import chain), but what I ended up doing was to load headers with the regular context. In order to circumvent the referentially opaque check, you need to either:
 - only ever load headers on the first import (guaranteed to be local)
 - disable the referentially opaque import check when loading headers

For efficiency the first is what every implementation will want to do, but I'm not sure which is easier to standardize.

Gabriel439:
Yes, the final approach will entail a change to the standard, but I can take care of that.  It should be a small change

timbertson:
Thanks 🙂

Gabriel439:
https://github.com/dhall-lang/dhall-lang/pull/1233

PR #2249, opened 2021-07-10
Upgrade Nixpkgs

Gabriel439:
I think I finally got the upgrade working this time on Linux by incorporating the fix from https://github.com/NixOS/nixpkgs/pull/129289


Gabriel439:
Grrr.  It looks like there is yet another problem.  GHCJS appears to be abandoned and doesn't support GHC versions newer than 8.6

basile-henry:
@Gabriel439 I thought `ghcjs` moved somewhere else in `nixpkgs`. Would `pkgsCross.ghcjs` work for the `dhall-try` use case?

Gabriel439:
@basile-henry: See: https://github.com/NixOS/nixpkgs/issues/130584

PR #2278, opened 2021-08-14
Add FromDhall instances for Contravariant functors

locallycompact:


Gabriel439:
It looks like this will require enabling the `StandaloneDeriving` and `DerivingStrategies` language extensions

Gabriel439:
Hmmm, for some weird reason it looks it is complaining about missing haddocks for the re-exported types.  One way to fix this is to change this regex to be more specific:

https://github.com/dhall-lang/dhall-haskell/blob/9b43e4b0c2b4ca72e4165124496e091b283e8dcb/nix/shared.nix#L131

… but it will probably simpler to not re-export the type that is missing haddocks

Gabriel439:
@locallycompact: Thank you for contributing this! 🙂

PR #2279, opened 2021-08-14
Feature/dhall csv docs

MarcosJLR:
I created haddocks on all the exported functions and types of the libraries and I added the verification of missing haddocks in `dhall-csv` to ci in `nix/shared.nix`.

I also improved (a lot) the error messages, both on `dhall-to-csv` and `csv-to-dhall`, mostly the latter but because those messages were worse before. I'm open to any suggestion on the error messages because I found it very hard to create descriptive and helpful messages haha.

EDIT: Removed verification for CI because I can't find where are the missing Haddocks and Hydra log only reports "Incomplete Haddocks"

### DIFF
--- a/dhall-csv/src/Dhall/Csv.hs
+++ b/dhall-csv/src/Dhall/Csv.hs
@@ -70,7 +70,7 @@
 
     Also, all Dhall expressions are normalized before translation to CSV:
 
-> $ dhall-to-csv <<< "[{ equailty = True == False }]"
+> $ dhall-to-csv <<< "[{ equality = True == False }]"
 > equality
 > false
 -}
--- a/dhall-csv/src/Dhall/Csv.hs
+++ b/dhall-csv/src/Dhall/Csv.hs
@@ -142,7 +142,8 @@ instance Exception CompileError where
             \The following Dhall expression could not be translated to CSV:                  \n\
             \                                                                                \n\
             \" <> insert e <>                                                               "\n\
-            \because it has type:                                                            \n\
+            \                                                                                \n\
+            \... because it has type:                                                        \n\
             \                                                                                \n\
             \" <> insert (fromRight e (TypeCheck.typeOf e))
 
@@ -157,7 +158,8 @@ instance Exception CompileError where
             \expression:                                                                     \n\
             \                                                                                \n\
             \" <> insert e <>                                                               "\n\
-            \which has type:                                                                 \n\
+            \                                                                                \n\
+            \... which has type:                                                             \n\
             \                                                                                \n\
             \" <> insert (fromRight e (TypeCheck.typeOf e))
 
@@ -171,7 +173,8 @@ instance Exception CompileError where
             \Expected a record but instead got the following expression:                     \n\
             \                                                                                \n\
             \" <> insert e <>                                                               "\n\
-            \which has type:                                                                 \n\
+            \                                                                                \n\
+            \... which has type:                                                             \n\
             \                                                                                \n\
             \" <> insert (fromRight e (TypeCheck.typeOf e))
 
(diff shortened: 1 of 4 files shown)


sjakobi (commented):
Nice docs! Very impressive! :)

dhall-csv/src/Dhall/Csv.hs
```diff
@@ -1,12 +1,91 @@
...
+> ABC
+> $ dhall-to-csv <<< '[{ exampleText = "ABC,ABC" }]'
+> exampleText
+> "ABC,ABC"
+
+    Dhall @Optional@ values translate to the empty string if absent and the unwrapped
+    value otherwise:
+
+> $ dhall-to-csv <<< '[{ exampleOptional = None Natural }]'
+> exampleInt,exampleOptional
```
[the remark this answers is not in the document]
MarcosJLR: My bad 😅 I was going back and forth on it and forgot to fix that. Thanks! I'll fix it right away

dhall-csv/src/Dhall/Csv.hs
```diff
@@ -29,38 +108,108 @@ import qualified Dhall.TypeCheck
...
+    It contains information on the specific cases that might
+    fail to give a better insight.
+-}
+data CompileError
+    = Unsupported (Expr Void Void)
+    | NotAList (Expr Void Void)
+    | NotARecord (Expr Void Void)
+    | BareNone
 
 instance Show CompileError where
```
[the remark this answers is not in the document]
MarcosJLR: Thanks! I wasn't aware of that, I'll change it and keep it in mind for the future

dhall-csv/src/Dhall/Csv.hs
```diff
@@ -29,38 +108,108 @@ import qualified Dhall.TypeCheck
...
+            \● Unions *                                                                      \n\
             \                                                                                \n\
-            \Explanation: Only records of primitive values can be                            \n\
-            \translated from Dhall to CSV.                                                   \n\
+            \* Unions can have empty alternatives or alternatives with valid                 \n\
+            \  record field types                                                            \n\
             \                                                                                \n\
             \The following Dhall expression could not be translated to CSV:                  \n\
             \                                                                                \n\
             \" <> insert e
```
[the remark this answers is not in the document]
MarcosJLR: I have thought of that but couldn't think of a way to get the type (because all the `TypeCheck` functions return `Either` values). I just came up with a solution to use `fromRight` function and put the original expression as default (though this default should never be used, as the expression was already type checked). 

Would this solution be good?

dhall-csv/src/Dhall/Csv.hs
```diff
@@ -1,12 +1,91 @@
...
+> $ dhall-to-csv <<< "[{ exampleUnion = < Left | Right : Natural>.Left }]"
+> exampleUnion
+> Left
+> $ dhall-to-csv <<< "[{ exampleUnion = < Left | Right : Natural>.Right 2 }]"
+> exampleUnion
+> 2
+
+    Also, all Dhall expressions are normalized before translation to CSV:
+
+> $ dhall-to-csv <<< "[{ equailty = True == False }]"
```
Gabriel439, line 73:
```suggestion
> $ dhall-to-csv <<< "[{ equality = True == False }]"
```

dhall-csv/src/Dhall/Csv.hs
```diff
@@ -25,42 +104,120 @@ import qualified Dhall.Import
...
-            \Explanation: Only records of primitive values can be                            \n\
-            \translated from Dhall to CSV.                                                   \n\
+            \* Unions can have empty alternatives or alternatives with valid                 \n\
+            \  record field types                                                            \n\
             \                                                                                \n\
             \The following Dhall expression could not be translated to CSV:                  \n\
             \                                                                                \n\
-            \" <> insert e
+            \" <> insert e <>                                                               "\n\
+            \because it has type:                                                            \n\
```
Gabriel439, line 145: I usually put a blank line after the insert to separate it from the prose and also lead into the prose with `...`, like this:

```suggestion
            \" <> insert e <>                                                               "\n\
            \… because it has type:                                                          \n\
```

Same comment for the other error messages

dhall-csv/src/Dhall/Csv.hs
```diff
@@ -25,42 +104,120 @@ import qualified Dhall.Import
...
+            \" <> insert e <>                                                               "\n\
+            \which has type:                                                                 \n\
+            \                                                                                \n\
+            \" <> insert (fromRight e (TypeCheck.typeOf e))
+
+    displayException BareNone =
+       Data.Text.unpack $
+            _ERROR <> ": ❰None❱ is not valid on its own                                      \n\
+            \                                                                                \n\
+            \Explanation: The conversion to JSON/YAML does not accept ❰None❱ in isolation as \n\
```
Gabriel439, line 182: I think this should say CSV instead of JSON/YAML

Gabriel439 (approved):
Looks great!  Just a few small comments

dhall-csv/src/Dhall/Csv.hs
MarcosJLR: Oops hahaha the dangers of copying and pasting. I'll change it right away

PR #2282, opened 2021-08-19
toml: fix bug converting nested union values

ear7h:
Lists of records should be represented as TOML tables, the logic for
this was only checking for `RecordLit`s. Howver, unions can also be
translated to TOML tables. This commit adds a check for this case.


Gabriel439 (approved):
Nice use of `PatternSynonyms`!

PR #2283, opened 2021-08-20
nix: include `cabal-install` in all shells

timbertson:
Discussed in https://github.com/dhall-lang/dhall-haskell/issues/2024#issuecomment-899883432

I noticed we already had a `toShell`, so I added it there and used it for all the `dhall-shell`, `dhall-json-shell`, etc exports, which are used in individual `shell.nix` files. I can move it to a different function if we don't want it in `toShell`, I'm not sure who uses that.

### DIFF
--- a/nix/shared.nix
+++ b/nix/shared.nix
@@ -444,7 +444,9 @@ let
       '';
     };
 
-  toShell = drv: drv.env;
+  toShell = drv: drv.env.overrideAttrs (base: {
+    buildInputs = (base.buildInputs or []) ++ [ pkgs.cabal-install ];
+  });
 
   possibly-static = {
     dhall            = makeStaticIfPossible "dhall"           ;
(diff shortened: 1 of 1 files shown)


nix/shared.nix
```diff
@@ -444,7 +444,9 @@ let
       '';
     };
 
-  toShell = drv: drv.env;
+  toShell = drv: drv.env.overrideAttrs (base: {
+    buildInputs = (base.buildInputs or []) ++ [ pkgs.cabal-install ];
```
Gabriel439, line 448:
```suggestion
    nativeBuildInputs = (base.nativeBuildInputs or []) ++ [ pkgs.cabal-install ];
```

This is a minor thing since the `env` derivation isn't actually built, but `nativeBuildInputs` are intended for things that are required only at build time, whereas `buildInputs` is for dependencies that are used at runtime.  The Nixpkgs manual has a longer explanation of this here:

https://nixos.org/manual/nixpkgs/stable/#variables-specifying-dependencies

Gabriel439 (approved):
Just one small comment, but this looks great!

PR #2291, opened 2021-08-23
Fix equivalence check for `Date`/`Time`/`TimeZone`

Gabriel439:
This is fixing a bug where a temporal literal fails to type-check
when given a type annotation (or more generally, when the type-checker
checks it against an expected type).

For example, the expression `00:00:00 : Date` would not type-check.

The reason why is that `Dhall.Eval.conv` was missing cases for
the `Date` / `Time` / `TimeZone` types and their respective literals,
which this change fixes.

I also fixed the test suite to catch future issues like this.  We do
have standard tests that check that temporal literals against
expected types, but we were running the tests in such a way that we
were not exercising the `Dhall.Eval.conv` code path that had this
bug.

So I updated the test suite so that it now catches this issue and
future issues like this.  I verified that the updated test suite
now fails without this fix and passes with the fix.

### DIFF
--- a/dhall/src/Dhall/Eval.hs
+++ b/dhall/src/Dhall/Eval.hs
@@ -981,6 +981,18 @@ conv !env t0 t0' =
             conv env t t'
         (VTextReplace a b c, VTextReplace a' b' c') ->
             conv env a a' && conv env b b' && conv env c c'
+        (VDate, VDate) ->
+            True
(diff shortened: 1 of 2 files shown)


Gabriel439:
I also plan to cut a small `dhall-1.40.1` release with this fix when it's merged

sjakobi (approved):

PR #2292, opened 2021-08-23
Add `dhall-{csv,toml}` to the release script

Gabriel439:


sjakobi (approved):

ELLIOTTCABLE:
Hrm. Despite this, the GitHub releases don't seem to have binaries for `dhall-toml`; and Homebrew is missing it as well.

Is something bunked up in the release process? :x

Gabriella439:
The last step of attaching the release artifacts is manual, so it's quite likely I just plain missed adding them to the release.  I'll look into this.

Gabriella439:
Also, I think the missing `dhall-toml` from the release page is unrelated to it being missing from `brew`.  `brew` builds and cache Haskell packages like the `dhall*` ones from source and doesn't use the release page.  It's probably more likely that the reason `brew` is missing `dhall-toml` is that it's one of the newer `dhall*` executables and nobody had gotten around to packaging it for `brew`, yet.

Gabriella439:
So it looks like there are `dhall-toml` binaries for the latest release:

Screenshot 2023-04-14 at 9 29 49 PM

However, there hidden behind a "Show all 30 assets" link that you have to click through.

ELLIOTTCABLE:
Okay, I've PR'd a `dhall-toml` formula on Homebrew. Build works fine for me, here's hoping it's accepted, since they've already got a couple other Dhall sub-formula. (=

Homebrew/homebrew-core#128312

Issue #2286, opened 2021-08-21
`FromDhall`/`ToDhall` for types from `time`?


amesgen:
I am not sure whether this has been discussed before, but what about implementing `FromDhall`/`ToDhall` for various types from `time`, like `Day`, `TimeOfDay`, `LocalTime`, `ZonedTime`, `UTCTime`, now with #2247?

Gabriel439:
Whoops, this is an oversight on my part!  Yeah, this is a great idea and it should be easy for me to add

Gabriel439:
Fix is up here: https://github.com/dhall-lang/dhall-haskell/pull/2294

amesgen:
Thanks for the instances for `Day`, `TimeOfDay` and `TimeZone`! I created #2300 to also add instances for `LocalTime`/`ZonedTime`/`UTCTime` (if that is desired).

georgefst:
Would you consider also adding `DayOfWeek` (as the obvious 7-element enum)? I've found this useful in some projects.

Gabriel439:
@georgefst: I opened a PR to add isntance for `DayOfWeek`: https://github.com/dhall-lang/dhall-haskell/pull/2413

PR #2300, opened 2021-09-11
Add `FromDhall`/`ToDhall` for `LocalTime`/`ZonedTime`/`UTCTime`

amesgen:
Follow up to #2286.

One thing to discuss is how to handle decoding `UTCTime`. Right now, any input timezone is accepted, and then converted to UTC. One could also argue that only the `00:00` input timezone should be accepted, with an error thrown otherwise. I would be fine with both approaches.


Gabriel439:
@amesgen: I think what you have now is fine, meaning that I think it's good for `utctime` to convert any timezone to UTC

Gabriel439:
Thank you! 😄

PR #2301, opened 2021-09-14
Upgrade Nixpkgs

Gabriel439:
I finally got this working on Linux

The main downside is that GHCJS is not supported yet in newer versions of Nixpkgs:

* https://github.com/NixOS/nixpkgs/issues/133271

… but that only means that we won't be able to update the dhall-lang.org demo to showcase new language features, which isn't too bad

The motivation for this is so that we can finally support development on Big Sur for macOS


sjakobi (approved):

Gabriel439:
I also verified that this works on Big Sur, too

PR #2303, opened 2021-09-18
Clarify what referentially opaque means, for the uninitiated

timbertson:
Just a small addition to the referentially opaque error message. I still have do a double-take to remember what this phrase means, so I thought I'd help myself and others ;)

### DIFF
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -286,7 +286,7 @@ instance Exception ReferentiallyOpaque
 
 instance Show ReferentiallyOpaque where
     show (ReferentiallyOpaque import_) =
-        "\nReferentially opaque import: " ++ Dhall.Pretty.Internal.prettyToString import_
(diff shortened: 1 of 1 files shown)


dhall/src/Dhall/Import.hs
```diff
@@ -286,7 +286,7 @@ instance Exception ReferentiallyOpaque
 
 instance Show ReferentiallyOpaque where
     show (ReferentiallyOpaque import_) =
-        "\nReferentially opaque import: " ++ Dhall.Pretty.Internal.prettyToString import_
+        "\nReferentially opaque import (local imports are not permitted from remote sources): " ++ Dhall.Pretty.Internal.prettyToString import_
```
Gabriel439, line 289: We can probably get rid of the term "referentially opaque" from the error:

```suggestion
        "\nLocal imports are not permitted from remote imports: " ++ Dhall.Pretty.Internal.prettyToString import_
```

Gabriel439 (approved):

PR #2304, opened 2021-09-20
change dhall-nix to generate FOD for URL imports

cdepillabout:
I'd like to be able to sue `dhall-nix` to generate Nix code from a derivation, and then recursively import it (IFD).  The problem with this is that remote URL imports in Dhall can't be resolved in a Nix derivation unless you make it a fixed-output derivation.

Luckily, if remote URL imports in Dhall have an integrity check, then we can make them into fixed-output derivation.

Currently, this PR modifies `dhall-nix` to generate Nix code that turns remote imports with an integrity check into fixed output derivations.  It then recursively imports them.

For example, imagine the following Dhall file:

`example-remote-import.dhall`:

```dhall
[   1
  + https://gist.githubusercontent.com/cdepillabout/2683131f078753fd24723ab8bf1e1b74/raw/7e1b26ff812000e868a4ff108b33a21d03a9a591/example-normal.dhall
      sha256:15f52ecf91c94c1baac02d5a4964b2ed8fa401641a2c8a95e8306ec7c1e3b8d2
, 3
]
```

If you look at https://gist.githubusercontent.com/cdepillabout/2683131f078753fd24723ab8bf1e1b74/raw/7e1b26ff812000e868a4ff108b33a21d03a9a591/example-normal.dhall, it is just:

```dhall
List/length Natural [ 100, 200, 300 ]
```

Running `nix-dhall` from this PR on `example-remote-import.dhall` will produce an output like the following:

```nix
{ dhallToNixUrlFOD, cacert, dhall, fetchurl, runCommand }:
  [
    (1 + dhallToNixUrlFOD (let
      packageSet = (fetchurl {
        url = "https://gist.githubusercontent.com/cdepillabout/2683131f078753fd24723ab8bf1e1b74/raw/7e1b26ff812000e868a4ff108b33a21d03a9a591/example-normal.dhall";
        hash = "sha256-FfUuz5HJTBuqwC1aSWSy7Y+kAWQaLIqV6DBux8HjuNI=";
        downloadToTemp = true;
        postFetch = "${dhall}/bin/dhall --alpha --plain --file \"\$downloadedFile\" | ${dhall}/bin/dhall encode > \$out";
        }).overrideAttrs (oldAttrs:
        {
          nativeBuildInputs = (oldAttrs.nativeBuildInputs or []) ++ [ cacert ];
          });
    in runCommand "decodedPackageSet" {} "${dhall}/bin/dhall decode --file \"${packageSet}\" > \$out") {
      inherit dhallToNixUrlFOD cacert dhall fetchurl runCommand;
      })
    3
    ]
```

There are two new things here:

1. With the current code in this PR, `dhall-nix` now produces a function that takes the following arguments:

    ```nix
    { dhallToNixUrlFOD, cacert, dhall, fetchurl, runCommand }:
    ```

    These values are used in the code for recursively importing remote URLs from Dhall.

1. Remote URLs with an integrity check in Dhall are generated into an expression like the following:

    ```nix
    dhallToNixUrlFOD
      ( let
          packageSet =
            (fetchurl {
              url = "https://gist.githubusercontent.com/cdepillabout/2683131f078753fd24723ab8bf1e1b74/raw/7e1b26ff812000e868a4ff108b33a21d03a9a591/example-normal.dhall";
              hash = "sha256-FfUuz5HJTBuqwC1aSWSy7Y+kAWQaLIqV6DBux8HjuNI=";
            downloadToTemp = true;
              postFetch = "${dhall}/bin/dhall --alpha --plain --file \"\$downloadedFile\" | ${dhall}/bin/dhall encode > \$out";
            }).overrideAttrs (oldAttrs: {
              nativeBuildInputs = (oldAttrs.nativeBuildInputs or []) ++ [ cacert ];
            });
        in
        runCommand "decodedPackageSet" {} "${dhall}/bin/dhall decode --file \"${packageSet}\" > \$out"
      )
      { inherit dhallToNixUrlFOD cacert dhall fetchurl runCommand; }
    ```

    The code in this PR changes Dhall integrity checks into a hash that Nix understands, and throws it into a call to `fetchurl`.  It normalizes the result and encodes it so the hash matches.  It then decodes it.

    This derivation is then passed into a function I'm calling `dhallToNixUrlFOD`.

    You can find a proof-of-concept definition of `dhallToNixUrlFOD` in the file `example.nix` in this PR.  It is basically the same as [`dhallToNix` in Nixpkgs](https://github.com/NixOS/nixpkgs/blob/9a8850aea97840dca1cc1087447ed89dbfe183dc/pkgs/build-support/dhall-to-nix.nix).

### DIFF
--- a/dhall-nix/exec/Main.hs
+++ b/dhall-nix/exec/Main.hs
@@ -48,14 +48,20 @@ main = handle (Dhall.detailed (do
         Left  err  -> Control.Exception.throwIO err
         Right expr -> return expr
 
-    expr' <- Dhall.Import.load expr
-    case Dhall.TypeCheck.typeOf expr' of
-        Left  err -> Control.Exception.throwIO err
-        Right _   -> return ()
+    -- expr' <- Dhall.Import.load expr
+    -- case Dhall.TypeCheck.typeOf expr' of
+    --     Left  err -> Control.Exception.throwIO err
+    --     Right _   -> return ()
 
-    nix <- case Dhall.Nix.dhallToNix expr' of
+    -- nix <- case Dhall.Nix.dhallToNix expr' of
+    --     Left err  -> Control.Exception.throwIO err
+    --     Right nix -> return nix
+
+    nix <- case Dhall.Nix.dhallToNix expr of
         Left err  -> Control.Exception.throwIO err
         Right nix -> return nix
+
+
     print (Nix.Pretty.prettyNix nix) ))
 
 handle :: IO a -> IO a
--- a/dhall-nix/src/Dhall/Nix.hs
+++ b/dhall-nix/src/Dhall/Nix.hs
@@ -1,10 +1,11 @@
-{-# LANGUAGE DeriveDataTypeable #-}
-{-# LANGUAGE OverloadedLists    #-}
-{-# LANGUAGE OverloadedStrings  #-}
-{-# LANGUAGE QuasiQuotes        #-}
-{-# LANGUAGE RecordWildCards    #-}
-{-# LANGUAGE TypeFamilies       #-}
-{-# LANGUAGE ViewPatterns       #-}
+{-# LANGUAGE DeriveDataTypeable  #-}
+{-# LANGUAGE OverloadedLists     #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE QuasiQuotes         #-}
+{-# LANGUAGE RecordWildCards     #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies        #-}
+{-# LANGUAGE ViewPatterns        #-}
 
 {-| This library only exports a single `dhallToNix` function for translating a
     Dhall syntax tree to a Nix syntax tree for the @hnix@ library
@@ -105,6 +106,10 @@ import Dhall.Core
     , DhallDouble (..)
     , Expr (..)
     , FunctionBinding (..)
+    , Import(..)
+    , ImportHashed(..)
+    , ImportMode(..)
+    , ImportType(..)
     , MultiLet (..)
     , PreferAnnotation (..)
     , Var (..)
@@ -124,6 +129,12 @@ import Nix.Expr
     , (==>)
     , (@.)
     , (@@)
+    )
+import Nix.Expr.Shorthands
+    ( attrsE
+    , letE
+    , mkSym
+    , mkStr
     )
 
 import qualified Data.Text
@@ -222,10 +233,11 @@ Right x: y: x + y
     Precondition: You must first type-check the Dhall expression before passing
     the expression to `dhallToNix`
 -}
-dhallToNix :: Expr s Void -> Either CompileError (Fix NExprF)
+dhallToNix :: forall s. Expr s Dhall.Core.Import -> Either CompileError (Fix NExprF)
 dhallToNix e =
     loop (rewriteShadowed (Dhall.Core.normalize e))
   where
+    untranslatable :: Fix NExprF
     untranslatable = Fix (NSet NNonRecursive [])
 
     -- This is an intermediate utility used to remove all occurrences of
@@ -239,7 +251,7 @@ dhallToNix e =
     -- If `Nothing` then the current bound variable doesn't need to be renamed.
     -- If any other number, then rename the variable to include the maximum
     -- depth.
-    maximumDepth :: Var -> Expr s Void -> Maybe Int
+    maximumDepth :: forall a. Var -> Expr s a -> Maybe Int
     maximumDepth v@(V x n) (Lam _ FunctionBinding {functionBindingVariable = x', functionBindingAnnotation = a} b)
         | x == x' =
             max (maximumDepth v a) (fmap (+ 1) (maximumDepth (V x (n + 1)) b))
@@ -259,7 +271,7 @@ dhallToNix e =
 
     -- Higher-level utility that builds on top of `maximumDepth` to rename a
     -- variable if there are shadowed references to that variable
-    rename :: (Text, Expr s Void) -> Maybe (Text, Expr s Void)
+    rename :: forall a. (Text, Expr s a) -> Maybe (Text, Expr s a)
     rename (x, expression) =
         case maximumDepth (V x 0) expression of
             Nothing ->
@@ -274,7 +286,7 @@ dhallToNix e =
               where
                 x' = x <> Data.Text.pack (show n)
 
-    renameShadowed :: Expr s Void -> Maybe (Expr s Void)
+    renameShadowed :: forall a. Expr s a -> Maybe (Expr s a)
     renameShadowed (Lam cs FunctionBinding { functionBindingVariable = x, functionBindingAnnotation = a} b) = do
         (x', b') <- rename (x, b)
 
@@ -294,6 +306,7 @@ dhallToNix e =
     rewriteShadowed =
         Dhall.Optics.rewriteOf Dhall.Core.subExpressions renameShadowed
 
+    loop :: Expr s Dhall.Core.Import -> Either CompileError (Fix NExprF)
     loop (Const _) = return untranslatable
     loop (Var (V a 0)) = return (Fix (NSym a))
     loop (Var  a     ) = Left (CannotReferenceShadowedVariable a)
@@ -660,4 +673,62 @@ dhallToNix e =
         loop (Dhall.Core.desugarWith a)
     loop (ImportAlt a _) = loop a
     loop (Note _ b) = loop b
-    loop (Embed x) = absurd x
+    loop (Embed (Import (ImportHashed maybeHash importType) mode)) =
+      case importType of
+        Local{} -> return untranslatable
+        Env{} -> return untranslatable
+        Missing -> return untranslatable
+        Remote url ->
+          case mode of
+            Location -> return untranslatable
+            RawText -> return untranslatable
+            Code ->
+              case maybeHash of
+                Nothing -> return untranslatable
+                Just hash ->
+                  --     dhallToNixUrlFOD (
+                  --       let
+                  --         packageSet = fetchurl {
+                  --           url = "https://github.com/purescript/package-sets/releases/download/psc-0.14.4-20210905/packages.dhall";
+                  --           hash = "sha256-FA82MIAfKwLV86QF1IcuCvMX5O8YcBamsA+X1Z1idcY=";
+                  --           downloadToTemp = true;
+                  --           postFetch = ''
+                  --             ${dhall}/bin/dhall encode --file "$downloadedFile" > $out
+                  --           '';
+                  --         };
+                  --       in
+                  --         runCommand "decodedPackageSet" {} ''
+                  --           ${dhall}/bin/dhall decode --file "${packageSet}" > $out
+                  --         '';
+                  --     )
+                  return $
+                    Fix
+                      (NBinary
+                        NApp
+                        (Fix (NSym "dhallToNixUrlFOD"))
+                        (letE
+                          "packageSet"
+                          ("fetchurl" @@
+                            attrsE
+                              [ ("url", "")
+                              , ("hash", "")
+                              , ("downloadToTemp", "")
+                              , ("postFetch", "")
+                              ]
+                          )
+                          ("runCommand" @@ mkStr "decodedPackageSet" @@
+                            attrsE [] @@
+                            (Fix
+                              (NStr
+                                (DoubleQuoted
+                                  [ Antiquoted "dhall"
+                                  , Plain "/bin/dhall decode --file \""
+                                  , Antiquoted "packageSet"
+                                  , Plain "\" > $out"
+                                  ]
+                                )
+                              )
+                            )
+                          )
+                        )
+                      )
--- a/dhall-nix/src/Dhall/Nix.hs
+++ b/dhall-nix/src/Dhall/Nix.hs
@@ -739,7 +739,9 @@ dhallToNix e =
                             , ("postFetch",
                                 mkStrAntiquote
                                   [ Antiquoted "dhall"
-                                  , Plain "/bin/dhall encoded --file \"$downloadedFile\" > $out"
+                                  , Plain "/bin/dhall normalize --alpha --file \"$downloadedFile\" | "
+                                  , Antiquoted "dhall"
+                                  , Plain "/bin/dhall encode > $out"
                                   ]
                               )
                             ]
--- a/dhall-nix/src/Dhall/Nix.hs
+++ b/dhall-nix/src/Dhall/Nix.hs
@@ -130,6 +130,7 @@ import Nix.Expr
     , NString (..)
     , Params (..)
     , ($+)
+    , ($++)
     , (==>)
     , (@.)
     , (@@)
@@ -140,10 +141,12 @@ import Nix.Expr.Shorthands
     , letE
     , mkBool
     , mkFunction
+    , mkList
     , mkNonRecSet
     , mkParamset
-    , mkSym
+    , mkSelector
     , mkStr
+    , mkSym
     )
 
 import qualified Data.ByteString.Char8 as ByteString.Char8
@@ -255,9 +258,10 @@ dhallToNix e =
           mkFunction
             (mkParamset
               [ ("dhallToNixUrlFOD", Nothing)
+              , ("cacert", Nothing)
+              , ("dhall", Nothing)
               , ("fetchurl", Nothing)
               , ("runCommand", Nothing)
-              , ("dhall", Nothing)
               ]
               False
             )
@@ -712,39 +716,60 @@ dhallToNix e =
               case maybeHash of
                 Nothing -> return untranslatable
                 Just hash ->
-                  --     dhallToNixUrlFOD (
-                  --       let
-                  --         packageSet = fetchurl {
+                  -- The following code block creates a function call to
+                  -- dhallToNixUrlFOD that looks like the following:
+                  --
+                  -- dhallToNixUrlFOD
+                  --   (let
+                  --       packageSet =
+                  --         (fetchurl {
                   --           url = "https://github.com/purescript/package-sets/releases/download/psc-0.14.4-20210905/packages.dhall";
-                  --           hash = "sha256-FA82MIAfKwLV86QF1IcuCvMX5O8YcBamsA+X1Z1idcY=";
+                  --           hash = "sha256-W1WfXQXIvufW57F5N7fxqRvi/YMBdvlnwlFn4eMyVLc=";
                   --           downloadToTemp = true;
-                  --           postFetch = ''
-                  --             ${dhall}/bin/dhall encode --file "$downloadedFile" > $out
-                  --           '';
-                  --         };
-                  --       in
-                  --         runCommand "decodedPackageSet" {} ''
-                  --           ${dhall}/bin/dhall decode --file "${packageSet}" > $out
-                  --         '';
-                  --     )
+                  --           postFetch = "${dhall}/bin/dhall --alpha --plain --file \"\$downloadedFile\" | ${dhall}/bin/dhall encode > \$out";
+                  --         }).overrideAttrs (oldAttrs: {
+                  --           nativeBuildInputs = (oldAttrs.nativeBuildInputs or []) ++ [ cacert ];
+                  --         });
+                  --    in
+                  --      runCommand "decodedPackageSet" {} "${dhall}/bin/dhall decode --file \"${packageSet}\" > \$out"
+                  --   )
+                  --   { inherit dhallToNixUrlFOD cacert dhall fetchurl runCommand; };
                   return $
                     "dhallToNixUrlFOD" @@
                       (letE
                         "packageSet"
-                        ("fetchurl" @@
-                          attrsE
-                            [ ("url", mkStr $ Dhall.Core.pretty url)
-                            , ("hash", mkStr $ dhallHashToSriHash hash)
-                            , ("downloadToTemp", mkBool True)
-                            , ("postFetch",
-                                mkStrAntiquote
-                                  [ Antiquoted "dhall"
-                                  , Plain "/bin/dhall --alpha --plain --file \"$downloadedFile\" | "
-                                  , Antiquoted "dhall"
-                                  , Plain "/bin/dhall encode > $out"
-                                  ]
-                              )
-                            ]
+                        (
+                          ("fetchurl" @@
+                            attrsE
+                              [ ("url", mkStr $ Dhall.Core.pretty url)
+                              , ("hash", mkStr $ dhallHashToSriHash hash)
+                              , ("downloadToTemp", mkBool True)
+                              , ("postFetch",
+                                  mkStrAntiquote
+                                    [ Antiquoted "dhall"
+                                    , Plain "/bin/dhall --alpha --plain --file \"$downloadedFile\" | "
+                                    , Antiquoted "dhall"
+                                    , Plain "/bin/dhall encode > $out"
+                                    ]
+                                )
+                              ]
+                          ) @.
+                          "overrideAttrs" @@
+                          mkFunction
+                            "oldAttrs"
+                            (attrsE
+                              [ ("nativeBuildInputs",
+                                  (Fix
+                                    (NSelect
+                                      "oldAttrs"
+                                      (mkSelector "nativeBuildInputs")
+                                      (Just $ mkList [])
+                                    )
+                                  ) $++
+                                  mkList [ "cacert" ]
+                                )
+                              ]
+                            )
                         )
                         ("runCommand" @@ mkStr "decodedPackageSet" @@
                           attrsE [] @@
@@ -759,9 +784,10 @@ dhallToNix e =
                       ( mkNonRecSet
                           [ inherit
                               [ "dhallToNixUrlFOD"
+                              , "cacert"
+                              , "dhall"
                               , "fetchurl"
                               , "runCommand"
-                              , "dhall"
                               ]
                               (Nix.SourcePos "unknown" (Nix.mkPos 1) (Nix.mkPos 1))
                           ]
--- a/dhall-nix/src/Dhall/Nix.hs
+++ b/dhall-nix/src/Dhall/Nix.hs
@@ -133,6 +133,7 @@ import Nix.Expr
 import Nix.Expr.Shorthands
     ( attrsE
     , letE
+    , mkBool
     , mkSym
     , mkStr
     )
@@ -702,33 +703,32 @@ dhallToNix e =
                   --         '';
                   --     )
                   return $
-                    Fix
-                      (NBinary
-                        NApp
-                        (Fix (NSym "dhallToNixUrlFOD"))
-                        (letE
-                          "packageSet"
-                          ("fetchurl" @@
-                            attrsE
-                              [ ("url", "")
-                              , ("hash", "")
-                              , ("downloadToTemp", "")
-                              , ("postFetch", "")
-                              ]
-                          )
-                          ("runCommand" @@ mkStr "decodedPackageSet" @@
-                            attrsE [] @@
-                            (Fix
-                              (NStr
-                                (DoubleQuoted
+                    "dhallToNixUrlFOD" @@
+                      (letE
+                        "packageSet"
+                        ("fetchurl" @@
+                          attrsE
+                            [ ("url", mkStr $ Dhall.Core.pretty url)
+                            , ("hash", "")
+                            , ("downloadToTemp", mkBool True)
+                            , ("postFetch",
+                                mkStrAntiquote
                                   [ Antiquoted "dhall"
-                                  , Plain "/bin/dhall decode --file \""
-                                  , Antiquoted "packageSet"
-                                  , Plain "\" > $out"
+                                  , Plain "/bin/dhall encoded --file \"$downloadedFile\" > $out"
                                   ]
-                                )
                               )
-                            )
-                          )
+                            ]
+                        )
+                        ("runCommand" @@ mkStr "decodedPackageSet" @@
+                          attrsE [] @@
+                          mkStrAntiquote
+                            [ Antiquoted "dhall"
+                            , Plain "/bin/dhall decode --file \""
+                            , Antiquoted "packageSet"
+                            , Plain "\" > $out"
+                            ]
                         )
                       )
+
+mkStrAntiquote :: [ Antiquoted Text (Fix NExprF) ] -> Fix NExprF
+mkStrAntiquote strs = Fix (NStr (DoubleQuoted strs))
--- a/dhall-nix/src/Dhall/Nix.hs
+++ b/dhall-nix/src/Dhall/Nix.hs
@@ -739,7 +739,7 @@ dhallToNix e =
                             , ("postFetch",
                                 mkStrAntiquote
                                   [ Antiquoted "dhall"
-                                  , Plain "/bin/dhall normalize --alpha --file \"$downloadedFile\" | "
+                                  , Plain "/bin/dhall --alpha --plain --file \"$downloadedFile\" | "
                                   , Antiquoted "dhall"
                                   , Plain "/bin/dhall encode > $out"
                                   ]
--- a/dhall-nix/src/Dhall/Nix.hs
+++ b/dhall-nix/src/Dhall/Nix.hs
@@ -94,6 +94,10 @@ module Dhall.Nix (
     ) where
 
 import Control.Exception (Exception)
+import Data.ByteArray.Encoding
+    ( Base (Base64)
+    , convertToBase
+    )
 import Data.Fix          (Fix (..))
 import Data.Foldable     (toList)
 import Data.Text         (Text)
@@ -134,12 +138,16 @@ import Nix.Expr.Shorthands
     ( attrsE
     , letE
     , mkBool
+    , mkFunction
+    , mkParamset
     , mkSym
     , mkStr
     )
 
+import qualified Data.ByteString.Char8 as ByteString.Char8
 import qualified Data.Text
 import qualified Dhall.Core
+import qualified Dhall.Crypto
 import qualified Dhall.Map
 import qualified Dhall.Optics
 import qualified Dhall.Pretty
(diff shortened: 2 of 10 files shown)


cdepillabout:
My big question is whether something like this would be accepted in the `dhall-nix` library/executable?  My guess is that this would not be accepted, because it is quite a big change.  It also needs to be paired with the new `dhallToNixUrlFOD` function, which would need to come from somewhere (Nixpkgs?).

-   If something like this would be accepted, then the current code in this PR needs a lot of cleanup.  We'd also have to decide how to expose this new functionality in `dhall-nix`.  Would we give `dhall-nix` a new CLI flag like `--url-fod` that uses this alternative output format?  Or possibly split this functionality into a separate executable?

-   If something like this would not be accepted, then I'll just maintain it in a separate repo.  This won't be a problem for me.

    However, in this case, it would be nice if we could generalize `Dhall.Nix.dhallToNix` from:

    ```haskell
    dhallToNix :: Expr s Void -> Either CompileError (Fix NExprF)
    ```

    To something like:

    ```haskell
    dhallToNix :: (a -> Fix NExprF) -> Expr s a -> Either CompileError (Fix NExprF)
    ```

    This new argument would define how to handle Dhall `Import`s

    Or potentially just add a new helper function with the same signature if you're worried about changing the signature of `dhallToNix`, which is exported and other packages might be relying on.  I'm not sure how widely the `dhall-nix` library is used in the wild.  (At the very least, it doesn't appear to have any reverse-dependencies on Hackage.)

This is my first time really working with Dhall, and I had a couple of other questions about various thing.  I'll leave them inline in comments below.

dhall-nix/src/Dhall/Nix.hs
```diff
@@ -660,4 +703,100 @@ dhallToNix e =
...
+                              [ ("url", mkStr $ Dhall.Core.pretty url)
+                              , ("hash", mkStr $ dhallHashToSriHash hash)
+                              , ("downloadToTemp", mkBool True)
+                              , ("postFetch",
+                                  mkStrAntiquote
+                                    [ Antiquoted "dhall"
+                                    , Plain "/bin/dhall --alpha --plain --file \"$downloadedFile\" | "
+                                    , Antiquoted "dhall"
+                                    , Plain "/bin/dhall encode > $out"
+                                    ]
```
cdepillabout, line 753: In order to make recursive URL imports work correctly, I setup my `dhallToNixUrlFOD` function to call itself recursively if it encounters a URL import.

Also, in order to make Dhall's hash line-up with Nix's hash, I run it through `dhall --alpha`.

My question is *if I run an expression through `dhall --alpha`, am I guaranteed that all remote imports will be resolved and inlined*?  

If this is true, then I guess I really don't need to worry about calling `dhallToNixUrlFOD` recursively, since there should never be any remote URL imports that need to evaluated recursively.  They should all be resolved during the first call to `dhall --alpha`.

dhall-nix/exec/Main.hs
```diff
@@ -48,14 +48,20 @@ main = handle (Dhall.detailed (do
         Left  err  -> Control.Exception.throwIO err
         Right expr -> return expr
 
-    expr' <- Dhall.Import.load expr
-    case Dhall.TypeCheck.typeOf expr' of
-        Left  err -> Control.Exception.throwIO err
-        Right _   -> return ()
+    -- expr' <- Dhall.Import.load expr
```
cdepillabout, line 51: I commented this out since I need to do remote URL imports within Nix.

However, it would be nice if I could use some function in `Dhall.Import` to only do local imports, but leave all remote URL imports.  I poked around in `Dhall.Import` to see if this was possible, but I couldn't really come up with anything.

Is there some easy way to do this?  Or should I just handle local imports in Nix as well?

dhall-nix/exec/Main.hs
```diff
@@ -48,14 +48,20 @@ main = handle (Dhall.detailed (do
...
+    -- case Dhall.TypeCheck.typeOf expr' of
+    --     Left  err -> Control.Exception.throwIO err
+    --     Right _   -> return ()
```
cdepillabout, line 54: It is unfortunate that I had to comment out the type-checking here.  I couldn't find any functions in `Dhall.TypeCheck` that would allow type-checking on Dhall expressions that still have imports (which makes sense).

Maybe it would make sense to change the implementation in this PR to instead be more like `dhall-nixpkgs`?  I could use the similar logic to `dhall-nixpkgs` to search through the Dhall expression tree for remote imports, and use FODs to load them into a cache.  I could then make use of this cache when evaluating my main Dhall file, similar to how `buildDhallDirectoryPackage` works.

I'm guessing this would also let me type-check the input Dhall file, since all of its remote url imports will be in the cache.

example-purescript-package/spago2.dhall
```diff
@@ -0,0 +1,6 @@
+{ name = "my-project"
+, dependencies = [ "console", "effect", "prelude", "psci-support" ]
+, packages = https://github.com/purescript/package-sets/releases/download/psc-0.14.4-20210905/packages.dhall sha256:140f3630801f2b02d5f3a405d4872e0af317e4ef187016a6b00f97d59d6275c6
+, sources = [ "src/**/*.purs", "test/**/*.purs" ]
+}
```
cdepillabout, line 5: This file is an example of my motivation for writing this PR.

I'd like to be able to package PureScript packages in Nix.  The user should only need this `spago.dhall` (and the accompanying `packages.dhall` file, but I've inlined `packages.dhall` in this `spago2.dhall` file).

In PureScript, the build tool `spago` is setup to always add an integrity check to the package set import, so in theory this should be possible to build with a FOD nix derivation.  This PR implements the FOD fetcher.  I still have to put together an example of actually building a PureScript application with this.

Gabriel439:
Are you familiar with how the current Nixpkgs support for Dhall works and specifically how `dhall-nixpkgs` works?

* [Nixpkgs - Dhall](https://nixos.org/manual/nixpkgs/stable/#sec-language-dhall)

The reason I ask is because I believe the `dhall-nixpkgs` tool already does something close to what you want and with a few tweaks it can probably do exactly what you're requesting.  For example, I just put up a change for gist support in https://github.com/dhall-lang/dhall-haskell/pull/2306 and with that change `dhall-nixpkgs` now produces this result for your example:

```bash
$ dhall-to-nixpkgs github https://gist.github.com/cdepillabout/2683131f078753fd24723ab8bf1e1b74.git --file example-remote-import.dhall
```
```nix
{ buildDhallGitHubPackage, cdepillabout_2683131f078753fd24723ab8bf1e1b74 }:
  buildDhallGitHubPackage {
    name = "2683131f078753fd24723ab8bf1e1b74";
    githubBase = "gist.github.com";
    owner = "cdepillabout";
    repo = "2683131f078753fd24723ab8bf1e1b74";
    rev = "0de0be4b0238974a33aac07580338105fa5c42e1";
    fetchSubmodules = false;
    sha256 = "0dx7clk1xqj624y34blj58k6j7l60idmzhnng0w6h73vx9l1rpbr";
    directory = "";
    file = "example-remote-import.dhall";
    source = false;
    document = false;
    dependencies = [
      (cdepillabout_2683131f078753fd24723ab8bf1e1b74.overridePackage {
        file = "example-normal.dhall";
        })
      ];
    }
```

Here `buildDhallGitHubPackage` plays a role similar to `fetchUrl` in your example: it fetches the gist using Nix instead of using Dhall's support for remote import.

The main differences from your example are:

* `dhall-nixpkgs` will not generate an all-in-one derivation containing the input file and all of its dependencies

  It's set up to create one file per Dhall package so that they can each be wrapped in `pkgs.dhallPackages.callPackage`.  This is so that it follows the same idioms as the rest of Nixpkgs.

* `dhall-nixpkgs` operates on directories/repositories

  The reason it does this is so that relative references work correctly, which makes things more ergonomic.  For example, since a gist is just a repository under the hood you could have instead had `./example-remote-import.dhall` import `./example-normal.dhall` with a relative import and without a hash:

  ```dhall
  [ 1 + ./example-normal.dhall
  , 3
  ]
  ```

  … and it would have worked with how `dhall-nixpkgs` is set up because it will package the Dhall expression alongside the other files in the same repository.

  However, you might conceivably want a gist to import from a different gist, in which case it would still do the right thing with the PR I created.

cdepillabout:
I did take a look at `dhall-nixpkgs` before starting on this PR, but I don't think I understood enough about Dhall to see how it was related to what I was trying to do.

Although after working on this PR (and with your comment), I think I have a better idea with the relationship between `dhall-nix` and `dhall-nixpkgs`.

What I want is a command similar to `dhall-to-nixpkgs directory`, but where it automatically turns all remote URL imports into FOD in Nix.

For instance, imagine that you had the following file in your current directory:

`example-remote-import.dhall`:

```dhall
[   1
  + https://gist.githubusercontent.com/cdepillabout/2683131f078753fd24723ab8bf1e1b74/raw/7e1b26ff812000e868a4ff108b33a21d03a9a591/example-normal.dhall
      sha256:15f52ecf91c94c1baac02d5a4964b2ed8fa401641a2c8a95e8306ec7c1e3b8d2
, 3
]
```

If you run `dhall-to-nixpkgs directory` on this file, you get the following Nix expression:

```console
$ dhall-to-nixpkgs directory --file "./example-remote-import.dhall"
```

output:

```nix
{ buildDhallDirectoryPackage, cdepillabout_2683131f078753fd24723ab8bf1e1b74 }:
  buildDhallDirectoryPackage {
    name = ".";
    src = ./.;
    file = "./example-remote-import.dhall";
    source = false;
    document = false;
    dependencies = [
      (cdepillabout_2683131f078753fd24723ab8bf1e1b74.overridePackage {
        file = "example-normal.dhall";
        })
      ];
    }
```

Like you say, this would be a good way to package this Dhall file in Nixpkgs.

However, you'd also have to create a separate derivation for the argument `cdepillabout_2683131f078753fd24723ab8bf1e1b74` (which you can now do with #2306 and the command `dhall-to-nixpkgs github https://gist.github.com/cdepillabout/2683131f078753fd24723ab8bf1e1b74.git --file example-normal.dhall`.

But for my own needs, what I want is for the `cdepillabout_2683131f078753fd24723ab8bf1e1b74` argument to be created automatically as a FOD.

Ideally what I'd like is a command like the following:

```console
$ dhall-to-nixpkgs directory --url-as-fod --file "./example-remote-import.dhall"
```

output:

```nix
{ buildDhallDirectoryPackage, fetchurl, dhall, cacert }:
  buildDhallDirectoryPackage {
    name = ".";
    src = ./.;
    file = "./example-remote-import.dhall";
    source = false;
    document = false;
    dependencies = [
      (fetchurl {
          url = "https://gist.githubusercontent.com/cdepillabout/2683131f078753fd24723ab8bf1e1b74/raw/7e1b26ff812000e868a4ff108b33a21d03a9a591/example-normal.dhall";
          hash = "sha256-FfUuz5HJTBuqwC1aSWSy7Y+kAWQaLIqV6DBux8HjuNI=";
          downloadToTemp = true;
          postFetch = "${dhall}/bin/dhall --alpha --plain --file \"\$downloadedFile\" | ${dhall}/bin/dhall encode > \$out";
        }).overrideAttrs (oldAttrs: {
          nativeBuildInputs = (oldAttrs.nativeBuildInputs or []) ++ [ cacert ];
        })
      ];
    }
```

My eventual goal would be to call `dhallToNix` on the output of this derivation to get the full output of `example-remote-import.dhall` in Nix, so that I could manipulate it from within Nix.

-----------------------------------------------------

Where would you suggest that I try to add this functionality?  Would it make sense to add a `--url-as-fod` argument to `dhall-to-nixpkgs directory` as I've shown above?  Or maybe I should just pull this functionality into a separate binary that I maintain?  (Currently in this PR I tried to add something like this functionality directly to `dhall-nix`, but it sounds like trying to fit it into `dhall-nixpkgs` and `buildDhallPackage` would be a better idea.)

Ah, and I guess I should add that while my example is pulling dhall files from gists, I'd like to be able to support arbitrary URLs (as long as they are integrity-protected), not just github urls and gists.

Gabriel439:
Alright, I think I understand what you're trying to do

So there are two ways to go about this:

* Approach 1 - Implement this in `dhall-to-nix`

  … by making the following changes:

  * `dhall-to-nix` would optionally no longer pre-resolve the expression (if we're going to convert Dhall's import resolution to Nix), which entails that it would no longer type-check the expression or β-normalize the expression either
  * `dhall-to-nix` would translate imports protected by integrity checks to self-contained Nix code (without using utilities from Nixpkgs)
  * `dhall-to-nix` would need to now translate a larger subset of Dhall to Nix, such as `let` expressions, since when in this mode the expression would not be β-normalized

* Approach 2 - Implement this in `dhall-to-nixpkgs`

  … by making the following changes:

  * Add a new `pkgs.dhallPackages.buildDhallURL` function to automate the boilerplate so that we only need to provide the `url` and `hash`
  * Use that whenever this mode is enabled for any URL protected by an integrity check that is not already supported by an existing `buildDhall*Package` utility

Approach 2 will probably be *waaay* easier, in my opinion.

I still think in general that we should prefer `dhall-to-nixpkgs` using `buildDhall*Package` utilities when they are available, because then relative imports will work correctly.  For example, suppose that you have a relative import in one of your GitHub repositories or gist protected by a semantic integrity check that refers to another Dhall expression within the same repository/gist: if you use `buildDhallGitHubPackage` to fetch the root expression then the relative import will succeed, but if you use `fetchurl` to retrieve it then it will not succeed.  Moreover, `dhall-to-nixpkgs` wouldn't even know how to translate such a relative import to Nix code.

cdepillabout:
@Gabriel439 Thanks for explaining the two general approaches.

This current PR is trying to implement approach one, but I agree that approach two makes more sense.

I created the repo https://github.com/cdepillabout/example-dhall-nix (as of this writing https://github.com/cdepillabout/example-dhall-nix/tree/78f83e18fb046bfbb6a41109d9b767b84f46f425) that is an example of the Nix code that would be required for approach two.

I've commented all the new functions I had to introduce for this, so please take a look through the code: https://github.com/cdepillabout/example-dhall-nix/blob/78f83e18fb046bfbb6a41109d9b767b84f46f425/default.nix

There are a couple things I specifically wanted to point out:

- The `buildDhallUrl` function is basically the function you mentioned in https://github.com/dhall-lang/dhall-haskell/pull/2304#issuecomment-924597654.  It builds a Dhall Nix package for a remote Dhall import that is protected by an integrity check using Nix fixed-output derivations (FODs).  I'm imagining that `buildDhallUrl` would be used by `dhall-to-nixpkgs`, as you suggest.

-  `myDhallPackage` is an example of what I think `dhall-to-nixpkgs` should output when you enable the (new) option to output dependencies as FODs.  This example exercises:
 
   - local imports
    - remote imports
    - remote imports that have local imports
    - remote imports that have remote imports

   so I think approach should hopefully be quite useful in general.

- `dhallDirectoryToNix` is the final API that I'd personally like to have, and the reason I've been working on this PR.  This makes it very easy for an end-user to read any arbitrary directory of Dhall files into Nix, even with remote imports (as long as there is an integrity check).  This does make use of IFD though, so it would not be allowed in Nixpkgs.  But I imagine it would be helpful to end-users.

cdepillabout:
Let me respond to some of your questions, and ask some of my own.

> if you use `buildDhallGitHubPackage` to fetch the root expression then the relative import will succeed, but if you use `fetchurl` to retrieve it then it will not succeed. Moreover, `dhall-to-nixpkgs` wouldn't even know how to translate such a relative import to Nix code.

I think this is a good point, and something I hadn't considered.

I took this into account in https://github.com/cdepillabout/example-dhall-nix/blob/78f83e18fb046bfbb6a41109d9b767b84f46f425/default.nix#L27 by using Dhall's functionality for resolving imports rather than Nix's `fetchurl`.

You can see in my `fetchDhallUrl` function (which is used as a helper function in `buildDhallUrl`), I just write out the URL and integrity check to a `.dhall` file, and then run `dhall --alpha --plain | dhall encode` on it.  This seems to work well.  Local imports from the remote file seem to work well.  This can be a FOD because we know the expected Nix hash of the output of `dhall encode`.

I couldn't come up with any places this `buildDhallUrl` function wouldn't work (well, other than remote imports without integrity checks), but if there is anything I am overlooking, please let me know.

--------------------------------------------------------

Assuming this approach in Nix is reasonable, the only thing left to do for this PR is to modify `dhall-to-nixpkgs` to output Nix code that makes use of `buildDhallUrl` (similar to https://github.com/cdepillabout/example-dhall-nix/blob/78f83e18fb046bfbb6a41109d9b767b84f46f425/default.nix#L156-L189).

There are a couple possibilities how we could do this:

- Add a new top-level command to `dhall-to-nixpkgs` like `directory-remote-fod` or something.  This would be in addition to the current top-level options, `github` and `directory`.  This new top-level command `directory-remote-fod` would be similar to the current `dhall-to-nixpkgs directory` command, but it would change all Dhall remote imports with integrity checks to use `buildDhallUrl` (similar to https://github.com/cdepillabout/example-dhall-nix/blob/78f83e18fb046bfbb6a41109d9b767b84f46f425/default.nix#L156-L189).

- Add a command-line option to `dhall-to-nixpkgs directory` like `--url-to-fod` or `--remote-import-to-fod`.  This would cause `dhall-to-nixpkgs directory` to make all Dhall remote imports with integrity checks use `buildDhallUrl` (similar to https://github.com/cdepillabout/example-dhall-nix/blob/78f83e18fb046bfbb6a41109d9b767b84f46f425/default.nix#L156-L189).

- Create a completely separate executable to perform this function.

My suggestion would be to add a flag like  `dhall-to-nixpkgs directory `--url-to-fod`, since that seems like the most straight-forward option.

> Use that whenever this mode is enabled for any URL protected by an integrity check that is not already supported by an existing buildDhall*Package utility

For my own use, I'd even be fine with using this mode for URLs with integrity checks that are already supported by an existing `buildDhall*Package` utility.  I'm not sure if you really gain anything by _not_ using `buildDhallUrl`, as long as the URL has an integrity check.

Gabriel439:
The main question I have is why `fetchDhallUrl` is permitted to resolve Dhall imports using Dhall instead of Nix.  The reason I ask is because: if we're fine with resolving imports using Dhall as part of the Nix evaluation or build then you could use `pkgs.dhallToNix` to implement `dhallDirectoryToNix`:

```nix
dhallDirectoryNix = { src, file ? "package.dhall" }: pkgs.dhallToNix "${src}/${file}"
```

… because `pkgs.dhallToNix` resolves everything using Dhall's import system before converting to Nix

cdepillabout:
> The main question I have is why `fetchDhallUrl` is permitted to resolve Dhall imports using Dhall instead of Nix. The reason I ask is because: if we're fine with resolving imports using Dhall as part of the Nix evaluation or build then you could use `pkgs.dhallToNix` to implement `dhallDirectoryToNix`

The reason that `fetchDhallUrl` is permitted to resolve Dhall imports using Dhall instead of Nix is because the derivation is a [fixed-output derivation (FOD)](https://nixos.org/manual/nix/stable/#fixed-output-drvs).  It seems to be possible to programmatically convert any Dhall remote import with an integrity check to a Nix FOD.

However, using `dhallToNix` directly doesn't work, since you don't know the hash of the output of `dhallToNix`.  You only know the integrity checks / hashes of the remote imports in the Dhall file you input to `dhallToNix`.

My solution to this above is a suggested modification to `dhall-to-nixpkgs` that pulls out the remote imports that are protected by an integrity check, and turns them into FOD using `buildDhallUrl` / `fetchDhallUrl`.

-----------------------------------------------------------------------------

When I started to play around with the current Dhall and Nix functionality, using `dhallToNix` to try to implement `dhallDirectoryToNix` was one of the first things I tried.  It ended up not working for Dhall files that have remote imports.

For instance, imagine a Dhall file like:

`example.dhall`:

```dhall
https://raw.githubusercontent.com/cdepillabout/example-dhall-nix/78f83e18fb046bfbb6a41109d9b767b84f46f425/mydhallfile.dhall sha256:e37e31f908c7473ba9ac1770d21eb09b0075663eb266a41be77de67bc5d
d3b87
  # [ "hello" ]
```

If I run this through Dhall, I get the following:

```command
$ dhall --file example.dhall
[ "BILLBILLbillbill"
, "JANEJANEjanejane"
, "TESTTESTtesttest"
, "TESTTESTtesttest"
, "TESTTESTtesttest"
, "hello"
]
```

I'd like to read this in as Nix code (using `dhall-to-nix`).

However, if I try to use `dhallToNix` on this file, there are problems with remote imports, since `dhall-to-nix` doesn't resolve remote imports:

```nix
$ nix repl '<nixpkgs>'
nix-repl> dhallToNix "${./example.dhall}" 
error: builder for '/nix/store/k3y5xsd7xbn4dgiwgcw1xmqbdrgfafna-dhall-compiled.nix.drv' failed with exit code 1;
       >
       > Error: Remote host not found
       >
       > URL: https://raw.githubusercontent.com/cdepillabout/example-dhall-nix/78f83e18fb046bfbb6a41109d9b767b84f46f425/mydhallfile.dhall
```

Alternatively, I could try to turn my `example.dhall` file into a Dhall package, build it with `buildDhallDirectoryPackage`, and then pass the output of that to `dhallToNix`.  But `buildDhallDirectoryPackage` doesn't work here either because there are remote imports:

```nix
nix-repl> :b dhallPackages.buildDhallDirectoryPackage { src = ./. ; file = "example.dhall" ; name = "example"; }
error: builder for '/nix/store/ffhhwy5ajni0wh8b4ggisq42lwj9x094-example.drv' failed with exit code 1;
       > Error: Cannot import HTTP URL.
       >
       > Dhall was compiled without the 'with-http' flag.
       >
       > The requested URL was: https://raw.githubusercontent.com/cdepillabout/example-dhall-nix/78f83e18fb046bfbb6a41109d9b767b84f46f425/mydhallfile.dhall
       >
       >
       > 1 https://raw.githubusercontent.com/cdepillabout/example-dhall-nix/78f83e18fb046bfbb6a41109d9b767b84f46f425/mydhallfile.dhall sha256:e37e31f908c7473ba9ac1770d21eb09b0075663eb266a41b
e77de67bc5dd3b87
       >
```

What I would need to get this working is some way to translate all remote imports to FOD, and pass them as dependencies to `buildDhallDirectoryPackage`.

This is the `buildDhallUrl` function I created in https://github.com/cdepillabout/example-dhall-nix/blob/78f83e18fb046bfbb6a41109d9b767b84f46f425/default.nix#L70:

```nix
nix-repl> myDhallStuff = import (builtins.fetchTarball "https://github.com/cdepillabout/example-dhall-nix/archive/78f83e18fb046bfbb6a41109d9b767b84f46f425.tar.gz")
nix-repl> :b dhallPackages.buildDhallDirectoryPackage {
  src = ./. ;
  file = "example.dhall" ;
  name = "example";
  dependencies = [
    (myDhallStuff.buildDhallUrl {
      url = "https://raw.githubusercontent.com/cdepillabout/example-dhall-nix/78f83e18fb046bfbb6a41109d9b767b84f46f425/mydhallfile.dhall";
      hash = "sha256-434x+QjHRzuprBdw0h6wmwB1Zj6yZqQb533me8XdO4c="; 
      dhall-hash = "sha256:e37e31f908c7473ba9ac1770d21eb09b0075663eb266a41be77de67bc5dd3b87";
    })
  ];
}

this derivation produced the following outputs:
  out -> /nix/store/51fhnd4ld48vcbis3v3fv12z5qldihhn-example
```

cdepillabout:
I put together an example of how `dhall-to-nixpkgs` could be modified to support building remote imports as FOD in #2318.

(Normally I wouldn't have gone ahead with implementing this until we fleshed out a plan for exactly what should be done, but I needed the functionality for putting together a prototype of a `purescript2nix` tool, so I decided to open a PR with my implementation of `dhall-to-nixpkgs directory --url-as-fod`.)

cdepillabout:
> how do you handle transitive imports of the Dhall imports that you turn into fixed-output derivations?

I was worried about this as well, so I worked it in to the example I linked to above (https://github.com/cdepillabout/example-dhall-nix/tree/78f83e18fb046bfbb6a41109d9b767b84f46f425).

In this example, https://github.com/cdepillabout/example-dhall-nix/blob/78f83e18fb046bfbb6a41109d9b767b84f46f425/mydhallfile.dhall is the main Dhall file that I import in https://github.com/cdepillabout/example-dhall-nix/blob/78f83e18fb046bfbb6a41109d9b767b84f46f425/default.nix.

This `mydhallfile.dhall` file has:

-   a local import to https://github.com/cdepillabout/example-dhall-nix/blob/78f83e18fb046bfbb6a41109d9b767b84f46f425/upperAndDouble.dhall

    This `upperAndDouble.dhall` contains a remote import with an integrity check to https://raw.githubusercontent.com/dhall-lang/dhall-lang/9758483fcf20baf270dda5eceb10535d0c0aa5a8/Prelude/Text/upperASCII.dhall.

-   two remote imports:

    -   https://raw.githubusercontent.com/cdepillabout/example-dhall-repo/c1b0d0327146648dcf8de997b2aa32758f2ed735/example1.dhall

        This `example1.dhall` file contains:

        - a remote import of https://prelude.dhall-lang.org/List/replicate (without an integrity check)
        - a local import of `./mkName.dhall` (which would be the file https://raw.githubusercontent.com/cdepillabout/example-dhall-repo/c1b0d0327146648dcf8de997b2aa32758f2ed735/mkName.dhall)

    -   https://raw.githubusercontent.com/dhall-lang/dhall-lang/9758483fcf20baf270dda5eceb10535d0c0aa5a8/Prelude/List/map.dhall

With the changes I added in #2318, running `dhall-to-nixpkgs directory --url-as-fod` on this local `./mydhallfile.dhall` file would result in the following output:

```console
$ dhall-to-nixpkgs directory ./example-dhall-nix --file mydhallfile.dhall --url-as-fod
{ buildDhallDirectoryPackage, buildDhallUrl }:
  buildDhallDirectoryPackage {
    name = "example-dhall-nix";
    src = ./example-dhall-nix;
    file = "mydhallfile.dhall";
    source = false;
    document = false;
    dependencies = [
      (buildDhallUrl {
        url = "https://raw.githubusercontent.com/cdepillabout/example-dhall-repo/c1b0d0327146648dcf8de997b2aa32758f2ed735/example1.dhall";
        hash = "sha256-ZTSiQUXpPbPfPvS8OeK6dDQE6j6NbP27ho1cg9YfENI=";
        dhall-hash = "sha256:6534a24145e93db3df3ef4bc39e2ba743404ea3e8d6cfdbb868d5c83d61f10d2";
        })
      (buildDhallUrl {
        url = "https://raw.githubusercontent.com/dhall-lang/dhall-lang/9758483fcf20baf270dda5eceb10535d0c0aa5a8/Prelude/List/map.dhall";
        hash = "sha256-3YRf+0Vo1AMn8qgX60LRxhOLkpynWNULwzES7zyIVoA=";
        dhall-hash = "sha256:dd845ffb4568d40327f2a817eb42d1c6138b929ca758d50bc33112ef3c885680";
        })
      (buildDhallUrl {
        url = "https://raw.githubusercontent.com/dhall-lang/dhall-lang/9758483fcf20baf270dda5eceb10535d0c0aa5a8/Prelude/Text/upperASCII.dhall";
        hash = "sha256-Ra5PvYFLBHTmXCik7pKyO5eYkvpbtzcwvJlnWueQyik=";
        dhall-hash = "sha256:45ae4fbd814b0474e65c28a4ee92b23b979892fa5bb73730bc99675ae790ca29";
        })
      ];
    }
```

You can see that the three remote imports (https://raw.githubusercontent.com/cdepillabout/example-dhall-repo/c1b0d0327146648dcf8de997b2aa32758f2ed735/example1.dhall and https://raw.githubusercontent.com/dhall-lang/dhall-lang/9758483fcf20baf270dda5eceb10535d0c0aa5a8/Prelude/List/map.dhall from `mydhallfile.dhall`, and `https://raw.githubusercontent.com/dhall-lang/dhall-lang/9758483fcf20baf270dda5eceb10535d0c0aa5a8/Prelude/Text/upperASCII.dhall` from the locally-imported `upperAndDouble.dhall` file) get turned into calls to `buildDhallUrl`.

`buildDhallUrl` is defined here: https://github.com/cdepillabout/example-dhall-nix/blob/78f83e18fb046bfbb6a41109d9b767b84f46f425/default.nix#L51-L70.  Most of the code here is just for turning a fetched dhall file into a Nix package.  It is very similar to `buildDhallPackage` in Nixpkgs.  The interesting part is `fetchDhallUrl`, which is called internally by `builddhallPackage`: https://github.com/cdepillabout/example-dhall-nix/blob/78f83e18fb046bfbb6a41109d9b767b84f46f425/default.nix#L19-L49.

Here's `fetchDhallUrl` (simplified slightly):

```nix
fetchDhallUrl =
  { url, hash, dhall-hash }:
  runCommand
    (baseNameOf url)
    {
      outputHashAlgo = null;
      outputHash = hash;
      name = baseNameOf url;
      nativeBuildInputs = [ cacert ];
    }
    ''
      echo "${url} ${dhall-hash}" > in-dhall-file
      ${dhall}/bin/dhall --alpha --plain --file in-dhall-file | ${dhallNoHTTP}/bin/dhall encode > $out
    '';
```

This is called with the same argument as `buildDhallPackage`, so a call to `fetchDhallUrl` might look like:

```nix
fetchDhallUrl {
  url = "https://raw.githubusercontent.com/cdepillabout/example-dhall-repo/c1b0d0327146648dcf8de997b2aa32758f2ed735/example1.dhall";
  hash = "sha256-ZTSiQUXpPbPfPvS8OeK6dDQE6j6NbP27ho1cg9YfENI=";
  dhall-hash = "sha256:6534a24145e93db3df3ef4bc39e2ba743404ea3e8d6cfdbb868d5c83d61f10d2";
}
```

You can see that the derivation produced by `fetchDhallUrl` is a fixed-output derivation (FOD), since it has its `outputHash` specified.  This means that it is able to use the network (as long as the hash of the output of the derivation is what we pass in for the `hash` argument).

`fetchDhallUrl` basically takes the `url` and `dhall-hash` arguments and writes them to a file called `in-dhall-file`.  It then calls `dhall` on this file, and alpha-normalizes the result.  It then encodes the result of alpha-normalization.  We know that this encoded result will match the `dhall-hash` passed in, as well as the Nix `hash`.  That is what enables this to be a FOD.

> how do you handle transitive imports of the Dhall imports that you turn into fixed-output derivations?

In the Nix/bash code here there is no specific logic for handling the transitive imports of the Dhall imports.  The Dhall URL integrity checks appear to take into account the fully alpha-normalized, encoded remote import.  As far as I could tell, this appears to "just work".

Is there some sort of import or way to construct a Dhall file that would cause this not to work?

cdepillabout:
> This was the missing piece of the puzzle for me. I didn't realize that if you added a fixed output hash then Nix would permit network access and therefore permit dhall to resolve imports.

Ah, sorry for not making this more clear earlier in the thread!

> What if we were to use this for dhallToNix instead of dhall-to-nixpkgs? In other words, what if there were a variation on dhallToNix that let you protect the intermediate drv derivation with a fixed output hash so that dhall can make network calls?

Let me make sure I understand.  You're suggesting adding a function like [`dhallToNix`](https://github.com/NixOS/nixpkgs/blob/79f5a81cb3855d7d2e8fe642c78d43da0d6fd9b8/pkgs/build-support/dhall-to-nix.nix) that additionally takes a hash.  I imagine it might look something like this:

```nix
{ stdenv, dhall-nix, writeText }:

let
  dhallToNixFOD = code: hash:
    let
      file = writeText "dhall-expression" code;

      drv = stdenv.mkDerivation {
        name = "dhall-compiled.nix";

        outputHashAlgo = null;

        outputHash = hash;

        buildCommand = ''
          dhall-to-nix <<< "${file}" > $out
        '';

        buildInputs = [ dhall-nix ];
      };

    in
      import drv;
in
  dhallToNixFOD
```

You could call `dhallToNixFOD` like the following:

```nix
dhallToNixFOD
  ''
    (https://prelude.dhall-lang.org/List/replicate sha256:d4250b45278f2d692302489ac3e78280acb238d27541c837ce46911ff3baa347)
      3
      Text
      "hello"
  ''
  "sha256-ZTSiQUXpPbPfPvS8OeK6dDQE6j6NbP27ho1cg9YfENI="
```

When running this, a `hash` has been provided to the derivation, making this a FOD, so `dhall-to-nix` is able to access the internet.  `dhall-to-nix` evaluates this Dhall expression to the value `["hello", "hello", "hello"]`.  It then translates this to Nix, and writes it to `$out`.  The hash `sha256-ZTSiQUXpPbPfPvS8OeK6dDQE6j6NbP27ho1cg9YfENI=` is for this output file `$out`.  

This means that we know that a file with the contents `["hello", "hello", "hello"]` has the hash `sha256-ZTSiQUXpPbPfPvS8OeK6dDQE6j6NbP27ho1cg9YfENI=`.

There are two problems I can see with this:

1. This doesn't support local, relative imports.

    This isn't a big problem though.  We could easily change the API for this `dhallToNixFOD` to be more like `buildDhallPackage`, and operate on an input directory instead of raw Dhall text.

2. This wouldn't work for the type of API I'd like.  I'd like the user to be able to read/convert Dhall files as Nix, without having to add any additional hashes (assuming all of the remote imports in their input Dhall files have integrity checks).

    For instance, imagine a user has a directory of Dhall files like https://github.com/cdepillabout/example-dhall-nix/tree/78f83e18fb046bfbb6a41109d9b767b84f46f425.  (Although here I'm assuming there is no `default.nix` already in the directory.)

    I'd like them to be able to add a file that looks like the following:

    `my-dhall-stuff.nix`:

    ```nix
    with import <nixpkgs> {};

    dhallDirectoryToNix {
      src = ./.;
      file = "mydhallfile.dhall";
    };
    ```

    They should then be able to run this in the repl, and have the Dhall files automatically converted to Nix:

    ```console
    $ nix repl
    nix-repl> x = import ./my-dhall-stuff.nix;
    nix-repl> x
    [ "BILLBILLbillbill" "JANEJANEjanejane" "TESTTESTtesttest" "TESTTESTtesttest" "TESTTESTtesttest" ]
    ```

    I'm imagining that this `dhallDirectoryToNix` function is provided by Nixpkgs, and uses `buildDhallDirectoryPackage` with `buildDhallUrl` as shown in previous comments (or in https://github.com/cdepillabout/example-dhall-nix/blob/78f83e18fb046bfbb6a41109d9b767b84f46f425/default.nix).  In this case, the user doesn't have to add an explicit hash for the output `.nix` file.

    If the user instead used `dhallToNixFOD` as above, they'd have to explicitly add a `hash` to the call to `dhallToNixFOD`.  They'd have to update this hash every time the output of `mydhallfile.dhall` changed.  This is the main downside of using non-automatically-generated FODs (and why people dislike them in Nixpkgs).

Gabriel439:
There's still one thing missing from my understanding, which is how `dhallDirectoryToNix` avoids both (A) the user having to specify a hash and (B) the user having to run an external command-line tool to generate Nix code.

My very rough understanding is that `dhallDirectoryToNix` side-steps the user having to specify a hash by using import-from-derivation on the output of some command-line tool (e.g. `dhall-to-nixpkgs --url-fod`), but wouldn't that command-line tool still require network access within its Nix build to resolve Dhall imports?

PR #2305, opened 2021-09-20
Support *ENV.dhall for import failure tests

timbertson:
Required for https://github.com/dhall-lang/dhall-lang/pull/1229

Given that this will quietly ignore any *ENV.dhall test cases, I wonder if we should add a dot to the suffix to avoid any surprises (i.e. `testName.ENV.dhall` instead of `testNameENV.dhall`)


Gabriel439 (approved):
I think the `ENV.dhall` suffix is fine the way it is

PR #2316, opened 2021-10-12
Assume Swagger integer is Dhall Natural unless evidence to contrary

arobertn:
This PR reverses the assumption made when converting a Swagger spec to Dhall types about the signedness of integers. In most cases when dealing with configuration, an "integer" is meant to be a nonnegative number, be it a timeout, a port number, a count, or some other real-world quantity. Thus rather than requiring a range _excluding_ negative numbers to make a Natural, we require a range explicitly _including_ them to make an Integer.

### DIFF
--- a/dhall-openapi/openapi-to-dhall/Main.hs
+++ b/dhall-openapi/openapi-to-dhall/Main.hs
@@ -182,9 +182,9 @@ parseNatIntExceptions =
     bimap errorBundlePretty id $ result (pack s)
   where
     parser = do
-      typ <- some (alphaNumChar)
+      typ <- some alphaNumChar
       char '.'
-      fld <- some (alphaNumChar)
+      fld <- some alphaNumChar
       return (typ, fld)
     result = parse ((Dhall.Parser.unParser parser `sepBy1` char ',') <* eof) "EXCEPTIONS"
 
--- a/dhall-openapi/openapi-to-dhall/Main.hs
+++ b/dhall-openapi/openapi-to-dhall/Main.hs
@@ -215,7 +215,7 @@ parseSplits =
         mo <- parseModel
         return mo
       return (path, model)
-    result = parse ((Dhall.Parser.unParser parser `sepBy1` char ',') <* eof) "MAPPING"
+    result = parse ((Dhall.Parser.unParser parser `sepBy1` char ',') <* eof) "SPLITS"
 
 
 parseOptions :: Options.Applicative.Parser Options
@@ -248,7 +248,7 @@ parseOptions = Options <$> parseSkip <*> parseNaturalInt <*> parseNatIntExceptio
         (  Options.Applicative.long "splitPaths"
         <> Options.Applicative.help
           "Specifiy path and model name pairs with paths being delimited by '.' and pairs separated by '=' for which \
-          \definitions should be aritifically split with a ref: \n\
+          \definitions should be artificially split with a ref: \n\
           \'(com.example.v1.Certificate).spec=com.example.v1.CertificateSpec'\n\
           \When the model name is omitted, a guess will be made based on the first word of the definition's \
           \description. Also note that top level model names in a path must use () when the name contains '.'"
--- a/dhall-openapi/openapi-to-dhall/Main.hs
+++ b/dhall-openapi/openapi-to-dhall/Main.hs
@@ -61,7 +61,7 @@ import qualified Text.Megaparsec.Char.Lexer as Megaparsec.Lexer
 data Options = Options
     { skipDuplicates :: Bool
     , preferNaturalInt :: Bool
-    , natIntExceptions :: String
+    , natIntExceptions :: [(String,String)]
     , prefixMap :: Data.Map.Map Prefix Dhall.Import
     , splits :: Data.Map.Map ModelHierarchy (Maybe ModelName)
     , filename :: String
@@ -176,6 +176,18 @@ parseImport :: String -> Types.Expr -> Dhall.Parser.Parser Dhall.Import
 parseImport _ (Dhall.Note _ (Dhall.Embed l)) = pure l
 parseImport prefix e = fail $ "Expected a Dhall import for " <> prefix <> " not:\n" <> show e
 
+parseNatIntExceptions :: Options.Applicative.ReadM [(String, String)]
+parseNatIntExceptions =
+  Options.Applicative.eitherReader $ \s ->
+    bimap errorBundlePretty id $ result (pack s)
+  where
+    parser = do
+      typ <- some (alphaNumChar)
+      char '.'
+      fld <- some (alphaNumChar)
+      return (typ, fld)
+    result = parse ((Dhall.Parser.unParser parser `sepBy1` char ',') <* eof) "EXCEPTIONS"
+
 parsePrefixMap :: Options.Applicative.ReadM (Data.Map.Map Prefix Dhall.Import)
 parsePrefixMap =
   Options.Applicative.eitherReader $ \s ->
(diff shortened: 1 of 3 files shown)


Gabriel439:
I would prefer to keep the current behavior, even if `Integer` literals are less ergonomic than `Natural` literals.  Keep in mind that although `openapi-to-dhall` was originally created for `dhall-kubernetes`, it's goal is to to do the right thing for any OpenAPI specification.  So even if most uses of integers in `dhall-kubernetes` are non-negative, I still prefer to try to mirror the OpenAPI spec as closely as possible

arobertn:
The ergonomics are of course a concern, but correctness is more important, and dhall-kubernetes is a major but not the only user of this tool. So I've now updated the implementation to be optional, controlled by two command-line parameters:

- `--preferNaturalInt` - swaps the assumption that openapi integer is Dhall Integer to Dhall Natural
- `natIntExceptions <str>` - provides a list of fields that should be processed with the opposite assumption

I've submitted a companion PR https://github.com/dhall-lang/dhall-kubernetes/pull/178 that utilizes these optional arguments to generate dhall-kubernetes.

As said, these arguments are completely optional to openapi-to-dhall and it will default to translating Swagger integer as Dhall Integer. However for what it's worth, I can't really see any evidence that the creators of the Swagger specification themselves have thought about the natural/unsigned – integer/signed distinction. For instance in the [first major example in the specification](https://swagger.io/specification/#components-object), they define clearly non-negative quantities using only "integer" of format "[signed] int32" with no mention of limits. Dhall is just more nuanced here while it may be the majority of Swagger instances don't pay much attention to the issue either. But as I say most quantities in Dhall's target domain of configuration are naturally unsigned.

dhall-openapi/openapi-to-dhall/Main.hs
```diff
@@ -60,6 +60,8 @@ import qualified Text.Megaparsec.Char.Lexer as Megaparsec.Lexer
 -- | Top-level program options
 data Options = Options
     { skipDuplicates :: Bool
+    , preferNaturalInt :: Bool
+    , natIntExceptions :: String
```
Gabriel439, line 64:
```suggestion
    , natIntExceptions :: [(String, String)]
```

Would it be possible to do the string splitting within the command-line parsing logic so that the field can have this type?

Gabriel439 (commented):
Sorry for the long delay in reviewing this!  I think this is good approach, and just have a couple of comments

arobertn:
OK great I will make those adjustments next week if I don't get to it this one.

dhall-openapi/openapi-to-dhall/Main.hs
arobertn: Done

dhall-openapi/src/Dhall/Kubernetes/Convert.hs
```diff
@@ -223,15 +231,22 @@ toTypes' prefixMap typeSplitter definitions toMerge
...
-              "integer" -> case (minimum_ definition, exclusiveMinimum definition) of
-                (Just min_, Just True) | min_ >= -1 -> (Dhall.Natural, Data.Map.empty)
-                (Just min_, _)         | min_ >= 0 -> (Dhall.Natural, Data.Map.empty)
-                _                      -> (Dhall.Integer, Data.Map.empty)
+              "integer" -> if prefNatInt then case (minimum_ definition, exclusiveMinimum definition,
+                          maximum_ definition, exclusiveMaximum definition) of
+                  (Just min_, Just True, _, _) | min_ < -1 -> (Dhall.Integer, Data.Map.empty)
+                  (Just min_, _, _, _)         | min_ < 0  -> (Dhall.Integer, Data.Map.empty)
+                  (_, _, Just max_, Just True) | max_ < 1  -> (Dhall.Integer, Data.Map.empty)
+                  (_, _, Just max_, _)         | max_ < 0  -> (Dhall.Integer, Data.Map.empty)
```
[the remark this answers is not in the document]
arobertn: All bounds are optional, so if we get past the first two min cases, then either there _is_ a min and it is >=0, or there is no min. Then there are three cases:

    1. no max -> no limits specified, so Nat by default
    2. max >0 -> Nat by default, assume intent is a bounded positive number
    3. max <=0 -> implies an unbounded negative number

So we are detecting this last case.

arobertn:
All concerns addressed and rebased on master in the upstream branch.

dhall-openapi/openapi-to-dhall/Main.hs
```diff
@@ -174,6 +176,18 @@ parseImport :: String -> Types.Expr -> Dhall.Parser.Parser Dhall.Import
...
 
+parseNatIntExceptions :: Options.Applicative.ReadM [(String, String)]
+parseNatIntExceptions =
+  Options.Applicative.eitherReader $ \s ->
+    bimap errorBundlePretty id $ result (pack s)
+  where
+    parser = do
+      typ <- some (alphaNumChar)
+      char '.'
+      fld <- some (alphaNumChar)
```
Gabriel439, line 187: Minor suggestion: the parentheses are not necessary

```suggestion
      typ <- some alphaNumChar
      char '.'
      fld <- some alphaNumChar
```

dhall-openapi/openapi-to-dhall/Main.hs
```diff
@@ -174,6 +176,18 @@ parseImport :: String -> Types.Expr -> Dhall.Parser.Parser Dhall.Import
...
+      return (typ, fld)
+    result = parse ((Dhall.Parser.unParser parser `sepBy1` char ',') <* eof) "EXCEPTIONS"
```
Gabriel439, line 189: I don't know if the name shows up in the error message or not, but just in case it does, maybe set the option name as the "file" name:

```suggestion
    result = parse ((Dhall.Parser.unParser parser `sepBy1` char ',') <* eof) "--natIntExceptions"
```

dhall-openapi/openapi-to-dhall/Main.hs
```diff
@@ -201,17 +215,28 @@ parseSplits =
         mo <- parseModel
         return mo
       return (path, model)
-    result = parse ((Dhall.Parser.unParser parser `sepBy1` char ',') <* eof) "MAPPING"
+    result = parse ((Dhall.Parser.unParser parser `sepBy1` char ',') <* eof) "SPLITS"
```
Gabriel439, line 218: Same suggestion here:

```suggestion
    result = parse ((Dhall.Parser.unParser parser `sepBy1` char ',') <* eof) "--splitPaths"
```

Gabriel439 (approved):
Thank you!  Just some cosmetic suggestions

dhall-openapi/openapi-to-dhall/Main.hs
arobertn: Done. Interestingly haskell-language-server does not pick that up.

dhall-openapi/openapi-to-dhall/Main.hs
arobertn: The usage string now reads like this:

```
Usage: openapi-to-dhall [--skipDuplicates] [--preferNaturalInt] 
                        [--natIntExceptions EXCEPTIONS] [--prefixMap MAPPING] 
                        [--splitPaths SPLITS] FILE [--crd]
```

An error will come out like so:

```
option --natIntExceptions: EXCEPTIONS:1:133:
  |
1 | ContainerStateTerminated.exitCode,ContainerStateTerminated.signal,PodSpec.priority,PriorityClass.valueCustomResourceColumnDefinition.priority
  |                                                                                                                                     ^
unexpected '.'
expecting ',', alphanumeric character, or end of input
```

The "EXCEPTIONS", "MAPPING", etc. are not variable names, but just strings shown in the usage and error messages respectively. I tried putting something more descriptive for the usage one like `type-1.field-1[,type-2.field-2]+` but it didn't look right and duplicates the `--help` output. I also tried putting something more generic like "ARG" for the error message one but that didn't improve things enough to be worth breaking the connection to the usage string. And duplicating the argument name obviously doesn't make sense. I'm open for any other suggestions but maybe it is not the worst as it is.

Gabriel439: Ah, okay, I didn't realize that the option name would already show up in the error message.  Then I think you were right to keep it the way it is

Gabriel439:
Thank you for doing this! 🙂

arobertn:
No problem; I'll work on the PR on the dhall-kubernetes side next week.

PR #2329, opened 2021-11-10
Build against hashable-1.4

Gabriel439:
Related to https://github.com/commercialhaskell/stackage/issues/6268

Note that I have not actually tested this change since it requires bumping or jailbreaking
way too many dependencies.  However, based on the `hashable` CHANGELOG and our usage
of it I'm fairly confident that this is a safe change


sjakobi (approved):
This builds cleanly for me with

```
cabal build dhall -O0 --constraint 'hashable >= 1.4' --allow-newer=hashable
```

I've pushed a matching Hackage revision.

Issue #2337, opened 2021-11-23
openapi-to-dhall discards type information in additionalProperties


BrianHicks:
Certain CRDs have name/value pairs, where the value is not a string type but something else. There are some examples in Zalando's Postgres operator: [`preparedDatabase`](https://github.com/zalando/postgres-operator/blob/a6d4f23f90fe12822a19e758ff400233c9212e00/manifests/postgresql.crd.yaml#L372-L393), [`users`](https://github.com/zalando/postgres-operator/blob/a6d4f23f90fe12822a19e758ff400233c9212e00/manifests/postgresql.crd.yaml#L515-L551).

openapi-to-dhall discards everything in `additionalProperties`, so these schemas end up being `{ field = None (List { mapKey : Text, mapValue : Text }) }`, and values generated by `dhall-to-yaml` and fed into Kubernetes are rejected due to schema validation failures.

Gabriel439:
I'm not that familiar with Kubernetes / CRDs, but it might help if you explain in terms of how the Haskell types for the CRD data structure need to be fixed

My very rough understanding is that based on what you are saying the Haskell type for `V1JSONSchemaProps` is not correct:

https://github.com/dhall-lang/dhall-haskell/blob/fc56822eb56ca280a7d574eb9c056c2b8bb0a05c/dhall-openapi/src/Dhall/Kubernetes/Convert.hs#L432-L444

… but the part that I don't understand is what change needs to be made to that type

BrianHicks:
I'm not sure what change would need to be made to the code either… else I'd make it! 😬

But in the hopes it clears things up: this is not a CRD-specific thing. In JSON schemas / OpenAPI, `additionalProperties` is able to specify a type for an arbitrary string key in the same way that `properties.NAME.type` or `properties.NAME.items` can. In that case, I would expect the conversion to work like:

| JSON Schema | Dhall |
|-|-|
| `{ "properties": { "x": { "type": "integer" } } }` | `{ x : Integer }` |
| `{ "properties": { "x": { "type": "array", "items": "integer" } } }` | `{ x : List Integer }` |
| `{ "additionalProperties": "integer" }` | `List { mapKey : Text, mapValue : Integer }` |

(pretending that all those have `"type": "object"` and `x` is required in the first two to save a little space)

Gabriel439:
Thanks!  Now I understand what you mean and I believe #2343 will fix this

BrianHicks:
[...]

PR #2343, opened 2021-12-03
dhall-openapi: Fix handling of additionalProperties

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/2337

Basically, `additionalProperties` is analogous to
`Map Text` in `dhall`, which this change codifies.

Before this change, an `object` with an `additionalProperties`
field would get encoded as `Map Text Text`, regardless of what
the property type was.  Now the property type is correctly
respected by the conversion code.


Gabriel439:
Here is the matching change to `dhall-kubernetes`: https://github.com/dhall-lang/dhall-kubernetes/pull/180

PR #2349, opened 2021-12-10
Add FromDhall instances for Int{,8,16,32,64}

jchia:
#2348

### DIFF
--- a/dhall/src/Dhall/Marshal/Decode.hs
+++ b/dhall/src/Dhall/Marshal/Decode.hs
@@ -146,7 +146,6 @@ import Data.Functor.Contravariant
     , Predicate (..)
     )
 import Data.Hashable                    (Hashable)
-import Data.Int                         (Int16, Int32, Int64, Int8)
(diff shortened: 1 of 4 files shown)


Gabriel439:
It looks like you will also need to fix this now that `Dhall.Marshall.Internal` re-exports `Int{8,16,32,64}`:

```
src/Dhall/Marshal/Decode.hs:149:1: error: [-Wunused-imports, -Werror=unused-imports]
    The import of ‘Data.Int’ is redundant
      except perhaps to import instances from ‘Data.Int’
    To import instances alone, use: import Data.Int()
    |
149 | import Data.Int                         (Int16, Int32, Int64, Int8)
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

jchia:
Fixed.

Gabriel439:
Thank you! 🙂

Issue #2351, opened 2021-12-17
dhall-to-json: should --pretty vs --compact change the type of numbers in the output?


mikeantonacci:
Minimal (contrived) example:

```bash
$ echo '{"n":2123234787847813783278587}' | json-to-dhall | dhall --annotate
{ n = 2123234787847813783278587 } : { n : Natural }
$ echo '{"n":2123234787847813783278587}' | json-to-dhall | dhall --annotate | dhall-to-json --compact
{"n":2123234787847813783278587}
$ echo '{"n":2123234787847813783278587}' | json-to-dhall | dhall --annotate | dhall-to-json --pretty
{
  "n": 2.123234787847813783278587e24
}
```
This creates a situation where I could serialize something to JSON from Python containing a large `int`, convert it to Dhall and see that it becomes a `Natural` and everything is fine, and then if I run `dhall-to-json` I may get back something that looks like a `float` and loses precision when I deserialize depending on whether I pretty-printed the output or not:
```python
>>> import json
>>> json.dumps({"n":2123234787847813783278587})
'{"n": 2123234787847813783278587}'
>>> json.loads('{"n":2123234787847813783278587}')
{'n': 2123234787847813783278587}
>>> json.loads('''{
...   "n": 2.123234787847813783278587e24
... }''')
{'n': 2.1232347878478138e+24}
```

Gabriel439:
Whoops!  The scientific notation for large numbers is unintentional.  Fortunately, we can fix it since the JSON pretty-printing library we're using lets us disable that behavior

Gabriel439:
Fix is up here: https://github.com/dhall-lang/dhall-haskell/pull/2352

PR #2352, opened 2021-12-17
dhall-json: Don't use scientific notation for large integers

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/2351


sjakobi (approved):
:+1:

Gabriel439:
Actually, I might still make one more change to use scientific notation for floating point numbers

Gabriel439:
So what I ended up doing is reusing the exact same logic as plain `aeson` (e.g. `--plain` mode) for pretty-printing numbers, to avoid inconsistencies in behavior between `--compact` and `--plain`

PR #2354, opened 2021-12-22
Allow `Some` in `with` expression

darichey:
Fix for small discrepancy between `dhall-haskell` and the reference implementation/[standard](https://github.com/dhall-lang/dhall-lang/blob/502856d9d09b555044e140a87ab34e120a1869d4/standard/dhall.abnf#L859).

Before:
```
⊢ { Some = 0 } with Some = 1

Error: Invalid input

(input):1:23:
  |
1 | { Some = 0 } with Some = 1
  |                       ^
expecting any label or whitespace
```

After:
```
⊢ { Some = 0 } with Some = 1

{ Some = 1 }
```


sjakobi (approved):
Thanks! Ideally we'd also have a test for this in [`tests/parser/success/unit`](https://github.com/dhall-lang/dhall-lang/tree/master/tests/parser/success/unit). Would you like to give this a try?


To generate the corresponding `.dhallb` CBOR output, you can run the `dhall:tasty` tests in this repo with the `--accept` flag, e.g.

```
cabal test dhall:tasty --test-options=--accept
```

Also see https://github.com/dhall-lang/dhall-lang/tree/master/tests#running-parser-tests.

darichey:
Gave it a try!

Gabriel439:
Thank you for fixing that! 🙂

PR #2365, opened 2022-01-16
Fix Mergify configuration

Gabriel439:
They have a new queue mechanism that supersedes "strict: smart"


Gabriel439:
The check is inaccurate since I believe it uses the Mergify configuration on the `master` branch, but I validated the new configuration file using the instructions here:

https://docs.mergify.com/configuration/#validation

PR #2371, opened 2022-01-22
Re-enable some CORS tests

Gabriel439:
These tests now work correctly after upstream fixes to
`test.dhall-lang.org`


Gabriel439:
It looks like tests are timing out due to the same IPv6-related issues as in https://github.com/dhall-lang/dhall-lang/issues/1268.  I've opened a support ticket with Linode to try to fix the underlying problem

PR #2388, opened 2022-02-17
Make IntOrString = Natural | Text if preferNaturalInt

arobertn:
#2316 added support for optionally translating Swagger `Int` to Dhall `Natural`, however as mentioned in https://github.com/dhall-lang/dhall-kubernetes/pull/178 this leads to a mismatch with IntOrString fields which were still translated to a union of Integer and String. This PR changes this to translate IntOrString to a new NatOrString union when `--preferNaturalInt` is active.

The `--natIntExceptions` argument applies to IntOrString as it applies to Integer. That is, named fields are rendered as IntOrString when `--prefNaturalInt` is active, and NatOrString when it is inactive.

For the case of dhall-kubernetes, in the openapi specs from 1.12 to 1.22, IntOrString is used either for ports (where a string port or service name can be used instead), or nonnegative amounts (where a percentage can be given instead, e.g., "25%"). Thus, no new exceptions are needed for dhall-kubernetes.

Testing done: diff of generated files against previous is limited to IntOrString fields, running various combinations of `--prefNaturalInt` and `--natIntExceptions` generates expected results.

### DIFF
--- a/dhall-openapi/dhall-openapi.cabal
+++ b/dhall-openapi/dhall-openapi.cabal
@@ -1,6 +1,6 @@
 Cabal-Version:  >=1.11
 Name:           dhall-openapi
-Version:        1.0.2
+Version:        1.0.3
 Homepage:       https://github.com/dhall-lang/dhall-haskell/tree/master/dhall-openapi#dhall-openapi
 Author:         Fabrizio Ferrai
 Maintainer:     Gabriel439@gmail.com
--- a/dhall-openapi/dhall-openapi.cabal
+++ b/dhall-openapi/dhall-openapi.cabal
@@ -1,6 +1,6 @@
 Cabal-Version:  >=1.11
 Name:           dhall-openapi
-Version:        1.0.3
+Version:        1.0.2
 Homepage:       https://github.com/dhall-lang/dhall-haskell/tree/master/dhall-openapi#dhall-openapi
 Author:         Fabrizio Ferrai
 Maintainer:     Gabriel439@gmail.com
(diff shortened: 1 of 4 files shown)


Gabriel439:
You can generate an additional `NatOrString` type even if it's not derived from the Kubernetes OpenAPI spec.  I believe the place you would do that is here:

https://github.com/dhall-lang/dhall-haskell/blob/a13c6564d3d04ce19cabf85b03067d808a9a0fd6/dhall-openapi/src/Dhall/Kubernetes/Convert.hs#L165-L166

Specifically, you would take the `Map ModelName Expr` returned by `toTypes'` and optionally extend it with a new entry for `NatOrString` by `Data.Map.insert`ing such an entry to extend the `Map`.  Something like:

```haskell
toTypes prefixMap typeSplitter preferNaturalInt natIntExceptions definitions =
    | preferNaturalInt =
        Map.insert "io.k8s.apimachinery.pkg.util.intstr.NatOrString.dhall" natOrStringType
    | otherwise = types
  where
    types = toTypes' prefixMap typeSplitter preferNaturalInt natIntExceptions definitions Data.Map.empty
```

Then I believe if you do that everything downstream will "just work", meaning that it will create a new file under `./types`, re-export that file from `./types.dhall`, and so forth.

arobertn:
Thanks, I got this working. The other piece (which I missed before) was subbing out the ref to IntOrString with one to NatOrString. I elected to generate `...NatOrString.dhall` regardless of the `--prefNatInt` argument; for one thing, you can pass `--natIntExceptions` to selectively Naturalize some Integer or IntOrString fields when `--prefNatInt` is not given.

dhall-openapi/dhall-openapi.cabal
```diff
@@ -1,6 +1,6 @@
 Cabal-Version:  >=1.11
 Name:           dhall-openapi
-Version:        1.0.2
+Version:        1.0.3
```
Gabriel439, line 3: You don't need to bump the version number.  We compute the new version for each package when we publish releases

Gabriel439 (approved):
This looks great to me!  If this works for you then I'm fine merging this

PR #2389, opened 2022-02-19
Simplify dhall-to-nix code

Gabriel439:
… by making use of the available short-hand utilities


Gabriel439:
This also contains a few changes to the implementation of certain utilities:

* `Natural/{even,odd}` now use a much more efficient implementation
* A few unnecessary `let`s were removed

PR #2398, opened 2022-03-08
dhall-lsp-server: 1.0.18 → 1.1.0

Gabriel439:


Gabriel439:
Note that this is a Hackage-only release just for `dhall-lsp-server` to help unblock GHC 9.0 upgrade efforts (such as https://github.com/NixOS/nixpkgs/pull/160733)

PR #2399, opened 2022-03-10
Consolidate GitHub workflows

Gabriel439:
This consolidates the GitHub workflows so that we only have a single
runner per operating system.  The motivation behind this change is to:

* improve the reliability of CI

  Fewer workflows means fewer spurious failures

* improve the speed of CI

  Fewer workflows means less time spent waiting for available runners


sjakobi (approved):
Nice! :) This looks much less wasteful than the old setup.

amesgen:
Just wanted to update https://github.com/justinwoo/easy-dhall-nix to 1.41.2, but stumpled upon dhall-nix no longer being packaged on Linux. Do you remember why this was done here?

Gabriella439:
It's a mistake of the release process.  You're right that it should be there and I'll look into it this evening

PR #2404, opened 2022-03-18
dhall-lsp-server: Support newer dependency versions

Gabriel439:
The main change is that in newer versions of `lsp-types` there is a `UInt` instead of an `Int` in various places.  However, the code is still compatible with older versions through the use of `fromIntegral` since both `UInt` and `Int` implement `Integral`


sjakobi (approved):
These conversions with `fromIntegral` can in principle wrap around and truncate, but I guess that's unlikely to happen in practice here:

```
ghci> fromIntegral @Int @UInt maxBound
2147483647
ghci> fromIntegral @Int @UInt (-1)
2147483647
```

PR #2412, opened 2022-04-20
'dhall text' support passing shell functions as `Text -> Text` arguments

mstksg:
Context/discussion: https://discourse.dhall-lang.org/t/idea-for-dhall-text-tool-accept-arguments-including-shell-commands-as-text-text/568

For my use case, the dhall tool alone is probably 90% of the way to a full templating solution with multiple backends.

Currently, `dhall text` requires the expression to be `Text`.  But what if it was able to also render expressions of type `(Text -> Text) -> Text`, and be given a shell argument as the `Text -> Text` ?

```dhall
-- testfile.dhall
let text = https://raw.githubusercontent.com/dhall-lang/dhall-lang/v21.1.0/Prelude/Text/package.dhall
in  \(f : Text -> Text) -> text.concatMapSep "," Text f [ "hello", "world" ]
```

Would give:

```
$ dhall text --file testfile.dhall --argCmd cat
hello,world
$ dhall text --file testfile.dhall --argCmd "tr '[:lower:]' '[:upper:]'"
HELLO,WORLD
$ dhall text --file testfile.dhall --argCmd "pandoc -f markdown -t html"
<p>hello</p>
,<p>world</p>
$ dhall text --file testfile.dhall --argCmd "md5sum -z"
5d41402abc4b2a76b9719d911017c592  -,7d793037a0760186574b0282f2f435e7  -
```

Error messages:

```
$ dhall text --file testfile.dhall

- Text
+ .. -> .. (a function type)
$ dhall text --file testfile.dhall --argCmd cat --argCmd cat

- .. -> ..  (a function type)
+ Text
```

Supports multiple arguments as well:

```dhall
-- testfile2.dhall
let text = https://raw.githubusercontent.com/dhall-lang/dhall-lang/v21.1.0/Prelude/Text/package.dhall
in  \(f : Text -> Text) -> \(g : Text -> Text) -> text.concatMapSep "," Text f [ "hello", g "world" ]
```

```
$ dhall text --file testfile2.dhall --argCmd cat --argCmd "tr '[:lower:]' '[:upper:]'"
hello,WORLD
```

Let me know if this is too "heavy" or "opinionated" to make sense to have in the main tool!  I definitely don't want to add anything to the tool that would compromise the spirit and purpose of it.  I can probably also make this an external tool as well if that's the case.

### DIFF
--- a/dhall/dhall.cabal
+++ b/dhall/dhall.cabal
@@ -238,6 +238,7 @@ Common common
         prettyprinter               >= 1.7.0    && < 1.8 ,
         prettyprinter-ansi-terminal >= 1.1.1    && < 1.2 ,
         pretty-simple                              < 4.2 ,
+        process                     >= 1.6.0    && < 1.7 ,
         profunctors                 >= 3.1.2    && < 5.7 ,
         repline                     >= 0.4.0.0  && < 0.5 ,
         serialise                   >= 0.2.0.0  && < 0.3 ,
--- a/dhall/src/Dhall/Main.hs
+++ b/dhall/src/Dhall/Main.hs
@@ -3,6 +3,7 @@
 -}
 
 {-# LANGUAGE ApplicativeDo     #-}
+{-# LANGUAGE BangPatterns      #-}
 {-# LANGUAGE LambdaCase        #-}
 {-# LANGUAGE NamedFieldPuns    #-}
 {-# LANGUAGE OverloadedStrings #-}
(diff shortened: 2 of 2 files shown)


MonoidMusician:
This looks cool! One thing I think should be discussed is the fact that shell commands are not necessarily pure, and Dhall assumes functions are pure. Particularly during evaluation you could run into issues with function “calls” being duplicated, and with some commands that could cause chaos due to effects being run repeatedly. So I could definitely see it being used in an unsafe capacity, as in “trust me – these shell commands are pure”, but in general it is not theoretically sound to treat shell commands as functions in Dhall.

mstksg:
In terms of unsafety, this all happens at the "final exit point", and so the responsibility is up to the invoker to ensure that the shell command behaves appropriately and cannot be abused.

But, I agree that's a tenuous line, and it does make me uneasy in ways I can't quite place.  Working through this PR, though, has shown me that this is easy enough to write as a standalone executable, so I might just do that!

My end goal is to not have to write a specialized tool for every single use case, but a general tool, and I think having a standalone executable that shells out commands might work well for me in the end without compromising the intent of the project.  Although this still doesn't solve the issue about it being "weird" to have a shell script take the place of a pure dhall function.  I guess in my mind I can think of it as FFI?

mstksg:
i did it :D https://github.com/mstksg/dhall-text-shell

Issue #2414, opened 2022-04-28
dhall-to-nix: malformed nix record keys 


Profpatsch:
```
> nix-instantiate --eval --strict -E $(echo '{ `text/*` = "abc" }' | dhall-to-nix)
```

```
error: syntax error, unexpected '/', expecting '.' or '='

       at «string»:1:7:

            1| { text/* = "abc"; }
             |       ^
```

generated nix code was:

```
> echo '{ `text/*` = "abc" }' | dhall-to-nix
{ text/* = "abc"; }
```

But it should be:

```
{ "text/*" = "abc"; }
```

This looks like a bug in `hnix`’s pretty printer, which does not surprise me:

https://hackage.haskell.org/package/hnix-0.14.0.8/docs/src/Nix.Pretty.html#prettyKeyName

```haskell
[prettyKeyName](https://hackage.haskell.org/package/hnix-0.14.0.8/docs/src/Nix.Pretty.html#prettyKeyName) :: NKeyName (NixDoc ann) -> Doc ann
prettyKeyName (StaticKey "") = "\"\""
prettyKeyName (StaticKey key) | HashSet.member key reservedNames = "\"" <> pretty key <> "\""
prettyKeyName (StaticKey  key) = pretty key
prettyKeyName (DynamicKey key) =
  runAntiquoted
    (DoubleQuoted [Plain "\n"])
    prettyString
    (\ x -> "${" <> withoutParens x <> "}")
    [key](https://hackage.haskell.org/package/hnix-0.14.0.8/docs/src/Nix.Pretty.html#local-6989586621680054214)
```

dhall generates a `StaticKey` via [attrsE](https://hackage.haskell.org/package/hnix-0.14.0.8/docs/Nix-Expr-Shorthands.html#v:attrsE), and that is not correctly escaped.

PR #2416, opened 2022-04-29
Fix dhall-to-nix encoding of nix keys with special symbols

Profpatsch:
hnix assumes that the parser will do “the right thing” and only use
`StaticKey` for keys that do not contain any special symbols, and the
right `DynamicKey` for everything else.

For example, the dhall

```dhall
{ `text/*` = "foo" }
```

would be encoded as the nix

```nix
{ text/* = "foo"; }
```

which fails the nix parser at `/`.

Instead it should be quoted, like this:

```nix
{ "text/*" = "foo"; }
```

I believe this is a bug in the hnix pretty printer, which should
correctly escape special symbols (for the reference, it does so for
`StaticKey`s that are nix keywords).

But anyway, we can work around this by always constructing a
`DynamicKey` with simple double quotes instead.

I opted against doing some heuristics to encode some “simple” keys
without code, because this should be the task of hnix, not of dhall.

Fixes https://github.com/dhall-lang/dhall-haskell/issues/2414

### DIFF
--- a/dhall-nix/src/Dhall/Nix.hs
+++ b/dhall-nix/src/Dhall/Nix.hs
@@ -120,7 +120,9 @@ import Dhall.Core
 import Nix.Expr
     ( Antiquoted (..)
     , NExpr
-    , NExprF (NStr)
+    , NExprF (NStr, NSet)
+    , NRecordType (NNonRecursive)
+    , Binding (NamedVar)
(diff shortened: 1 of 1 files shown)


Gabriel439:
Sorry for the delay!  Thanks for doing this 🙂

PR #2422, opened 2022-06-10
fix dhall-lsp-server/default.nix to build lsp

SnO2WMaN:
I fixed dhall-lsp-server/default.nix that was trying to build dhall instead of dhall-lsp-server.

P.S. The problem with https://github.com/dhall-lang/vscode-dhall-lsp-server/issues/37 is probably resolved in the current modified build. I haven't investigated it in detail, so it may not be exactly the same.


Gabriella439:
Thank you for fixing that!

Issue #2425, opened 2022-06-18
dhall-nix Use z-encoding for nix symbol names


Profpatsch:
When trying to encode a function like 

```
\(Foo/Bar: Text) -> Foo/Bar
```

I get the nix code

```nix
Foo/Bar:
  Foo/Bar
```

which fails with the nix parsing error

```
error: syntax error, unexpected ':', expecting end of file

       at «string»:1:8:

            1| Foo/Bar:
             |        ^
```

… this is not good, because “namespaced” symbols like that are quite frequent in dhall code.

I would suggest using the [Z-Encoding scheme](https://hackage.haskell.org/package/zenc-0.1.2/docs/Text-Encoding-Z.html) that is also used by GHC to create nix symbols.

Gabriella439:
You can escape identifiers using backticks, like this:

```dhall
λ(`Foo/Bar` : Text) → `Foo/Bar`
```

Profpatsch:
Yeah, but the generated nix code is still not valid nix syntax, this generates

```
Foo/Bar:
  Foo/Bar
```

PR #2426, opened 2022-06-19
Fix dhall-to-nix encoding of symbols with special keys

Profpatsch:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/2425


Profpatsch:
I have to say that I pretty much implemented this blindly, I don’t know if there’s any golden tests that have to be corrected now, or if I missed any symbols in the AST.

Profpatsch:
I kept as closely to the GHC version of z-encoding as possible, most of the code is copied verbatim (except I switched the function to `Text` and added a check that skips the conversion to `String` if there is no characters that have to be escaped).

Profpatsch:
Oh, another problem I found is that sum keys can be the same nearly arbitrary symbols in Dhall, but the nix representation of anonymous function keys can not:

The dhall

```dhall
<Foo | Bar/Baz>.Bar/Baz
```

generates the nix code

```nix
{ Bar/Baz, Foo }:
  Bar/Baz
```

but nix function attribute arguments cannot be quoted, and cannot contain `/` for example. So we are gonna have to Z-Encode every union key in the generated nix.

Profpatsch:
I added a second commit which should fix this problem.

Now only function arguments and let-bindings should be z-encoded anymore, because those can’t really be reduced any further. Maybe there could be an encoding via a record field, but that would impact performance.

Profpatsch:
I haven’t found a way to make a `let` binding appear in the nix code though, maybe that is not even possible? They always normalize out in my examples and never appear in the beta-reduced output.

Profpatsch:
And another one, we also need to quote field selectors.

dhall-nix/src/Dhall/Nix.hs
```diff
@@ -738,3 +728,130 @@ dhallToNix e =
...
+--         alphabetic character x
+--
+--   * The others translate as 'znnnU' where 'nnn' is the decimal number
+--         of the character
+--
+-- @
+--         Before          After
+--         --------------------------
+--         Trak            Trak
+--         foo_wib         foozuwib
```
ocharles, line 784: `foo_wib` isn't a great example because that's not actually what happens in this module. Maybe drop this one?

Profpatsch: Replaced with `foo-wib`.

dhall-nix/src/Dhall/Nix.hs
```diff
@@ -738,3 +728,130 @@ dhallToNix e =
...
+needsEncoding :: Char -> Bool
+needsEncoding 'Z' = True
+needsEncoding 'z' = True
+needsEncoding c   = not
+                  ( c >= 'a' && c <= 'z'
+                  || c >= 'A' && c <= 'Z'
+                  || c >= '0' && c <= '9' )
+
+-- If a digit is at the start of a symbol then we need to encode it.
+-- Otherwise package names like 9pH-0.1 give linker errors.
```
ocharles, line 817: I understand this was taken from GHC, but this comment makes no sense in the context of Dhall.

ocharles:
I don't really use Dhall anymore, so don't have much input here. It looks generally fine, but I haven't given it a thorough review. I'm probably not the right person to be reviewing, so I'm untagging myself.

dhall-nix/src/Dhall/Nix.hs
Profpatsch: Ah, yeah, but it also isn’t a valid nix symbol, I’ll change the docstring.

Profpatsch:
I totally forgot about this; @Gabriella439 I think I addressed your comments, PTAL

dhall-nix/src/Dhall/Nix.hs
```diff
@@ -738,3 +728,129 @@ dhallToNix e =
...
+-- @{ foo/bar = 42; }.foo/bar@ is not syntactically valid nix.
+--
+-- This is only done if necessary (where “necessary” is not super defined right now).
+mkDoubleQuotedIfNecessary :: Text -> NKeyName r
+mkDoubleQuotedIfNecessary key =
+    if Text.all simpleChar key
+    then StaticKey (VarName key)
+    else DynamicKey (Plain (DoubleQuoted [Plain key]))
+    where
+        simpleChar c = isAsciiLower c || isAsciiUpper c
```
Gabriella439, line 761: Minor suggestion: I think this makes the intention more clear

```suggestion
        simpleChar c = isAscii c || isLetter c
```

This would also require a matching change to the imports

dhall-nix/src/Dhall/Nix.hs
```diff
@@ -738,3 +728,129 @@ dhallToNix e =
...
+        encodeDigitChar c
+        <> Text.concatMap encodeChar cs'
+
+-- | Whether the given characters needs to be z-encoded.
+needsEncoding :: Char -> Bool
+needsEncoding 'Z' = True
+needsEncoding 'z' = True
+needsEncoding c   = not
+                  ( isAsciiLower c
+                  || isAsciiUpper c
```
Gabriella439, line 813: Perhaps reuse the `simpleChar` function here

dhall-nix/src/Dhall/Nix.hs
```diff
@@ -738,3 +728,129 @@ dhallToNix e =
...
+--         Trak            Trak
+--         foo-wib         foozmwib
+--         \>               zg
+--         \>1              zg1
+--         foo\#            foozh
+--         foo\#\#           foozhzh
+--         foo\#\#1          foozhzh1
+--         fooZ            fooZZ
+--         :+              ZCzp
+-- @
```
Gabriella439, line 799: If you begin each line with `> ` then you don't need to escape special characters like `\#`

Gabriella439 (approved):
Looks great!  Just some tiny suggestions

dhall-nix/src/Dhall/Nix.hs
Profpatsch: I … don’t think all ascii values are a valid nix symbol. It includes stuff like 0x04 (Ctrl+c); and I’d have to verify if they are allowed to start with a digit

Profpatsch: For example

```
nix-repl> { 1 = "foo"; }
error: syntax error, unexpected INT

       at «string»:1:3:

            1| { 1 = "foo"; }
             |   ^
            2|
```

Gabriella439: Whoops, I meant to say `isAscii c && isLetter c`

dhall-nix/src/Dhall/Nix.hs
Profpatsch: Hm, I’d have to think about whether they are semantically the same.

dhall-nix/src/Dhall/Nix.hs
Profpatsch: But `isAscii` still includes all control characters for example

Gabriella439: The additional `isLetter` requirement would exclude those control characters

Profpatsch: i.e. 

```
-- | Selects the first 128 characters of the Unicode character set,
-- corresponding to the ASCII character set.
isAscii :: Char -> Bool
isAscii c = c < '\x80'
```

Gabriella439: Please read my responses more closely

Profpatsch: Oh, using `&&` … at least to me filtering out by boolean and is not more intuitive than set logic.

Gabriella439: The main reason I suggested using `isLetter` is that it seems indirect to specify that something is a letter by saying that it is uppercase or lowercase.  What you've written reads to me like:

```haskell
simpleChar c = (isAscii c && isUpper c) || (isAscii c && isLower c)
```

dhall-nix/src/Dhall/Nix.hs
Profpatsch: Indeed, thinking about it they are not the same, since we either double quote where necessary (using the `simpleChar` filter), or z-encode where necessary (internal symbols that cannot be quoted). In particular, z-encoded values should never land inside a struct for example.

I think this made me notice a place where I’d forgotten the encoding, namely `Project`.

Gabriella439: What I mean is that right now `simpleChar c` is defined as `isAsciiLower c || isAsciiUpper c`, so it seems like you should be able to replace `isAsciiLower c || isAsciiUpper c` right here with `simpleChar c`

Profpatsch:
The fix unfortunately does not apply anymore. I don’t know if it was fixed on HEAD in the meantime

Gabriella439:
This is not fixed in `HEAD`

Profpatsch:
Rebased on master

dhall-nix/src/Dhall/Nix.hs
Profpatsch: Anyway, coming back to this, apart from the aesthetics, `isLetter` needs support for unicode ICU (it’s not just a simple `ord` check):

```
isLetter :: Char -> Bool
isLetter c = case generalCategory c of
 GeneralCategory
 UppercaseLetter -> True
 LowercaseLetter -> True
 TitlecaseLetter -> True
 ModifierLetter -> True
 OtherLetter -> True
 _ -> False

generalCategory :: Char -> GeneralCategory
generalCategory c = toEnum $ fromIntegral $ wgencat $ fromIntegral $ ord c

foreign import ccall unsafe "u_gencat"
 wgencat :: Int -> Int
```

vs 

```
isAsciiLower :: Char -> Bool
isAsciiLower c = c >= 'a' && c <= 'z'
```

dhall-nix/src/Dhall/Nix.hs
Gabriella439: This is the change I'm referring to:

```suggestion
-- >        Before          After
-- >        --------------------------
-- >        Trak            Trak
-- >        foo-wib         foozmwib
-- >        >               zg
-- >        >1              zg1
-- >        foo#            foozh
-- >        foo##           foozhzh
-- >        foo##1          foozhzh1
-- >        fooZ            fooZZ
-- >        :+              ZCzp
```

dhall-nix/src/Dhall/Nix.hs
Gabriella439: I still feel like you can reuse the `simpleChar` function here:

```haskell
needsEncoding c = not (simpleChar c)
```

Profpatsch:
Is this intended to stay broken?

PR #2429, opened 2022-06-28
Update dhall-lang reference

Gabriella439:
… and fix/ignore the test failures


Gabriella439:
closing as this is going to be supersedd by https://github.com/dhall-lang/dhall-haskell/pull/2493

PR #2433, opened 2022-07-15
Marshal Data.Text.Short.ShortText

j6carey:
...so that proto3-suite can use it as a string
representation within generated Haskell code
while still supporting Dhall marshaling.


sjakobi (approved):
We already depend on `text-short` transitively via `aeson`, so the additional direct dependency doesn't seems acceptable to me.

Gabriella439:
Thank you! 🙂

j6carey:
Thanks for the review, and you're welcome!

PR #2435, opened 2022-07-24
Version 1.41.1 → 1.41.2

Gabriella439:


sjakobi (approved):
I guess some of the version bumps and releases aren't strictly necessary, because e.g. the additional support for `vector-0.13` has already been published in the form of Hackage revisions. But they probably don't hurt much either, apart from increasing the size of the package index a bit.

PR #2445, opened 2022-09-04
Fix my deadname

Gabriella439:


nix/shared.nix
```diff
@@ -236,7 +236,7 @@ let
                   [ (pkgsNew.haskell.lib.packagesFromDirectory { directory = ./packages; })
                     extension
                     doCheckExtension
-                    doBenchmarkExtension
+                    # doBenchmarkExtension
```
SiriusStarr, line 239: Should this be a separate PR?  Wasn't sure if you'd meant to include this change.

Gabriella439: Yeah, that's a mistaken change that I accidentally committed.  I'll remove that

PR #2446, opened 2022-09-06
Compile merging Optionals correctly in dhall-nix

### LINKED ISSUE #2443: [dhall-nix] merge operator on Optional not transpiled correctly
I’m currently working on creating bindings for nix library functions in dhall so that you can write more of your NixOS configuration in dhall and I have run into some friction surrounding the merge operator, as it is not translated correctly.

```dhall
let Optional/default =
      ./default.dhall
        sha256:5bd665b0d6605c374b3c4a7e2e2bd3b9c1e39323d41441149ed5e30d86e889ad

in  λ(v : Optional Natural) → Optional/default Natural 0 v
```

is translated as 

```nix
v:
  v { None = 0; Some = _: _; }
```

and not as the working

```nix
v: if v != null then (_: _) v else 0
```

It generates the same code as normal unions, which is inappropriate because `Some v` is mapped to `v` and `None t` is mapped to `null`.

ghost:
This is done by checking the type of the merge input, and emitting special code for Optionals.

This commit fixes #2443.

Please let me know if i did something poorly, i do not know haskell.

### DIFF
--- a/dhall-nix/src/Dhall/Nix.hs
+++ b/dhall-nix/src/Dhall/Nix.hs
@@ -99,7 +99,6 @@ import Data.Fix (Fix (..))
 import Data.Foldable (toList)
 import Data.List.NonEmpty (NonEmpty(..))
 import Data.Text (Text)
-import Data.Traversable (for)
 import Data.Typeable (Typeable)
 import Data.Void (Void, absurd)
 import Lens.Family (toListOf)
@@ -111,11 +110,12 @@ import Dhall.Core
     , Expr (..)
     , FieldSelection (..)
     , FunctionBinding (..)
-    , MultiLet (..)
     , PreferAnnotation (..)
     , Var (..)
     , WithComponent (..)
     )
+import Dhall.Parser  (Src)
+import Dhall.Context (Context)
 
 import Nix.Expr
     ( Antiquoted (..)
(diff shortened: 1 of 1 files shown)


Gabriella439:
Sorry for the long delay in responding to this.

I think there is a simpler approach that doesn't require threading through the context or inferring the type of any expressions.  Rather, what you can do is to check if the argument to the `Merge` expression is of the form `Some _` or `App None _` and then change the translation.  This makes use of the fact that the expression passed to `loop` is already pre-normalized so if it's an `Optional` value then it can only be one of those two cases

I can also implement my suggestion myself if you want, especially given how delayed my review was

DarkKirb:
#2443 is only an issue for merges where the expression cannot be reduced to a form `merge {Some = _, None = _} [Some _|None _]`.

some cases like `λ(a: Natural) → merge { None = 0, Some = λ(a: Natural) → a }  (Some a)` compiles correctly, however `λ(a: {None: Natural, Some: Natural → Natural }) → merge a (Some 1)` also results in the nonsensical `a: 1 a` (i think that would be fixed with your suggestion though)

Gabriella439:
What about in the general case translating the `merge` keyword to something like:

```
merge handlers unionLiteral

=>

if builtins.attrNames handlers == [ "Some", "None" ]
then … special case for Optional …
else … normal case …
```

PR #2448, opened 2022-09-18
Enable translation of maps into directory trees

nikita-volkov:
Resolves https://github.com/dhall-lang/dhall-lang/issues/1305

### DIFF
--- a/dhall/src/Dhall/DirectoryTree.hs
+++ b/dhall/src/Dhall/DirectoryTree.hs
@@ -31,7 +31,7 @@ import qualified System.FilePath             as FilePath
 
     * Records are translated into directories
 
-    * @Map@s are also translated into directories
+    * @Map@s are translated into directory trees
 
     * @Text@ values or fields are translated into files
 
@@ -124,13 +124,20 @@ toDirectoryTree path expression = case expression of
         empty
 
     process key value = do
-        if Text.isInfixOf (Text.pack [ FilePath.pathSeparator ]) key
-            then die
-            else return ()
+        (dirPath, fileName) <- case reverse keyPathSegments of
+            h : t ->
+                return
+                  ( Foldable.foldl' (</>) path (reverse t)
+                  , h )
+            _ ->
+                die
 
-        Directory.createDirectoryIfMissing False path
+        Directory.createDirectoryIfMissing True dirPath
 
-        toDirectoryTree (path </> Text.unpack key) value
+        toDirectoryTree (dirPath </> fileName) value
+      where
+        keyPathSegments =
+            fmap Text.unpack $ Text.splitOn "/" key
 
     die = Exception.throwIO FilesystemError{..}
       where


nikita-volkov:
Thought. It might make sense to add a check on paths not being absolute and not containing `..` segments to ensure that it's impossible to affect the filesystem outside of the working dir.

Opinions?

nikita-volkov:
I went ahead and added the according security checks.

mmhat:
@nikita-volkov I do have a pending PR https://github.com/dhall-lang/dhall-haskell/pull/2437 with another possibility of constructing a directory tree and I am interested in you feedback there. Would it help you with your use case?
Also, I like your ideas of prohibiting some file paths by default and I think those are valuable on there own.

mmhat:
> @mmhat I've taken a look at [the usage example](https://github.com/mmhat/dhall-haskell/blob/05a0569e4461a3b68e099bfa1483b6f95b8a8130/dhall/examples/to-directory-tree.dhall) provided in your PR. It looks like your solution provides a finer grain control, which should definitely find use in non-trivial tasks. However it also comes at a price of a more complicated API for the developer.
> 
> I think both solutions should be merged. Mine will be useful for simple cases, providing a simpler API, yours will be useful for advanced cases.

@nikita-volkov Thanks for the feedback; I think its a fair assessment! I also don't see a problem to support both ways and I'd be happy to see it merged.

Gabriella439 (approved):
I will let @nikita-volkov and @mmhat work out how to resolve any merge conflicts that might arise

nikita-volkov:
Trying to resolve the conflicts after the changes by @mmhat. The only issue is with the added `allowSeparators` global option in that change. I don't see an intuitive interpretation of that option with the changes that I've introduced here. Do we really need it? It doesn't seem necessary for the changes introduced in your PR.

nikita-volkov:
Anyway, I've adapted the changes. Now setting `allowSeparators` to False will effectively disable the tree construction using maps as introduced in this PR.

mmhat:
Great, I'll have another look tomorrow :+1:

nikita-volkov:
@mmhat I don't like the idea of parameterising the compiler with various modes of interpretation of one codebase. This will likely lead to all sorts of confusion for the users and will require authors to inform the users about the compiler settings needed to compile their codebase. That's the reason I'm skeptical about the added `allowSeparators` flag in the first place, your PR to my fork adds even more options.

There can only be one truth and IMO it's best for build tools to exhibit the same property: there can only be one interpretation of one codebase. That's what makes builds predictable.

Pinging @Gabriella439 to draw some attention.

mmhat:
I don't really disagree with you on the fact that it is a good thing to have one and only one interpretation of a codebase. My point is more about the defaults here: I see all three checks as safety features guarding against programming errors (e.g. no path separators) or security risks (e.g. no ".."). The CLI flags are simply escape hatches in the case You Know What You Are Doing. So to me its more about disabling safety checks than interpreting the program in a different way.

If we disagree here on the defaults of those settings then I am happy to discuss that, but I'd like a switch for enabling/disabling anyway.

dhall/src/Dhall/DirectoryTree.hs
```diff
@@ -215,12 +222,39 @@ toDirectoryTree allowSeparators path expression = case expression of
...
 
     process key value = do
-        when (not allowSeparators && Text.isInfixOf (Text.pack [ FilePath.pathSeparator ]) key) $
-            die
-
-        Directory.createDirectoryIfMissing allowSeparators path
-
-        toDirectoryTree allowSeparators (path </> Text.unpack key) value
+        case keyPathSegments of
+            -- Fail if path is absolute, which is a security risk.
```
Gabriella439, line 226: Perhaps use `System.FilePath.isRelative`?

dhall/src/Dhall/DirectoryTree.hs
```diff
@@ -215,12 +222,39 @@ toDirectoryTree allowSeparators path expression = case expression of
...
+            h : t ->
+                return
+                    ( Foldable.foldl' (</>) path (reverse t)
+                    , h )
+            _ ->
+                die
+
+        Directory.createDirectoryIfMissing True dirPath
+
+        toDirectoryTree allowSeparators (dirPath </> fileName) value
```
Gabriella439, line 254: I think you want something like this:

```suggestion
        (dirPath', fileName) <- case reverse keyPathSegments of
            h : t ->
                return
                    ( Foldable.foldl' (</>) path (reverse t)
                    , h )
            _ ->
                die

        let dirPath = path </> dirPath'
        
        Directory.createDirectoryIfMissing True dirPath

        toDirectoryTree allowSeparators (dirPath </> fileName) value
```

In other words, `dirPath` needs to incorporate `path` as a prefix, otherwise the `Map` fields will be created as subdirectories relative to `PWD` when they should be relative to `path`.

Note that https://github.com/nikita-volkov/dhall-haskell/pull/1 fixes this, too

Gabriella439 (commented):
I personally think it's fine to add the additional options that @mmhat suggested.  I'd be more concerned if the options changed the shape of the expected directory tree type, but that's not the case here

PR #2449, opened 2022-09-19
Split Dhall.Syntax

### LINKED ISSUE #2313: Hydra sometimes runs out of memory while building `dhall`

mmhat:
This PR splits `Dhall.Syntax` into several smaller modules to mitigate the effect of https://gitlab.haskell.org/ghc/ghc/-/issues/5642 and https://gitlab.haskell.org/ghc/ghc/-/issues/5642 - namely the excessive space usage at built time caused by the `Generic` instance of `Dhall.Syntax.Expr`. Most datatypes from `Dhall.Syntax` were moved to own modules in the `Dhall.Syntax` namespace and the derivation/declaration of type class instances for the datatypes were grouped in modules below `Dhall.Syntax.Instances`. Everything from the new modules gets re-exported by `Dhall.Syntax` and are internal to the `dhall` package; Nothing should change from a users point of view.
The memory consumed during a run of `cabal build dhall` drops from ~1200 MB to ~800 MB and the build time decreases slightly.

I also removed the superfluous `PreferFromWith` data constructor which simplifies the `PreferAnnotation` type. Note that this _is_ a breaking change.

Fixes #2313


Gabriella439 (approved):
Looks great!  I think the breaking change is fine

Feel free to merge once tests pass

PR #2456, opened 2022-09-29
Run GitHub actions on pull requests instead of pushes

Gabriella439:
See:

https://github.com/dhall-lang/dhall-haskell/pull/2455#issuecomment-1261500084


Gabriella439:
I'm going to force this one through if everything passes except the `dhall-lsp-server` tests on Windows so that we can get CI running again for other PRs

PR #2469, opened 2022-11-10
dhall-toml: support integers

lf-:
The docs *said* we support integers, but that seems to not actually have been empirically true (actually, most amusingly, fromInteger is partial so it would crash TomlToDhall).

I fixed it.


Gabriella439:
Thank you! ❤️

PR #2473, opened 2022-11-26
Require `time >= 1.9` for `DayOfWeek`

amesgen:
`DayOfWeek`, which is being used since #2413, has been added in `time-1.9`, but e.g. GHC 8.6.5 has `time-1.8.0.2` as its boot version.

https://hackage.haskell.org/package/time-1.12.2/changelog


Gabriella439:
Thank you!  I published the matching revision

Issue #2472, opened 2022-11-24
Support hnix-0.16


sternenseemann:
Seems like we may need to move to this version in nixpkgs: https://github.com/NixOS/nixpkgs/pull/202688

As far as I can tell `dhall-nix`  doesn't support it yet.

Gabriella439:
Fix is up here: https://github.com/dhall-lang/dhall-haskell/pull/2474

PR #2474, opened 2022-11-26
dhall-nix: Build against `hnix-0.16`

Gabriella439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/2472


sternenseemann:
Thanks, applying this on top of the latest hackage release of dhall-nix seems to work!

PR #2478, opened 2022-12-29
Freeze expressions providing a custom context and normalizer

mmhat:
This PR adds variants for the functions found in the `Dhall.Freeze` module that also take a context and a normalizer. Those functions are prefixed with `custom` in order to maintain backwards compatibility. Code common to the existing functions and the new ones were moved to a bunch of helper functions such that there is no unnecessary code duplication.

### DIFF
--- a/dhall/src/Dhall/Freeze.hs
+++ b/dhall/src/Dhall/Freeze.hs
@@ -8,33 +8,32 @@
 module Dhall.Freeze
     ( -- * Freeze
       freeze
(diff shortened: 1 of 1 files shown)


Gabriella439:
Maybe these could take an `EvaluateSettings` since that has what we need (and some other useful stuff, too)

Gabriella439 (approved):
Perfect!  This looks amazing.  Thank you for doing this

PR #2479, opened 2023-01-05
Change `input` to `inputFile` in tutorial

soenkehahn:
I think `input` doesn't work here, does it?


Gabriella439:
`input` does work!  It's because `input` accepts any Dhall expression and Dhall expressions may import other Dhall files by their paths

soenkehahn:
Ah, thanks for explaining! Closing this.

Gabriella439:
You're welcome! 🙂

Issue #2481, opened 2023-01-26
Can `ExtractError` have an `Eq` instance?


NicolasT:
Currently, the `InvalidDecoder` type doesn't have an `Eq` instance. Hence, the result of `extract auto expr`, which has type `Extractor Src Void a` == `Validation (ExtractErrors Src Void)` == `Validation (DhallErrors (ExtractError Src Void))`, doesn't have an `Eq` instance (even though `Validation` has one for `(Eq e, Eq a) => Eq (Validation e a)` and `DhallErrors` for `Eq e => Eq (DhallErrors e)`).

`ExtractError` is defined as
```haskell
data ExtractError s a =
    TypeMismatch (InvalidDecoder s a)
  | ExpectedTypeError ExpectedTypeError
  | ExtractError Text
```
where `ExpectedTypeError` and `Text` have their proper `Eq` instance, but `InvalidDecoder s a` doesn't. This seems to be a pair of `Expr s a`s which has an `Eq` instance: `(Eq s, Eq a) => Eq (Expr s a)`.

In my case, `s` is `Src` and `a` is `Void`, which both have `Eq` instances.

Hence, is there a reason not to derive an `Eq` for `InvalidDecoder s a` (and then for `ExtractError` as well)?

The reason I'm asking: I'm using Hedgehog's `tripping` property test to ensure `ToDhall` and `FromDhall` instances obey (basically) `\v -> let e = embed inject v in pure e == extract auto e`. Given how `tripping` is defined (`tripping :: (MonadTest m, Applicative f, Show b, Show (f a), Eq (f a), HasCallStack) => a -> (a -> b) -> (b -> f a) -> m () `), I need to throw out the errors (since there's no `Eq` instance for `f a`) by taking the result of `extract auto b`, turning it into `Maybe a`, discarding any potential errors (which would otherwise end up in Hedgehog's error message).

Gabriella439:
There's no particular reason why it's missing an `Eq` instance.  I'd accept a PR to add it

PR #2482, opened 2023-01-26
dhall: derive `Eq` instances for `InvalidDecoder` and `ExtractError` (#2481)

NicolasT:
Fixes: https://github.com/dhall-lang/dhall-haskell/issues/2481


mmhat:
Looks good to me; Thank you @NicolasT !

PR #2483, opened 2023-01-28
Require prettyprinter-ansi-terminal >= 1.1.2

Bodigrim:
`Prettyprinter.Render.Terminal` is not available before `prettyprinter-ansi-terminal-1.1.2`.


Bodigrim:
As a Hackage trustee I revised `dhall-1.40.2..1.41.1` accordingly.

PR #2486, opened 2023-02-01
dhall freeze: Remove cache import expression before a secure freeze

mmhat:
Doing so allows us to protected imports securely that were previously only protected for caching purposes. For example, given the following Dhall file `foo.dhall`:
```dhall
  missing
    sha256:9bb9dcb5bf6f795291686f59383bcd01c8e79b87fc3fb63351d46dea100ac51b
? ./bar.dhall
```

Previously `dhall freeze --all foo.dhall` threw an "No valid imports" error as it was failing to add a hash to the `missing` import. With the changes in this PR freezing `foo.dhall` results in the following expected content of that file:
```dhall
./bar.dhall
  sha256:9bb9dcb5bf6f795291686f59383bcd01c8e79b87fc3fb63351d46dea100ac51b
```


Gabriella439 (approved):
Perhaps add a test, but otherwise this looks great!

PR #2487, opened 2023-02-01
Added dhall package command

### LINKED ISSUE #1645: Automated packaging (`dhall package`)
I was going to write a script to do this, but thinking about it, I'm wondering if it should be included in `dhall-haskell`.

The standard for the prelude (and several community packages it seems) is to have a file in each folder named `package.dhall` that consists of a record of frozen imports of all files in that folder and subfolders/etc.  For the Prelude, at the very least, maintaining this is easy but inconvenient (and leads to a lot of failing builds).

As such, I'm (idly) proposing a new command, `dhall package` to generate such files.  `dhall package` would be called on a folder, e.g.
```
Prelude
├── Bool
│   ├── and
│   └── show.dhall
├── Integer
│   ├── toDouble
│   └── toNatural
```
and generate
```
Prelude
├── Bool
│   ├── and
│   ├── package.dhall
│   └── show.dhall
├── Integer
│   ├── package.dhall
│   ├── toDouble
│   └── toNatural
├── package.dhall
```

**Behavior**:

*Files Included*: By default, `*.dhall[b]` and extensionless files.  Packaging fails with an error message of any such files fail to typecheck/hash/etc. (e.g. are not valid Dhall files).  (Perhaps there should be an ignore flag, so that a non-dhall, extensionless file could be ignored?  Or a flag to only consume files with the explicit Dhall extensions?)

*Extant `package.dhall` Files*:  Any new entries, removed entries, or changed hashes should be logged in the terminal output.  Perhaps should ask for confirmation in such cases (with a flag to force)?  Alternately, could include a format indicator comment at the start of `package.dhall` files so-packaged, e.g.
```
{- Packaged by dhall package vX.XX -}
```
and automatically overwrite such `package.dhall`s but abort if there's a `package.dhall` lacking that (since it was presumably made by hand then). 

Anyways, as I said, somewhat idle thought.

mmhat:
Fixes #1645

### DIFF
--- a/dhall/dhall.cabal
+++ b/dhall/dhall.cabal
@@ -332,6 +332,7 @@ Library
         Dhall.Marshal.Encode
         Dhall.Map
         Dhall.Optics
+        Dhall.Package
         Dhall.Parser
         Dhall.Parser.Expression
         Dhall.Parser.Token
@@ -420,6 +421,7 @@ Test-Suite tasty
         Dhall.Test.Import
         Dhall.Test.Lint
         Dhall.Test.Normalization
+        Dhall.Test.Package
         Dhall.Test.Parser
         Dhall.Test.QuickCheck
         Dhall.Test.Regression
--- a/dhall/src/Dhall/Main.hs
+++ b/dhall/src/Dhall/Main.hs
@@ -163,7 +163,7 @@ data Mode
     | DirectoryTree { allowSeparators :: Bool, file :: Input, path :: FilePath }
     | Schemas { file :: Input, outputMode :: OutputMode, schemas :: Text }
     | SyntaxTree { file :: Input, noted :: Bool }
-    | Package { output :: Output, inputs :: NonEmpty Input }
+    | Package { name :: Maybe String, files :: NonEmpty FilePath }
 
 -- | This specifies how to resolve transitive dependencies
 data ResolveMode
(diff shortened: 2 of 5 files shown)


Gabriella439:
Sorry for the delay, but I think the command-line interface I would suggest is:

```ShellSession
$ dhall package "${DIRECTORY}"  # Create a `package.dhall` in that directory from its contents
$ dhall packages "${FILES[@]}"  # Create a `package.dhall` from those files if they all share the same directory, otherwise fail
```

… possibly with an optional `--name` option to select a name other than `package.dhall`.  But generally I feel like we should try to promote the convention that the package file is named `package.dhall` and make that as ergonomic as possible.

Another advantage of doing it this way is that you don't need to worry about computing relative paths from the input files to the output package file, so it simplifies the implementation a lot.

mmhat:
Thanks for the feedback; Yes, that would indeed simplify the implementation. I am a bit hesitant to add more than one new top-level command to the CLI - especially since the two you proposed are quite similar. How about:
```bash
$ dhall package directory "${DIRECTORY}"
$ dhall package files "${FILES[@]}"
```

mmhat:
I changed the implementation to allow only files from the same directory and it is closer to the suggested behavior now. I did not yet change the CLI as the discussion about that is ongoing.

mmhat:
No idea why the tests are failing on the Hydra build. They succeed on all other runners and I can't reproduce the failure locally.

Gabriella439:
this usually means that you need to add something to `extra-source-files`: https://github.com/dhall-lang/dhall-haskell/blob/0f1b05c82d8a7a88381f1d217de371499acba251/dhall/dhall.cabal#L26

… because the thing that the Nix build does differently is that it builds `dhall` from the source distribution that `cabal` would have generated and `cabal` will omit files that aren't explicitly mentioned in `extra-source-files` from the source distribution

Gabriella439 (approved):

PR #2496, opened 2023-03-28
Support template-haskell 2.20.0 and unix-compat 0.7

locallycompact:
Caught by https://gitlab.horizon-haskell.net/package-sets/horizon-core/-/jobs/197815

### DIFF
--- a/dhall/dhall.cabal
+++ b/dhall/dhall.cabal
@@ -250,6 +250,7 @@ Common common
         th-lift-instances           >= 0.1.13   && < 0.2 ,
         time                        >= 1.9      && < 1.13,
         transformers                >= 0.5.2.0  && < 0.7 ,
+        unix                        >= 2.7      && < 2.9 ,
         unix-compat                 >= 0.4.2    && < 0.7 ,
         unordered-containers        >= 0.1.3.0  && < 0.3 ,
         uri-encode                                 < 1.6 ,
--- a/dhall/src/Dhall/DirectoryTree.hs
+++ b/dhall/src/Dhall/DirectoryTree.hs
@@ -53,8 +53,8 @@ import qualified Dhall.Util                  as Util
 import qualified Prettyprinter.Render.String as Pretty
 import qualified System.Directory            as Directory
 import qualified System.FilePath             as FilePath
+import qualified System.Posix.User           as Posix
 import qualified System.PosixCompat.Files    as Posix
-import qualified System.PosixCompat.User     as Posix
 
 {-| Attempt to transform a Dhall record into a directory tree where:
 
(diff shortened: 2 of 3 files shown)


locallycompact:
Oh, right. In that case this fix is probably wrong. I'm not sure why PosixCompat.User was removed in 0.7, then. The Changelog doesn't say.

locallycompact:
Regarding https://github.com/haskell-pkg-janitors/unix-compat/issues/2#issuecomment-1487446182 . What is expected of these codepaths on Windows?

Gabriella439:
I'm going to check if the corresponding `dhall` subcommand is broken on windows; if so in the worst case scenario we can just remove support for that subcommand on Windows

Gabriella439:
Yeah, so it looks like some of the functionality that `dhall to-directory-tree` depended on from `System.Posix.Compat` was broken on Windows.  In the common case users don't hit this code path (it's a rarely-used feature of this subcommand) which is probably why nobody noticed this problem just yet.

I think the best approach here is to disable that code path on Windows and I can put up the change for that.

Gabriella439:
Actually, it seems like there are some tests that depended on Unix permissions that were running successfully on Windows.  That seems surprising because as far as I can tell those tests would have transitively run `System.PosixCompat.User.get{User,Group}EntryFromName` and failed:

https://github.com/dhall-lang/dhall-haskell/blob/7b40b1cebb1eb9d16358edf2cbfcafbd2c9efdf8/dhall/tests/Dhall/Test/DirectoryTree.hs#L32

cc: @mmhat: I wanted to know if you had any insight here.

mmhat:
Ok, the test didn't fail because we never tried to actually set the user or group in `fixpointedUserGroup`: This test just checks that we can parse the encoded directory tree and doesn't interpret it. That is due to the restrictions on the various platforms: On Windows, the implementation of `unix-compat` simply doesn't work while on Linux the restriction of `chown(2)` makes it difficult to test. From the man page:

> The owner of a file may change the group of the file to any group of which that owner is a member.

That means, writing a proper test for that would mean that we make assumptions about the groups the user running the test is a member of... Unless I am missing something here.

Since metadata is broken on Windows I think the best is to disable it there and throw an error if the user attempts to set it explicitly: https://github.com/dhall-lang/dhall-haskell/pull/2507

Gabriella439:
@locallycompact: If you merge https://github.com/locallycompact/dhall-haskell/pull/1 into your branch then I think it will fix the build failure on Windows

locallycompact:
Awesome thanks.

Incidentally though this th fix is wrong though because it's 2.11.0 that should be the future-proof bound. However, PvP isn't respected for unreleased commits on ghc master, 2.11.0 doesn't exist yet, so I'm rethinking how to approach future proofing.

I'll resubmit what can be merged.

ysangkok:
@locallycompact Are you still working on this? Dhall was remove from Stackage because of this issue.

locallycompact:
I will tidy this up, I'm a little indisposed atm but by all means feel free to do it before me.

PR #2499, opened 2023-04-01
Implement language support for `Bytes`

Gabriella439:
… as standardized in https://github.com/dhall-lang/dhall-lang/pull/1323


mmhat:
Nice, thanks for the implementation :+1:

Gabriella439:
you're welcome! 😊

Issue #2475, opened 2022-12-09
Questions and possible improvements for Dhall.TH


ndwarshuis:
This might be several issues, but I thought I would leave them together (for now) since they might have overlapping solutions and affect the same small set of functions.

For my specific use-case, I have dhall types that I would like to serve as the source of truth, and I would like to generate haskell types from these using Dhall.TH. 

This approach appears to have several limitations. Some of these might already have existing solutions/workarounds that I missed, but if not they are potential features/enhancements.

### 1. Importing higher-kinded type constructors

I would like to create the following Haskell type:

```haskell
data Foo a = Foo { bar :: a }
```

from the following Dhall code using `makeHaskellTypes`:

```dhall
\(t : Type) -> { bar : a }
```

Unfortunately this doesn't appear to be supported per the two `DataD` clauses in [toDeclaration](https://github.com/dhall-lang/dhall-haskell/blob/49b9b3e3ce1718a89773c2b1bfa3c2af1a6e8752/dhall/src/Dhall/TH.hs#L221) which simply have blank lists for the type variable fields.

I'm assuming it is possible to inspect the Dhall syntax and figure out if there are arguments that need to be added to the type constructor, but I'm also not sure if there are theoretical/technical limitations for why this isn't implemented.

### 2. Support for strictness annotations

Going with the above example again, I don't see a way to add bangs to fields:

```haskell
data Foo a = Foo { bar :: !a }
```

This also appears to be a hardcoded limitation in [toConstructor](https://github.com/dhall-lang/dhall-haskell/blob/49b9b3e3ce1718a89773c2b1bfa3c2af1a6e8752/dhall/src/Dhall/TH.hs#L305) which sets strictness (and unpack) to "no".

The simple way I can think of for implementing this would be to extend `GenerateOptions` (or make a new type like it) which will include a list of fields to which bangs should be applied. I imagine adding `UNPACK` to this would also be useful.

### 3. Coercing Haskell-specific types from more general Dhall Types

Suppose I have the following Dhall type:

```dhall
{ silly : List (List (List Natural)) }
```

and I want to import it as the following Haskell type:

```haskell
data Foo = Foo { silly :: [Vector (Seq Natural)] }
```

The key ideas are a) mapping a Dhall `List` to a specific 'list-like' type in Haskell (or any analogous type pair) and b) doing so at any level within the type as necessary.

[...]

PR #2506, opened 2023-04-09
add higher-kinded type support to TH

ndwarshuis:
addresses other half of https://github.com/dhall-lang/dhall-haskell/issues/2475

Overview of changes to logic: rather than immediately parsing the dhall type into a union/single constructor, first recursively peel back any lambdas at the front of the dhall type and collect these in a bound variable list. Then when constructing the nested data type, add two paths for a) higher kinded haskell types (besides Maybe and List) and b) any of these bound variables from the front of the type.

The main gotcha is that this won't work with built-in From/ToDhall derivations as produced by `GenerateOptions` (or if this is possible I don't know how to do it). In any case the docs are updated to reflect this.


Gabriella439 (approved):
this looks great!  thank you for contributing this!!! 😊

PR #2507, opened 2023-04-13
to-directory-tree: Do not support setting file metadata on Windows

mmhat:
... as it was never supported properly on that platform.


Gabriella439 (approved):
thank you for doing this!

PR #2508, opened 2023-04-14
Improved package command

mmhat:
You can now include files that reside below the output directory in the package. This is for example useful if you want to include sub-packages in a top-level package, i.e:
```
> dhall package toplevel.dhall Sub/package.dhall Other/file.dhall
{ toplevel = ./toplevel.dhall
, Sub = ./Sub/package.dhall
, Other.file = ./Other/file.dhall
}
```

### DIFF
--- a/dhall/src/Dhall/Package.hs
+++ b/dhall/src/Dhall/Package.hs
@@ -136,7 +136,7 @@ getPackagePathAndContent outputFn (path :| paths) = do
 filepathToMap :: FilePath -> FilePath -> Map Text (RecordField s Import)
 filepathToMap outputFn = go [] . splitDirectories
     where
-        go _acc [] = error "filepathToMap: Path is empty"
+        go acc [] = go acc ["."]
         go !acc [x] =
                     let import_ = Import
                             { importHashed = ImportHashed
--- a/dhall/src/Dhall/Package.hs
+++ b/dhall/src/Dhall/Package.hs
@@ -172,8 +172,8 @@ filepathToMap outputFn = go [] . splitDirectories
 --
 mergeMaps :: Map Text (RecordField s Import) -> Map Text (RecordField s Import) -> IO (Map Text (RecordField s Import))
 mergeMaps x y = do
-    let x' = fmap ((:| [])) x
-        y' = fmap ((:| [])) y
+    let x' = fmap (:| []) x
+        y' = fmap (:| []) y
         z = Map.unionWith (<>) x' y'
     for z $ \case
         v@RecordField{recordFieldValue = Embed{}} :| [] -> return v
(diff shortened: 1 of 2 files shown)


dhall/src/Dhall/Package.hs
```diff
@@ -44,72 +47,152 @@ writePackage characterSet outputFn inputs = do
...
+-- > List = ./List/package.dhall
+--
+-- in there and not:
+--
+-- > List = { package = ./List/package.dhall }
+--
+filepathToMap :: FilePath -> FilePath -> Map Text (RecordField s Import)
+filepathToMap outputFn = go [] . splitDirectories
+    where
+        go _acc [] = error "filepathToMap: Path is empty"
```
Gabriella439, line 139:
```suggestion
        go acc [] = go acc ["."]
```

Even though we currently don't hit this case in practice, it's probably worth handling it anyway in case the rest of the code changes.

dhall/src/Dhall/Package.hs
```diff
@@ -44,72 +47,152 @@ writePackage characterSet outputFn inputs = do
...
+        go !acc (x:xs) = Map.singleton (Text.pack x) $ makeRecordField $ RecordLit $ go (Text.pack x : acc) xs
 
-                    let resultMap = if takeFileName p == outputFn'
-                            then Map.empty
-                            else Map.singleton key (makeRecordField $ Embed import_)
-
-                    go (resultMap <> acc) checkOutputDir ps
-                | otherwise -> throwIO $ InvalidPath p
+-- | Merge two 'Map's constructed with 'filepathToMap'.
+-- It will throw an ... error if the arguments are not compatible with each other, e.g.
```
Gabriella439, line 167:
```suggestion
-- It will throw an error if the arguments are not compatible with each other, e.g.
```

dhall/src/Dhall/Package.hs
```diff
@@ -44,72 +47,152 @@ writePackage characterSet outputFn inputs = do
...
+-- we cannot merge the following two maps:
+--
+-- > fromList [ ("file", ./file.dhall) ]
+-- > fromList [ ("file", fromList [("nested", ./file/nested.dhall)]) ]
+--
+mergeMaps :: Map Text (RecordField s Import) -> Map Text (RecordField s Import) -> IO (Map Text (RecordField s Import))
+mergeMaps x y = do
+    let x' = fmap ((:| [])) x
+        y' = fmap ((:| [])) y
```
Gabriella439, line 176:
```suggestion
    let x' = fmap (:| []) x
        y' = fmap (:| []) y
```

Gabriella439 (approved):
Just a few tiny suggestions, but overall this looks great!

PR #2512, opened 2023-04-16
add TimeZone/Date/TimeOfDay to TH

ndwarshuis:
these seemed to be missing from the valid nested type list


Gabriella439 (approved):
thank you for fixing that!!

ndwarshuis:
thank you :)

PR #2513, opened 2023-04-19
Improved TH generation

mmhat:
This PR adds four new data constructors to `Dhall.TH.HaskellType`:

 * `SingleConstructorWith` and `MultipleConstructorsWith` are similar to their counterparts without the "With" suffix. They take an addional `GenerateOptions` value that is used instead of the global one. They are suitable if you have some weird Dhall types which do not translate to Haskell easily and you don't want to apply the modifiers to all other generated types.
 * `Predefined`: Use an existing Haskell type for a Dhall type. Suitable if you want to use `Data.Sequence.Seq SomeType` instead of `[SomeType]` or if you want to translate Dhall maps to Haskell maps.
 * `Scoped`: Limit the scope of the lookup of other Haskell types used in the translation process. Together with `Predefined` this allows you to use e.g. `Data.Sequence.Seq` for some particular types while all the others use the usual Haskell lists. It also allows you to generate two distinct Haskell types that use the same Dhall type as input.


mmhat:
The documentation is not yet as good as it could be; Please have a look at the tests that exemplify the intended use cases until the documentation improves.

winitzki:
Looks like Hydra builds are getting stuck again? Another PR also currently fails its Hydra build but succeeds in other tests.

winitzki:
All tests pass but Hydra build fails. The same thing is happening in another PR: https://github.com/dhall-lang/dhall-haskell/pull/2611

I'm not sure what needs to be done; perhaps Hydra is caching a stale failed build?

PR #2515, opened 2023-04-20
add TH support for nested higher kinded types

ndwarshuis:
This is a follow up to https://github.com/dhall-lang/dhall-haskell/pull/2506 for a complicated case I forgot to add.

In the case where an imported type uses another higher kinded type, the current algorithm will compare this inner type after its arguments have already been applied, in which case the search will fail.

Example:

``` dhall
let InnerS = \(x : Type) -> \(y : Type) -> { ifoo : x, ibar : y }

let InnerU = \(x : Type) -> \(y : Type) -> < Ifoo : x | Ibar : y >

let Outer =
      \(i : Type) ->
      \(j : Type) ->
        { ofoo : InnerS i Text, obar : InnerU j Text }
```

Basic idea is to extend the nested type loop so that when matching a dhall type with an already-cached haskell type, it will store the type parameters for the haskell type constructor, gather the dhall types corresponding to the type parameters, recursively find the constructors for these nested dhall types, and finally apply the results to the nested type constructor.

### DIFF
--- a/dhall/src/Dhall/TH.hs
+++ b/dhall/src/Dhall/TH.hs
@@ -45,6 +45,7 @@ import Language.Haskell.TH.Syntax (DerivClause (..), DerivStrategy (..))
 
 import qualified Data.List                   as List
 import qualified Data.Map                    as Map
+import qualified Data.Maybe                  as Maybe
 import qualified Data.Set                    as Set
 import qualified Data.Text                   as Text
 import qualified Data.Time                   as Time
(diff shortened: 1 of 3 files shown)


Gabriella439:
Sorry for taking so long to get to this

I'm a bit reluctant to merge this because (based on my cursory understanding of how this PR works), this is essentially implementing a form of unification (specifically `gatherNestedVars`) as part of this `TemplateHaskell` subroutine and I'm not sure I want to commit to maintaining a robust unification algorithm in order to support this feature.

PR #2519, opened 2023-05-09
Build against `optparse-generic-1.5.0`

Gabriella439:
Related to https://github.com/commercialhaskell/stackage/issues/6968


felixonmars:
The test failures look irrelevant. dhall-bash builds fine here with the new version.

PR #2524, opened 2023-06-11
Update Stack setup to LTS 19 (GHC 9.0.2)

mmhat:
Greetings from ZuriHac!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,16 +26,16 @@ jobs:
     runs-on: ${{ matrix.os.runner }}
     steps:
       - name: "Checkout repository"
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           submodules: true
       - id: setup-haskell-cabal
         name: "Setup Haskell environment"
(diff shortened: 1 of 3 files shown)


Gabriella439:
hmmm the test failures seem a bit surprising; it's not clear how they relate to the change

mmhat:
> hmmm the test failures seem a bit surprising; it's not clear how they relate to the change

I suspect that the doctest version (0.18.2; previously: 0.17) causes the trouble. I had a quick look at the changelog and the source code diff but it is not clear to me how to fix that. Someone at ZuriHac suggested to switch to cabal-docspec; I'll try that.

mmhat:
Regarding the doctest issue: The upstream bug report is https://github.com/sol/doctest/issues/327.

I tried using cabal-docspec (did not push the code to a branch yet) and while the output is much better it requires to change some tests. I am not sure if that is worth the effort if we we can get the tests running with doctest.

mmhat:
I am closing this in favor of https://github.com/dhall-lang/dhall-haskell/pull/2531. Supporting GHC 9.0 is IMHO not that important if we have a working setup for 9.2.

PR #2525, opened 2023-06-12
CI: Use Cabal instead of Stack

mmhat:


winitzki:
@mmhat What is the advantage of this change? 

BTW this week I have merged a change that fixes CI to use local web servers - this made CI much more stable than before.

PR #2531, opened 2023-08-19
Add Stack setup for GHC 9.2.8

mmhat:
This PR adds a `stack.yaml` for LTS 20, i.e. GHC 9.2.8. This one is now used by default but we run an additional CI job with the LTS 18 setup to ensure that Dhall builds for both GHCs. Additional changes:

 - Bumped GHC 8.10 stack setup to LTS 18.28.
 - Require lsp>=1.5.0.0 in dhall-lsp-server. This made the setup for GHC 9.2.8 easier.
 - Added kind signatures in `Dhall.Syntax.Expr` in order to fix doctests.
 - Updated the use GitHub Actions. In particular, use the new Haskell action (since the old one is [deprecated](https://github.com/haskell/actions/issues/304)).


dhall-lsp-server/dhall-lsp-server.cabal
```diff
@@ -46,22 +46,22 @@ library
...
     , data-default         >= 0.7.1.1  && < 0.8
     , directory            >= 1.2.2.0  && < 1.4
     , dhall                >= 1.38.0   && < 1.43
     , dhall-json           >= 1.4      && < 1.8
     , filepath             >= 1.4.2    && < 1.5
-    , lsp                  >= 1.2.0.0  && < 1.5
-    , rope-utf16-splay     >= 0.3.1.0  && < 0.5
-    , hslogger             >= 1.2.10   && < 1.4
+    , lsp                  >= 1.5.0.0  && < 2
     , lens                 >= 4.16.1   && < 5.2
```
dbaynard, line 57: Would you bump the upper bound on lens, please, too, if it is safe to do so — this is causing breakage in stackage and thus also nixpkgs. The tool does build without the upper bound (I haven't yet tested it).

mmhat:
Apparently Hydra does not like the new `text-rope` dependency: https://hydra.dhall-lang.org/eval/21496896#tabs-errors
I am not really familiar with the Nix setup (and I am also not a Nix user) so a little help here would be highly appreciated.

dhall-lsp-server/dhall-lsp-server.cabal
mmhat: It is safe to do so; I bumped the upper bounds.

Gabriella439:
ah, whoops, it's because of this issue:

https://hydra.dhall-lang.org/jobset/dhall-haskell/2531#tabs-errors

```
error: anonymous function at …/default.nix:1:1 called without required argument 'text-rope'
```

If you do `cabal2nix cabal://text-rope > ./nix/packages/text-reop.nix` I think that should fix it or at least give you a new error.  I'd do it myself but I don't have push rights to your branch (also, for future reference, you can push your branches to this repository if you want me to be able to push fixes to them for you).

Vekhir:
@mmhat There was a possible solution suggested by @Gabriella439, did you investigate if it solves the issue, so this PR can be merged?

Also, this PR doesn't fully address the tests, where lsp-test < 0.15 is specified (0.16 is the newest version). Have you looked at that? Else I'd do that.

mmhat:
@Vekhir Thanks for the ping! I think lsp-test < 0.15 is fine: All lsp-test version that are newer require lsp >= 2.0 which we exclude in our version constraints; IMHO we should relax the lsp-test upper bound once we moved to lsp >= 2.0.

@Gabriella439 Thanks a lot for the hint; That totally worked out! I took the liberty to merge the branch and adjust the required GitHub status checks accordingly. I also deleted the stale Travis CI webhook that didn't work anyway.

PR #2532, opened 2023-08-25
Support GHC 9.4

lrworth:
A more conservative approach than https://github.com/dhall-lang/dhall-haskell/pull/2496. Older versions of libraries should still work, and the behaviour on windows should not have changed at all (except for the error string).

I was able to build `dhall` with GHC 8.10 and 9.4; I don't have time to update the other packages.

### DIFF
--- a/dhall/src/Dhall/DirectoryTree.hs
+++ b/dhall/src/Dhall/DirectoryTree.hs
@@ -55,8 +55,12 @@ import qualified Prettyprinter               as Pretty
 import qualified Prettyprinter.Render.String as Pretty
 import qualified System.Directory            as Directory
 import qualified System.FilePath             as FilePath
+#ifdef mingw32_HOST_OS
+import System.IO.Error           (illegalOperationErrorType, mkIOError)
+#else
+import qualified System.Posix.User           as Posix
+#endif
 import qualified System.PosixCompat.Files    as Posix
-import qualified System.PosixCompat.User     as Posix
 
 {-| Attempt to transform a Dhall record into a directory tree where:
 
@@ -264,7 +268,13 @@ makeType = Record . Map.fromList <$> sequenceA
 -- | Resolve a `User` to a numerical id.
 getUser :: User -> IO UserID
 getUser (UserId uid) = return uid
-getUser (UserName name) = Posix.userID <$> Posix.getUserEntryForName name
+getUser (UserName name) =
+#ifdef mingw32_HOST_OS
+    ioError $ mkIOError illegalOperationErrorType x Nothing Nothing
+    where x = "System.Posix.User.getUserEntryForName: not supported"
+#else
+    Posix.userID <$> Posix.getUserEntryForName name
+#endif
 
 -- | Resolve a `Group` to a numerical id.
 getGroup :: Group -> IO GroupID
--- a/dhall/src/Dhall/DirectoryTree.hs
+++ b/dhall/src/Dhall/DirectoryTree.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP               #-}
 {-# LANGUAGE OverloadedLists   #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards   #-}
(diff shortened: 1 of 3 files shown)


dhall/src/Dhall/DirectoryTree.hs
```diff
@@ -54,8 +55,12 @@ import qualified Prettyprinter               as Pretty
 import qualified Prettyprinter.Render.String as Pretty
 import qualified System.Directory            as Directory
 import qualified System.FilePath             as FilePath
+#ifdef mingw32_HOST_OS
```
lrworth, line 58: This is exactly the condition used by `unix-compat` prior to 0.7: https://github.com/haskell-pkg-janitors/unix-compat/commit/87ffe0554e85875780cedca610a9accc9c23cd4c#diff-93ed634b109420481984ea0fecc52a958c1d98c2728f8ec0b4c6ed741a8d4a24L36

dhall/src/Dhall/DirectoryTree.hs
```diff
@@ -409,57 +420,57 @@ instance Show FilesystemError where
...
-          \    ┌───────────────────────────────────────────┐                                  \n\
-          \    │ { directory = { `example.txt` = \"ABC\" } } │                                \n\
-          \    └───────────────────────────────────────────┘                                  \n\
-          \                                                                                   \n\
-          \                                                                                   \n\
-          \You tried to translate the following expression to a directory tree:               \n\
-          \                                                                                   \n\
-          \" <> Util.insert unexpectedExpression <> "\n\
-          \                                                                                   \n\
+          Util._ERROR <> ": Not a valid directory tree expression                             \n\\
```
lrworth, line 423: The GHC documentation recommends appending a space so that cpp doesn't fall over, but that causes a different error. Adding a slash like this seems to work. Perhaps I should tell the GHC maintainers.

lrworth:
All the check failures seem to be related to a 504 when fetching https://httpbin.org/user-agent. I don't know how to resolve that.

Gabriella439:
Sorry for the delay on this.  I've put up a PR to disable the httpbin.org-related tests:

https://github.com/dhall-lang/dhall-haskell/pull/2533

… and once that merges then it will hopefully unblock your PR

georgefst:
Note that this works for GHC 9.6 with `--allow-newer='cborg-json:base,dhall:template-haskell'`. At least for the core `dhall` library.

In other words, the library's bound on `template-haskell` needs bumping (though `cabal oudated` suggests many other bumps that wouldn't hurt), and then we're just blocked on https://github.com/well-typed/cborg/pull/313.

dhall/src/Dhall/DirectoryTree.hs
Gabriella439: Wait, something seems amiss here.  I was looking at this more closely and it does not seem to work in my local testing.  The following sample code I tested:

```haskell
module Main where

main :: IO ()
main = putStrLn "foo\n\\
                \bar\n\\
                \baz\n"
```

… gives me a lexical error:

```
Main.hs:4:25: error:
    lexical error in string/character literal at character '\n'
  |
4 | main = putStrLn "foo\n\\
  |                         ^
```

Yet clearly CI passed for your branch so I'm a bit confused.  My intuition, though, is that the double trailing slash should not work, though, since I thought that would get translated to a literal slash in the string.

lrworth: This works for me:

```haskell
{-# LANGUAGE CPP #-}

module Main where

main :: IO ()
main = putStrLn "foo\n\\
                \bar\n\\
                \baz\n"
```

Gabriella439: Ohhhh, I didn't realize that `CPP` changed the semantics of string literals.  Alright, that clears up my confusion.

georgefst:
Could we possibly get a Hackage patch release with this and other recent dependency fixes? I'm unfortunately a bit stuck because of a use of `dhall` in a cabal script, and those don't yet support `source-repository-package` stanzas (https://github.com/haskell/cabal/issues/8024#issuecomment-1441185774).

georgefst:
> In other words, the library's bound on `template-haskell` needs bumping

Hopefully, this will be merged via https://github.com/dhall-lang/dhall-haskell/pull/2542.

georgefst:
Looks like this was released via https://github.com/dhall-lang/dhall-haskell/pull/2545. Thanks @Gabriella439!

Gabriella439:
You're welcome! 😊

Issue #2516, opened 2023-04-20
dhall-1.42.0 fails to build in Stackage Nightly


juhp:
```
       Building library for dhall-1.42.0..
       [69 of 77] Compiling Dhall.TH
       
       /var/stackage/work/unpack-dir/unpacked/dhall-1.42.0-d29671176e59c6861b5d2ff055de0a8160793e10de281c46a8279821c9efc88d/src/Dhall/TH.hs:274:38: error:
           • Couldn't match type: flag0 -> Syntax.TyVarBndr flag0
                            with: Syntax.TyVarBndr ()
             Expected: [Syntax.TyVarBndr ()]
               Actual: [flag0 -> Syntax.TyVarBndr flag0]
           • In the third argument of ‘DataD’, namely ‘params’
             In the expression:
               DataD [] name params Nothing constructors derivingClauses
             In the first argument of ‘pure’, namely
               ‘[DataD [] name params Nothing constructors derivingClauses]’
           |
       274 |                 [pure [DataD [] name params Nothing constructors derivingClauses]] <>
           |                                      ^^^^^^
```

sjakobi:
I can reproduce this with GHC 9.0.1, 9.0.2, 9.2.4 and 9.4.4, and have made a corresponding revision on Hackage: https://hackage.haskell.org/package/dhall-1.42.0/revisions/

The issue was apparently introduced in https://github.com/dhall-lang/dhall-haskell/pull/2506 or an earlier change to that file (CC @ndwarshuis).

I don't understand why CI didn't catch this issue though – I thought the `ubuntu-latest` and `windows-latest` jobs are using GHC 9.0.1.

sjakobi:
> I don't understand why CI didn't catch this issue though – I thought the `ubuntu-latest` and `windows-latest` jobs are using GHC 9.0.1.

These jobs don't make use of the installed "Haskell environment" though – they use `stack` with `lts-18.18` which is based on GHC-8.10.7.

I guess it's high time to add CI with a recent version of GHC.

alaendle:
@sjakobi - to enable `dhall-1.42.0` in nightly we also need relaxed bounds for `template-haskell` and `unix-compat`.

template-haskell-2.19.0.0 ([changelog](http://hackage.haskell.org/package/template-haskell-2.19.0.0/changelog)) is out of bounds for:
[...]

PR #2537, opened 2023-09-27
Dhall.TH: Pass explicit flag in `toTypeVar`

JackKelly-Bellroy:
And raise the `template-haskell` lower bound.

Partial fix of #2516, but doesn't address the unix-compat stuff. `cabal test` failed to run the doctests, so I can't tell whether or not the generated TH is correct — help?


JackKelly-Bellroy:
CI is failing because it's using a stackage LTS which include GHC 8.10.7. How would you like to proceed with that?

PR #2538, opened 2023-09-30
Add Natural/showHex builtin

kukimik:
Implementation of https://github.com/dhall-lang/dhall-lang/pull/1351.

Mostly copy-pasted from https://github.com/dhall-lang/dhall-haskell/commit/3aadd5730bc23d39f765aea99bd2ee38a38f6947.

* Unfortunately I don't know nix and I wasn't able to add a proper nix interpretation of `showHex` in `dhall-nix/src/Dhall/Nix.hs`. I've found something called [`lib.trivial.toHexString`](`https://github.com/NixOS/nixpkgs/blob/7cdd86dc087553a9eb5013b19bd2c4e242e30bad/lib/trivial.nix#L470), but this probably can't be used directly.
* I don't know if (and how/at which moment) I should update the `dhall/dhall-lang` subrepo.


kukimik:
>  Unfortunately I don't know nix and I wasn't able to add a proper nix interpretation of `showHex` in `dhall-nix/src/Dhall/Nix.hs`. I've found something called `lib.trivial.toHexString`, but this probably can't be used directly.

I've learnt enough nix to write this function down :tada: (which turned out to be rather easy), now I just need to find a moment to hnix-ify it.

kukimik:
I added a nix translation of `Natural/showHex`. I don't really like the `elemAt` approach; I'd rather do it [like this](https://github.com/NixOS/nixpkgs/blob/master/lib/trivial.nix#L485), but I didn't have the time to dig through hnix's Antiquotes. Perhaps the current solution is good enough.

Gabriella439:
Just FYI: I'm probably going to cut a release of the Haskell implementation before merging this and the corresponding `dhall-lang` PR, just because I'm a bit overdue on cutting a release

kukimik:
Closing the PR for now; see the discussion in https://github.com/dhall-lang/dhall-lang/pull/1351.

PR #2539, opened 2023-10-03
Relax restrictive upper bound on lens

dalpd:
Hello!

This is done so that the dhall-lsp-server build is fixed on nixpkgs.


mmhat:
@dalpd Thank you!

PR #2543, opened 2023-10-19
treewide: Raise bounds on optparse-applicative

JackKelly-Bellroy:
optparse-applicative-0.18.0.0 is a deprecated release which used layoutSmart from package prettyprinter, which causes the runtime to blow up in certain cases.

See: https://github.com/pcapriotti/optparse-applicative/issues/475#issuecomment-1563698884

Tested with GHC 8.10.7 using `cabal test all --constraint 'optparse-applicative == 0.18.1.0' --allow-newer=hnix:optparse-applicative --allow-newer=dhall-nixpkgs:turtle --keep-going`. Doctests are failing, I'm not sure why. They also fail for me on `main`.

### DIFF
--- a/dhall-docs/dhall-docs.cabal
+++ b/dhall-docs/dhall-docs.cabal
@@ -84,7 +84,7 @@ Library
         text                 >= 0.11.1.0  && < 2.1 ,
         transformers         >= 0.2.0.0   && < 0.7 ,
         mtl                  >= 2.2.1     && < 2.4 ,
-        optparse-applicative >= 0.14.0.0  && < 0.18
+        optparse-applicative >= 0.14.0.0  && < 0.19
     Exposed-Modules:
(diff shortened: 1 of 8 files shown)


Gabriella439:
Usually I don't explicitly blacklist releases via bounds and instead rely on the fact that Cabal's solver will avoid deprecated releases

JackKelly-Bellroy:
Cleaned up the bounds and force-pushed.

Gabriella439:
This looks great!  Thank you 😊

Gabriella439 (approved):

lrworth:
I'd appreciate an analogous hackage revision for dhall-json.

Gabriella439:
I published matching revisions for all the remaining packages (including `dhall-json`)

PR #2547, opened 2023-11-20
Import fileAnIssue from Dhall.Docs.Util

kukimik:
Removes code duplication. Fixes a typo along the way.


Gabriella439:
Thank you!

PR #2555, opened 2023-12-03
Add support for accessing/projecting record type fields

Gabriella439:


winitzki:
What are the conflicts in the git submodule `dhall/dhall-lang`? Should this branch get updated with the latest dhall-lang?

winitzki:
This PR fails with the corner case that I was describing in the discussion here:

```dhall
⊢ { a : Bool, b : Text }.{}

{=}

⊢ {  }.{}

{=}
```

If an empty field selection is applied to a literal record type, the result must be `{}`. In all other cases, the result must be `{=}`.

PR #2560, opened 2023-12-23
feat: Add support for LSP 2

Vekhir:
@mmhat 
Building upon #2531, this further adds support for
- `lsp 2.1`
- `lsp-types 2.0` 
- `lsp-test 0.16`

This is a minimum refactoring by essentially following the compile errors and fixing them one by one without major structural changes. Most changes are just renaming.

No thoughts were given to backwards compatibility within LSP - the above versions are also the minimum requirement for `dhall-lsp-server`.
This is consistent with GHC 9.6.

#### Note
I had this diff lying around for a few days and it has been verified in my build for updating Arch packages to GHC 9.6.

Supersedes #2428

### DIFF
--- a/dhall-lsp-server/src/Dhall/LSP/Handlers.hs
+++ b/dhall-lsp-server/src/Dhall/LSP/Handlers.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE MultiWayIf     #-}
 {-# LANGUAGE ViewPatterns   #-}
+{-# LANGUAGE TypeOperators  #-}
 
 module Dhall.LSP.Handlers where
 
@@ -67,18 +68,18 @@ import Data.Aeson                       (FromJSON(..), Value(..))
 import Data.Maybe                       (maybeToList)
 import Data.Text                        (Text, isPrefixOf)
 import Language.LSP.Server              (Handlers, LspT)
-import Language.LSP.Types               hiding (Range(..), line)
-import Language.LSP.Types.Lens
+import Language.LSP.Protocol.Types      hiding (Range(..))
+import Language.LSP.Protocol.Message
+import Language.LSP.Protocol.Lens
 import System.FilePath
 import Text.Megaparsec                  (SourcePos (..), unPos)
 
 import qualified Data.Aeson              as Aeson
-import qualified Data.HashMap.Strict     as HashMap
 import qualified Data.Map.Strict         as Map
 import qualified Data.Text.Utf16.Rope    as Rope
 import qualified Data.Text               as Text
 import qualified Language.LSP.Server     as LSP
-import qualified Language.LSP.Types      as LSP.Types
+import qualified Language.LSP.Protocol.Types as LSP.Types
 import qualified Language.LSP.VFS        as LSP
 import qualified Network.URI             as URI
 import qualified Network.URI.Encode      as URI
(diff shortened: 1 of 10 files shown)


Vekhir:
@mmhat Rather obvious in hindsight, I've now updated the stack configuration to use the newer `lsp` libraries.
Not sure what to do about the hydra/nix issue.

mmhat:
@Vekhir Thank you for doing this! I am about to review the work you did so far. Regarding the Hydra issue:
I am an ArchLinux user as well, so I faced the same problems while working on #2531 . I used the following to update packages in the Nix setup:
```shell
# We assume that the current directory is the root of the checked out work tree.
# Start a container with the work tree mounted under /work.
# (I use podman nowadays; docker should work as well.)
podman run -it --rm -v .:/work -w /work docker.io/nixos/nix
# Now we are in the container. We need to install cabal-install and cabal2nix:
nix-env -i cabal-install -i cabal2nix
# cabal2nix needs a hackage index tarball:
cabal update
# Now we generate the Nix expression for the lsp-2.1.0.0 package and store it in the right location:
cabal2nix cabal://lsp-2.1.0.0 > ./nix/packages/lsp.nix
```

dhall-lsp-server/src/Dhall/LSP/Server.hs
```diff
@@ -78,7 +79,7 @@ run = withLogger $ \ioLogger -> do
               ]
         }
 
-  let staticHandlers =
+  let staticHandlers _clientCapabilities =
```
mmhat, line 82: Do we need to respect the client's capabilities indicated by this argument?

mmhat (commented):
LGTM.

dhall-lsp-server/src/Dhall/LSP/Server.hs
Vekhir: The [documentation](https://hackage.haskell.org/package/lsp-2.3.0.0/docs/Language-LSP-Server.html#t:ServerDefinition) heavily implies it's optional and the handlers didn't previously use them, so I think we can ignore it.

dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -132,7 +133,7 @@ rangeToJSON (Range (x1,y1) (x2,y2)) =
 
 hoverHandler :: Handlers HandlerM
 hoverHandler =
-    LSP.requestHandler STextDocumentHover \request respond -> handleErrorWithDefault respond Nothing do
+    LSP.requestHandler SMethod_TextDocumentHover \request respond -> handleErrorWithDefault respond (maybeToNull Nothing) do
```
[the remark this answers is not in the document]
Vekhir: Just saying `Null` here leads to ambiguity about `Aeson.Null` vs `LSP.Types.Null`, but otherwise is fine.

Vekhir:
@mmhat I've implemented your suggestions and fixed the CI build - it will hopefully pass now.

Vekhir:
Hydra now works, lets see if the rest works now too

Vekhir:
This check will most probably fail since `hashable 1.3.5.0` needs `ghc-bignum`, but I want to see the error message.
If `stack` provides a compatible version for `ghc-bignum`, then we are likely done.

Vekhir:
Failed due to typo...

Vekhir:
@mmhat Looks like we are done. If you don't have any objections, you can merge it.
Is there something you want/need done to make a Hackage release? Would make Arch integration easier.

mmhat:
@Vekhir Can you please rebase your changes onto `main`?

Regarding a new release: Those are done by @Gabriella439 , so I think we have to be a bit more patient.

mmhat:
@Vekhir Nice, excellent work! Thank you!

Vekhir:
It's not urgent anyway. Thanks for your suggestions and reviewing!

PR #2565, opened 2024-01-03
dhall-docs: render standalone text files as preformatted text

kukimik:
I noticed that text files that are not imported by dhall files are not rendered as preformatted text, which often looks awful. See the `README.md` for `dhall-semver`:

image

The raw text looks much better: https://raw.githubusercontent.com/Gabriella439/dhall-semver/main/README.md. Moreover, displaying as preformatted text is much closer to the real contents of the file.

I can't be 100% sure, but this looks like an unintentional omission in #2407. The test didn't catch it because the `Plain.txt` file is imported by `AsText.dhall` and thus treated differently (I didn't look for the part of code responsible for this behaviour, I'm guessing based on what I see).

Side note: it's also rather easy to add support for rendering Markdown files as HTML (I have a working proof of concept, but it needs some polish). This could be nice for files like `README.md`, which are common in Dhall packages. On the other hand, Markdown is quite readable on its own. Anyway, that's a story for another PR.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,4 +5,5 @@
 *.golden binary
 dhall-csv/tasty/data/* binary
 dhall-docs/tasty/data/golden/**/*.html binary
+dhall-docs/tasty/data/package/StandaloneTextFile.txt binary
 dhall-yaml/tasty/data/* binary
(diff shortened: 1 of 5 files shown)


kukimik:
Hmm.. I don't know why the tests failed on Windows. I guess it may be LF vs CRLF inside the `<pre>...</pre>` block? Is there any way to see what file was generated during the tests? I don't have a machine with Windows configuration, so it's difficult for me to try it out myself.

kukimik:
It was a `\n` vs `\r\n` (dos vs unix newlines) issue caused by git. Solved by adding a line to `.gitattributes`. The problem is discussed in [`tasty-golden` docs](https://hackage.haskell.org/package/tasty-golden-2.3.5/docs/Test-Tasty-Golden.html).

PR #2571, opened 2024-02-23
lsp-server: Compatibility with lsp ==2.3.0.0

rvl:
This change lets dhall-lsp-server build and pass unit tests with [lsp-2.3.0.0](https://hackage.haskell.org/package/lsp-2.3.0.0/changelog).

The PR is in draft until I've done some actual testing with eglot.


Gabriella439:
If you merge https://github.com/rvl/dhall-haskell/pull/1 I think that will fix the build failures

rvl:
Thanks @Gabriella439, that's merged now.

About `stack.ghc-8.10.yaml` -- I'm not sure whether `lsp-2.3.0.0` will build with Stackage LTS 18.28. :question:  So even though `dhall` works with ghc-8.10, `dhall-lsp-server` may not.

Gabriella439:
I wouldn't worry about it.  I don't think people use those other `stack-*.yaml` files that much and if they have any issues they'll put up a PR to fix them.

PR #2574, opened 2024-03-11
dhall-docs: Fix white flashes on page load in dark mode

kukimik:
Loading a page with dark mode enabled currently results in a short white flash. This happens because the page is first rendered in light mode, and only after it is loaded the class indicating that dark mode is enabled is toggled. The effect is even more visible because of the 0.5s transition effect set in CSS.

test

This PR solves the problem by toggling the class indicating that dark mode is enabled before the page is rendered (the script is called inside `<head>`). The class is now set on the `<html>` element instead of `<body>` because `document.body` is not ready yet when the script is executed.

### DIFF
--- a/dhall-docs/src/Dhall/data/assets/index.css
+++ b/dhall-docs/src/Dhall/data/assets/index.css
@@ -187,7 +187,7 @@ span.of-type-token {
     color: #EBEBEB;
 }
 
-body.dark-mode {
+.dark-mode body {
     background-color: #484848;
 }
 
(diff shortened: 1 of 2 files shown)


kukimik:
The hydra build failed with:

```
building
Preprocessing library 'metamodel' for lsp-types-2.0.1.0..
Building library 'metamodel' for lsp-types-2.0.1.0..
[1 of 2] Compiling Language.LSP.MetaModel.Types ( metamodel/Language/LSP/MetaModel/Types.hs, dist/build/metamodel/Language/LSP/MetaModel/Types.o, dist/build/metamodel/Language/LSP/MetaModel/Types.dyn_o )

<no location info>: error:
    /tmp/nix-build-lsp-types-2.0.1.0.drv-0/ghc1648975_0/ghc_4.s: hClose: resource exhausted (No space left on device)
```

which is probably not my fault.

PR #2579, opened 2024-04-03
Render Markdown files as HTML

kukimik:
This PR introduces HTML rendering of Markdown files, as announced in #2565. The main usecase is rendering `README.md` files accompanying a package.

If a file has the `.md` extension and can be parsed by the MMark parser, we display both the HTML output generated by MMark, and the Markdown sources, which are hidden by default using a `<details>` tag. If a problem occurs in the parsing phase, we treat the file as plain text and display a warning.

One issue that I am aware of is handling of local (non-external) links. For example in the Prelude's `README.md` we have a link pointing to `./Natural/fold`. In the generated documentation this will stil point to `./Natural/fold`, but there is no such file in the docs: we are missing the `.html` extension. (And for links to folders we should add a trailing `/index.html`.) A "full" solution of this could be rather tricky, as it should involve handling of symlinks etc., and would probably need IO and some restructuring of the code. I think a simple solution that works in most real-life scenarios can be implemented using a [MMark extension](https://hackage.haskell.org/package/mmark-0.0.7.6/docs/Text-MMark-Extension.html), a helper function similar to `shake`'s [`NormaliseEx`](https://hackage.haskell.org/package/shake-0.19.8/docs/Development-Shake-FilePath.html#v:normaliseEx) and the list of rendered files. However, this would still require some work, and I believe the present changes already improve the `dhall-docs` tool a little bit, so I'd rather take care of this in a separate PR. (Moreover, rendering of comments in `.dhall` files has the same issue, so it could be fixed in both places at once.)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,4 +6,6 @@
 dhall-csv/tasty/data/* binary
 dhall-docs/tasty/data/golden/**/*.html binary
 dhall-docs/tasty/data/package/StandaloneTextFile.txt binary
+dhall-docs/tasty/data/package/MarkdownFile.md binary
+dhall-docs/tasty/data/package/InvalidMarkdownFile.md binary
 dhall-yaml/tasty/data/* binary
--- a/dhall-docs/dhall-docs.cabal
+++ b/dhall-docs/dhall-docs.cabal
@@ -22,12 +22,14 @@ Category: Compiler
 Extra-Source-Files:
     CHANGELOG.md
     README.md
+    tasty/data/package/*.md
     tasty/data/package/*.txt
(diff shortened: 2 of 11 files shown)


kukimik:
@Gabriella439 There's a problem with the hydra build (a cached out-of-disk-space failure), it affects all new PRs. No rush, I'm just pinging in case you missed it.

Gabriella439:
I fixed the disk issues, although the `dhall-docs` build still has one test failure remaining

kukimik:
Thanks! I forgot to add the new test files as `Extra-Source-Files` in `dhall-docs.cabal`. Fixed.

PR #2580, opened 2024-04-12
dhall-docs: fix an error and a warning reported by the w3c validator

kukimik:
Except some special cases `` elements should have an `alt` attribute; otherwise it's an accessibility issue. The [w3c validator](https://validator.w3.org/) reports this as an error.

The validator also suggested (in the form of warning) to remove the `type` attribute of the `<script>` tag, because it is unnecessary.


Gabriella439:
Thank you for doing this!

Issue #2583, opened 2024-04-23
Questions regarding the right strategy for upgrading to ghc-9.8


lucasdicioccio:
Hi there, I'm attempting to increase this repository to ghc-9.8.2 with an aim for a Hackage release (so that my blog-engine can be on Hackage as well) at some point before this summer (personal reasons for the deadline). I work on a fork at https://github.com/lucasdicioccio/dhall-haskell .

My patch is not ready yet but an early attempt shows:
- a few bounds could be/would have to be relaxed
- some deeper failures will need bound-relaxing/uploads as well (hnix for `dhall-lsp-server`, `dhall-nix` and `dhall-nixpkgs`)
- some CPP will be needed to support (without warning) old imports (liftA2 being part of the class Control.Applicative)
- some rework around (safe? -- i haven't convinced myself) calls to `Prelude.head` which turns a deprecation warning
- my auto-formatting vim autocommand to fourmolu-format on save trapped me
- some more CPP needed as megaparsec now has an additional packages

I don't mind continuing, however I'm interested to know whether:
- are you aware of someone already on a further track than myself?
- is there some target-compatibility level with older GHCs/megaparsec? (to cut on CPP)
- is there some plan to fourmolu-format the code (in which case I'd rather see that done in a lockstep before xor after porting to 9.8.2 to reduce diff noise/rebase conflicts)?

ysangkok:
dhall-1.42.2 is out which offers compatibility:

```
% cabal repl -b dhall -w ghc-9.8.4
Resolving dependencies...
Build profile: -w ghc-9.8.4 -O1
In order, the following will be built (use -v for more details):
 - aeson-2.2.3.0 (lib) (requires build)
 - atomic-write-0.2.1.0 (lib) (requires download & build)
 - megaparsec-9.7.0 (lib) (requires build)
 - parsers-0.12.12 (lib) (requires download & build)
 - streaming-commons-0.2.3.0 (lib) (requires download & build)
 - tls-2.1.7 (lib) (requires download & build)
 - aeson-pretty-0.8.10 (lib) (requires build)
 - http-client-0.7.18 (lib) (requires build)
 - crypton-connection-0.4.3 (lib) (requires build)
 - cborg-json-0.2.6.0 (lib) (requires download & build)
 - http-client-tls-0.3.6.4 (lib) (requires build)
 - dhall-1.42.2 (lib) (requires download & build)
 - fake-package-0 (interactive) (lib) (first run)
 ```

[...]

PR #2584, opened 2024-04-25
Port to GHC-9.8.2 (missing: dhall-{lsp-server,nix,nixpkgs})

lucasdicioccio:
Starting something to upgrade to 9.8.2, taking baby-steps absent #2583 .

My goal is to compile with `-Werror` with 9.8.2 (I run cabal 3.10.3.0 but that shouldn't matter much) then pass tests (without -Werror then with).

- [x] I've mainly replaced safe `head` calls to `NonEmpty.head . NonEmpty.fromList` except in `dhall/src/Dhall/Tags.hs` (a more human-friendly symbol is welcomed) and `dhall-openapi/src/Dhall/Kubernetes/Types.hs` (we may expect the parser to occur at a boundary where human-readable errors are paramount)
- [x] upgrade some package bounds here and there
- [x] CPP-guard `liftA2` imports 
- [ ] pass tests
- [ ] same for `dhall-{lspt-server,nix,nixpkgs})` (tbc: blocked on `hnix`)

 I'm struggling to find cycles to feel free to take over the pull-request.

### DIFF
--- a/dhall-csv/dhall-csv.cabal
+++ b/dhall-csv/dhall-csv.cabal
@@ -33,7 +33,7 @@ Library
     Hs-Source-Dirs: src
     Build-Depends:
         base                 >= 4.12.0.0  && < 5   ,
-        bytestring                           < 0.12,
+        bytestring                           < 0.13,
         cassava              >= 0.5.0.0   && < 0.6 ,
         containers           >= 0.5.9     && < 0.7 ,
         either                                     ,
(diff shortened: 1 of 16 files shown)


ysangkok:
I think these look like sensible changes. Given that GHC 9.8.3 will be out in a few days, and the Stackage LTS after that,  I think it would be good to merge and release this, so we can keep Dhall in Stackage LTS.

Any reason this is still 'draft' @lucasdicioccio ?

ysangkok:
I am not sure how much of the work here is still pending on the default branch, but basic GHC 9.8 usage should already work since

* https://github.com/dhall-lang/dhall-haskell/pull/2622

lucasdicioccio:
> I am not sure how much of the work here is still pending on the default branch, but basic GHC 9.8 usage should already work since
> 
>     * [Fix TyVar BndrVis #2622](https://github.com/dhall-lang/dhall-haskell/pull/2622)

I think we can close this PR. Sorry the time I can allocate for open-source contribs is pretty narrow since a while ago.

PR #2585, opened 2024-05-11
optimize Natural/fold in the strict case

winitzki:
The proposed optimization follows the discussion in https://github.com/dhall-lang/dhall-lang/issues/1213#issuecomment-1855878600 

The beta-normalization for `Natural/fold n t succ zero` corresponds to evaluating `succ(succ(succ(...(succ(zero))...)))` where we apply `succ` exactly `n` times.

The current code uses a loop where `succ` is always evaluated `n` times.
However, sometimes the loop can be stopped early because the values stop changing.

For example:

```dhall
let f : Natural → Natural = λ(x : Natural) → if Natural/isZero x then 1 else x
in Natural/fold 10000000 Natural f 5
```

With this definition, we have `f x = x` when x is nonzero. The expression evaluates to 5, but the evaluation takes time linear in the first argument (10000000) because the function `f` will be applied that many times to the value `5` and beta-normalization will be performed each time.

The new code rewrites `strictLoop` via `strictLoopShortcut`, which stops the loop as soon as `f x = x`.

This optimization is only applied in the "strict" case (`Normalization.hs` introduces `strict` and `lazy` cases when evaluating `Natural/fold`.)

There should be no change of behavior after this optimization. 
This is because all Dhall functions are pure; if `f x = x` then the result of evaluating `f (f (f x))` is exactly the same as just `x`.
So, the only effect of the new code is that some `Natural/fold` expressions will be beta-normalized faster than before.

I'd like to add some tests for the new behavior but I don't know how. The new behavior returns exactly the same normal forms as before, but does it faster when the "shortcut" is detected (when `f x = x` is detected). I am not a Haskell programmer, so any suggestions will be welcome here.

Another issue is that I am not familiar with the Haskell build systems, and  both `cabal new-build dhall` as well as `stack build` fail on my machine in this project, with dependency errors that I am unable to fix.

### DIFF
--- a/dhall/src/Dhall/Eval.hs
+++ b/dhall/src/Dhall/Eval.hs
@@ -527,9 +527,13 @@ eval !env t0 =
                                 -- following issue:
                                 --
                                 -- https://github.com/ghcjs/ghcjs/issues/782
-                                let go !acc 0 = acc
-                                    go  acc m = go (vApp succ acc) (m - 1)
-                                in  go zero (fromIntegral n' :: Integer)
+                                go zero (fromIntegral n' :: Integer) where
+                                  go !acc 0 = acc
+                                  go (VNaturalLit n) m =
+                                    case vApp succ (VNaturalLit n) of
+                                      VNaturalLit n' | n == n' -> VNaturalLit n
+                                      next -> go next (m - 1) 
+                                  go acc m = go (vApp succ acc) (m - 1)
                             _ -> inert
         NaturalBuild ->
             VPrim $ \case
--- a/dhall/src/Dhall/Eval.hs
+++ b/dhall/src/Dhall/Eval.hs
@@ -122,7 +122,7 @@ instance Semigroup (VChunks a) where
   VChunks xys z <> VChunks [] z' = VChunks xys (z <> z')
   VChunks xys z <> VChunks ((x', y'):xys') z' = VChunks (xys ++ (z <> x', y'):xys') z'
 
-instance Monoid (VChunks b) where
+instance Monoid (VChunks a) where
   mempty = VChunks [] mempty
 
 {-| Some information is lost when `eval` converts a `Lam` or a built-in function
--- a/dhall/src/Dhall/Eval.hs
+++ b/dhall/src/Dhall/Eval.hs
@@ -529,10 +529,10 @@ eval !env t0 =
                                 -- https://github.com/ghcjs/ghcjs/issues/782
                                 go zero (fromIntegral n' :: Integer) where
                                   go !acc 0 = acc
-                                  go (VNaturalLit n) m =
-                                    case vApp succ (VNaturalLit n) of
-                                      VNaturalLit n' | n == n' -> VNaturalLit n
-                                      next -> go next (m - 1) 
+                                  go (VNaturalLit x) m =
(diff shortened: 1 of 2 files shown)


dhall/src/Dhall/Normalize.hs
```diff
@@ -211,8 +211,13 @@ normalizeWithM ctx e0 = loop (Syntax.denote e0)
...
 
-                        strictLoop 0 = loop zero
-                        strictLoop !n = App succ' <$> strictLoop (n - 1) >>= loop
+                        strictLoop !n = strictLoopShortcut n (loop zero)
+                        strictLoopShortcut 0 !res = res
+                        strictLoopShortcut !n !res = do
+                            x <- res
+                            let next_res = App succ' <$> res >>= loop
+                            y <- next_res
+                            if judgmentallyEqual x y then res else strictLoopShortcut (n - 1) next_res
```
Gabriella439, line 220: I think you can avoid evaluating `rest` and `next_res` twice by structuring it like this:

```suggestion
                        strictLoop !n = do
                            z <- loop zero
                            strictLoopShortcut n z

                        strictLoopShortcut 0 !previous = pure previous
                        strictLoopShortcut !n !previous = do
                            current <- loop (App succ' previous)
                            if judgmentallyEqual previous current
                                then pure previous
                                else strictLoopShortcut (n - 1) current
```

winitzki:
@Gabriella439  Thank you for the suggestion.

I'm not sure why the build fails for MacOS. I also have trouble building `dhall-haskell` on my Mac. I opened another PR to see if the project builds with no changes. https://github.com/dhall-lang/dhall-haskell/pull/2586

winitzki:
@Gabriella439  My changes do not actually seem to work. I built a local executable and I don't observe any speedup in the test code shown above.

```dhall
let f : Natural → Natural = λ(x : Natural) → if Natural/isZero x then 1 else x
in Natural/fold 100000000 Natural f 5
```
takes about 5 seconds but should be instantaneous. Do you have any idea how to fix this? I tried my previous code, and I also tried using `strict` instead of `lazy` but nothing helped.

winitzki:
I added some new code to `Eval.hs` but I couldn't figure out how to do it properly. I need to compare two values of type `Val a` and it doesn't seem to be possible in general.

dhall/src/Dhall/Eval.hs
```diff
@@ -527,9 +527,13 @@ eval !env t0 =
                                 -- following issue:
                                 --
                                 -- https://github.com/ghcjs/ghcjs/issues/782
-                                let go !acc 0 = acc
-                                    go  acc m = go (vApp succ acc) (m - 1)
-                                in  go zero (fromIntegral n' :: Integer)
+                                go zero (fromIntegral n' :: Integer) where
+                                  go !acc 0 = acc
+                                  go (VNaturalLit x) m =
```
winitzki, line 532: Here, we detect the shortcut only when the accumulator is a `Natural` literal. However, we would like to detect the shortcut in all cases. I could not find any way of comparing `Val` values. What kind of code would need to be written here?

Gabriella439: `Dhall.Eval.conv` is what you want

PR #2586, opened 2024-05-12
fixing the MacOS build

winitzki:
The goal of this PR is to fix any problems with the build of dhall-haskell.

Current issues:

- Mac OS build failing on Sonoma 14.4.1 (Intel x86_64) due to failures with building `saltine` after `stack build`.
- Mac OS build failing due to libsodium in the PR https://github.com/dhall-lang/dhall-haskell/pull/2585 (but builds are successful on other OSes)

The error is in installing `libsodium`. 

```
tried: '/opt/homebrew/lib/libsodium.dylib' (mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64'))
```

This discussion https://stackoverflow.com/questions/72970363 suggests using the command: `arch -x86_64 /usr/local/bin/brew install libsodium`


winitzki:
Currently the macos build seems to be broken. I've been trying randomly poking around to fix it but I don't really know what I'm doing in this build system. https://github.com/dhall-lang/dhall-haskell/pull/2586  Anyone has any ideas? The failure is around `libsodium` and has to do with macos architectures (x86 vs arm).

The error seems to be confusingly worded. The error message says that it found libsodium with arm64 architecture but it requires libsodium with x86 architecture. However, it is wrong that it should have looked for libsodium with x86 architecture, because the build is for `macos-latest`  which is arm64. See here for the list. https://github.com/actions/runner-images?tab=readme-ov-file#available-images

It seems that macos-latest is configured to build arm64. However, I am also not able to change that into building x86.

What I have tried so far:

- Adding a command to install libsodium (`brew install libsodium`). This did not help because libsodium was already installed by another command somewhere.
- Adding a command to install libsodium on mac (`brew reinstall libsodium`) with x86 architecture specifically. This fails because `brew` is installed for arm64 and refuses to install x86 artifacts. I don't know how I can reinstall `brew` correctly.
- Changing the build in `main.yaml` to use a different macos runner: macos-latest-large, macos-14-large and so on. This simply fails even to start the build, with no error messages.

winitzki:
The build fails immediately if the runner is set to `macos-14`, `macos-14-large`, `macos-latest-large`. (Although those runners are supposed to be working now. https://github.com/actions/runner-images?tab=readme-ov-file#available-images )

The previous working version was `macos-13` and I'm trying it right now. However, the github settings for this repo require `macOS-latest` build to succeed.

winitzki:
I found this page:
https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#about-macos-larger-runners
which says:
> Larger runners are only available for organizations and enterprises using the GitHub Team or GitHub Enterprise Cloud plans.
Probably, this is why the CI job simply fails immediately when I specified `macos-14-large` or other "large" runners.

So, we have a choice: we can keep the macos build at version 13 (this still works) or we need to somehow fix the macOS ARM64 build. Currently, the ARM64 build fails because of some dependency problems with `hnix` that depends on `libsodium` and appears to be misconfigured: when building `hnix` on ARM64 architecture, the build expects `libsodium` to be for the x86 architecture.

mmhat:
@winitzki  This appears to be an upstream bug: https://github.com/actions/runner-images/issues/9760

.github/workflows/main.yml
```diff
@@ -49,7 +49,7 @@ jobs:
             ${{ matrix.os.runner }}-
           path: |
             ${{ steps.setup-haskell-cabal.outputs.stack-root }}
-      - name: Install libsodium
+      - name: Install libsodium x86_64
```
mmhat, line 52:
```suggestion
      - name: Install libsodium
```

mmhat (approved):
I think we should pin the MacOS runner first in order to get working mac builds again, but https://github.com/dhall-lang/dhall-haskell/issues/2587 should stay open and the subject should be adjusted. Thought?

mmhat:
@winitzki It might be worth trying to use haskell-actions/setup@v2.7 too.

winitzki:
> I think we should pin the MacOS runner first in order to get working mac builds again, but #2587 should stay open and the subject should be adjusted. Thought?

In my view, the root cause is that `dhall-haskell` cannot build under macos-arm64 due to some dependency problems in `hnix-store` and/or elsewhere. The macos-arm64 build was actually never tested in `dhall-haskell`, and in fact there were never any macos-arm64 executables released for dhall. When github changed the `macos-latest` runners to arm64, all builds started failing.

To resolve this issue, I propose the following path forward:

1. Use `macos-13` instead of `macos-latest` (as in this PR). Change the repo settings so that `macos-latest` is no longer a requirement for passing the CI builds. (I don't have permissions to do this.) Mention in the README that macos-arm64 is currently unsupported.
2. Fix the dependency problems so that macos-arm64 build becomes successful. Enable macos-13 and macos-latest builds and publish both dhall-darwin-x86 and dhall-darwin-arm64 executables. Mention in the README that macos-arm64 is now supported.

winitzki:
The macos build passed with haskell actions 2.7. Thank you for suggesting that!

Let me add macos-13 build as well.

winitzki:
Let me know if this is OK to merge. This will unblock other builds.

mmhat:
> > I think we should pin the MacOS runner first in order to get working mac builds again, but #2587 should stay open and the subject should be adjusted. Thought?
> 
> In my view, the root cause is that `dhall-haskell` cannot build under macos-arm64 due to some dependency problems in `hnix-store` and/or elsewhere. The macos-arm64 build was actually never tested in `dhall-haskell`, and in fact there were never any macos-arm64 executables released for dhall. When github changed the `macos-latest` runners to arm64, all builds started failing.

Indeed.

> 
> To resolve this issue, I propose the following path forward:
> 
> 1. Use `macos-13` instead of `macos-latest` (as in this PR). Change the repo settings so that `macos-latest` is no longer a requirement for passing the CI builds. (I don't have permissions to do this.) Mention in the README that macos-arm64 is currently unsupported.

Would you mind adding that comment to the README?

> 2. Fix the dependency problems so that macos-arm64 build becomes successful. Enable macos-13 and macos-latest builds and publish both dhall-darwin-x86 and dhall-darwin-arm64 executables. Mention in the README that macos-arm64 is now supported.

:+1: 

> Let me know if this is OK to merge. This will unblock other builds.

LGTM, merge it when you're done. If you include bigger changes, don't hesitate to ask for another review!

> 
> What needs to be done in order to publish both macos-x86 and macos-arm64 executables when a release is made?

I'd proceed as follows:

1. Parameterize the objects of the workflow with `runner.arch` from https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context. In particular https://github.com/dhall-lang/dhall-haskell/blob/e281834c9ae9b27e7e42810465c8802b66fd6488/.github/workflows/main.yml#L31, https://github.com/dhall-lang/dhall-haskell/blob/e281834c9ae9b27e7e42810465c8802b66fd6488/.github/workflows/main.yml#L47 (maybe the `restore-key:` too) and https://github.com/dhall-lang/dhall-haskell/blob/e281834c9ae9b27e7e42810465c8802b66fd6488/.github/workflows/main.yml#L158
2. Also use `runner.arch` in the conditional executions here: https://github.com/dhall-lang/dhall-haskell/blob/e281834c9ae9b27e7e42810465c8802b66fd6488/.github/workflows/main.yml#L74, https://github.com/dhall-lang/dhall-haskell/blob/e281834c9ae9b27e7e42810465c8802b66fd6488/.github/workflows/main.yml#L99, https://github.com/dhall-lang/dhall-haskell/blob/e281834c9ae9b27e7e42810465c8802b66fd6488/.github/workflows/main.yml#L148
3. I think it is good to get rid of https://github.com/dhall-lang/dhall-haskell/blob/e281834c9ae9b27e7e42810465c8802b66fd6488/.github/workflows/main.yml#L130 and use `runner.arch` in https://github.com/dhall-lang/dhall-haskell/blob/e281834c9ae9b27e7e42810465c8802b66fd6488/.github/workflows/main.yml#L137.

PR #2588, opened 2024-05-13
dhall: raise upper bound on Diff

JackKelly-Bellroy:
We have set `allow-newer` internally and it is working.

### DIFF
--- a/dhall/dhall.cabal
+++ b/dhall/dhall.cabal
@@ -218,7 +218,7 @@ Common common
         contravariant                              < 1.6 ,
         data-fix                                   < 0.4 ,
         deepseq                                    < 1.6 ,
-        Diff                        >= 0.2      && < 0.5 ,
+        Diff                        >= 0.2      && < 0.6 ,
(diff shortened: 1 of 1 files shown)


JackKelly-Bellroy:
CI failure appears unrelated.

winitzki:
Currently the macos build seems to be broken. I've been trying randomly poking around to fix it but I don't really know what I'm doing in this build system. https://github.com/dhall-lang/dhall-haskell/pull/2586  Anyone has any ideas? The failure is around `libsodium` and has to do with macos architectures (x86 vs arm).

mmhat:
@JackKelly-Bellroy Thank you!

ysangkok:
@Gabriella439 Would it be possible to get this bump as a Hackage revision? Then we could get Dhall back into Stackage.

kokobd:
We are also waiting for a Hackage revision so `allow-newer` can be removed from our `cabal.project`

Gabriella439:
I published a matching revision

PR #2589, opened 2024-05-24
improve Dhall.Map.traverseWithKey performance

imuli:
`traverseWithKey` currently calls `fromList`, which creates a new list of keys (and calls `nubOrd` on it). this is unnecessary, because a traversal doesn't change the keys.

I haven't done any benchmarking of `traverseWithKey` specifically, but this change reduces my overall parse+typecheck+inject+normalize+extract time by ~20%.


imuli:
Thank you @winitzki!

Gabriella439 (approved):

PR #2591, opened 2024-05-31
Add Stack setup for GHC 9.4.8

mmhat:


Gabriella439:
I retriggered the Hydra builds, although it looks like at least `dhall-nixpkgs` still needs some changes (it's emitting warnings)

mmhat:
@Gabriella439 Thank you very much! Unfortunately Hydra still reports that there is no space left...

PR #2593, opened 2024-06-05
Use normalized filepaths in dhall testsuite

mmhat:
Missing filepath normalization caused some failures in #2591 ; See https://github.com/dhall-lang/dhall-haskell/actions/runs/9338722607/job/25702207276?pr=2591 for an example.

Other changes:
- Require `turtle >=1.6` in all packages.
- Use `hnix-0.17` in Stack configurations.
- Also add `mkParsec` method to `MonadParsec` instance for `megaparsec >=9.4`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,7 +59,7 @@ jobs:
           if [ '${{matrix.os.runner}}' == 'ubuntu-latest' ]; then
             sudo apt-get install -y libsodium-dev
           fi
-          pkg-config --path sodium || true
+          pkg-config --libs --modversion sodium || true
       - name: "Build"
         shell: bash
         run: |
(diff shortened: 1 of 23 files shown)


winitzki:
A hydra failure might be cached; I had to make a dummy change (renamed a dummy variable) to make hydra forget the failure in a previously cached build.

mmhat:
@winitzki Unfortunately that didn't work.
Hydra is out of disk space; See https://hydra.dhall-lang.org/build/115335/nixlog/1 and https://hydra.dhall-lang.org/build/115308/nixlog/1 .

I don't have access to that machine; @Gabriella439 can you have a look?

winitzki:
"Hydra can't really be killed. You cut one head off, two more will grow."

mmhat:
> "Hydra can't really be killed. You cut one head off, two more will grow."

@winitzki :smile: It seems you are right; It's working again.

PR #2599, opened 2024-07-01
Allow aeson-2.2 in dhall-yaml and dhall-lsp-server

kokobd:
Currently `dhall-json` has `aeson < 2.3` but `dhall-yaml` still has `aeson < 2.2`. This PR extends the later upper bound to 2.3 too.


mmhat:
@kokobd Thank you!

kokobd:
Would you mind creating a new metadata revision on Hackage so that we won't need `allow-newer` in `cabal.project`?

Gabriella439:
I published matching revisions to Hackage

Issue #2606, opened 2024-09-30
`dhall-toml` appears to handle relative paths incorrectly


bcdarwin:
```
$ cat test.sh
#!/bin/bash

mkdir src

echo '{ a = 3 }' > src/a.dhall
echo '{ a = "wrong" }' > a.dhall
echo '(./a.dhall)' > src/config.dhall

dhall type --file src/config.dhall
echo "dhall-yaml: $(dhall-to-yaml --file src/config.dhall)"
echo "dhall-toml: $(dhall-to-toml --file src/config.dhall)"

$ nix shell nixpkgs#haskellPackages.dhall nixpkgs#haskellPackages.dhall-json nixpkgs#haskellPackages.dhall-toml
$ bash ./test.sh
{ a : Natural }
dhall-yaml: a: 3
dhall-toml: a = "wrong"
$ echo $(dhall --version), $(dhall-to-toml --version)
$ 1.42.1, 1.0.3
```

I expect this to be local to dhall-to-toml, hence relatively straightforward; would be willing to look into it, but not immediately.

First noticed by @cfhammill.

PR #2607, opened 2024-10-01
`dhall-toml`: Fix loading of relative paths

Gabriella439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/2606


TristanCacqueray (approved):
That's what the other CLI does.

PR #2611, opened 2024-10-21
Use Data.Map.Strict instead of Data.Map

phagenlocher:
When trying to read a 93Mb Dhall file with `inputExprWithSettings` (using a small list of subtitutions), Dhall currently crashes due to a spaceleak (presumably). Profiling showed that `Dhall.Map` is to blame for the allocations (with `toList` being the main retainer) and the only way to get this under control was to switch to `Data.Map.Strict`. 

Is there a specific reason why the lazy `Data.Map` is used in this module? It seems to me that laziness provides no benefit and moving to the strict version is the way to go.

Profiling eventlogs (for a smaller file) can be found here: [profiling.zip](https://github.com/user-attachments/files/17456672/profiling.zip)

### DIFF
--- a/dhall/src/Dhall/Map.hs
+++ b/dhall/src/Dhall/Map.hs
@@ -77,7 +77,7 @@ import Language.Haskell.TH.Syntax (Lift)
 import Prelude                    hiding (filter, lookup)
 
 import qualified Data.List
-import qualified Data.Map
+import qualified Data.Map.Strict
 import qualified Data.Set
 import qualified GHC.Exts
 import qualified Prelude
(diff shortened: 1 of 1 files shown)


winitzki:
Not sure why the Hydra build failed.

phagenlocher:
> Not sure why the Hydra build failed.

@winitzki This shouldn't be related to the code changes, right?

winitzki:
> > Not sure why the Hydra build failed.
> 
> @winitzki This shouldn't be related to the code changes, right?

Right, I believe it has nothing to do with the code changes. All tests passed in this PR. A similar situation is happening right now in another PR: https://github.com/dhall-lang/dhall-haskell/pull/2513

Most likely it's a stale failed build that Hydra has cached for some reason. Someone needs to restart the server somewhere.

Gabriella439:
Yeah, this is an issue with Hydra where it sometimes run out of memory.  The main reason I still use Hydra is because it ensures that the cache is preserved between builds, which at the time was hard to get with other CI solutions.  However, I have been thinking of switching to Garnix for Nix builds, which would fix this problem (I think).

TristanCacqueray:
@Gabriella439 there is also `DeterminateSystems/magic-nix-cache-action@main` to keep the cache between GitHub actions, perhaps that would work for Dhall?

Issue #2612, opened 2024-10-26
LSP does not handle STextDocumentDidClose


ShrykeWindgrace:
Hi!

I see the following error in vscode's window for dhall's lsp output:
```
lsp:no handler for:  STextDocumentDidClose
```
I run on windows with
```console
PS> dhall-lsp-server --version
1.1.3
PS> code --version
1.94.2
```
The plugin for vscode: "Dhall LSP Server v0.0.4"

I'll see if I can make PR with a fix.

mmhat:
Fixed in #2613; Closing.

PR #2613, opened 2024-10-27
Add handler for 'documentDidClose'

ShrykeWindgrace:
Addresses #2612


dhall-lsp-server/src/Dhall/LSP/Handlers.hs
```diff
@@ -641,6 +641,11 @@ cancelationHandler :: Handlers HandlerM
 cancelationHandler =
     LSP.notificationHandler SMethod_CancelRequest \_ -> return ()
 
+-- this handler is a stab to prevent `lsp:no handler for:` messages.
```
mmhat, line 644:
```suggestion
-- This handler is a stub to prevent `lsp:no handler for:` messages.
```

mmhat (approved):
LGTM

PR #2618, opened 2024-11-29
Added support for binary files to `to-directory-tree`

mmhat:
Since `Byte` is part of the Dhall language we can produce binary files with the `to-directory-tree` command as well.
This PR implements that feature.

It also deprecates the `file` constructor used to construct files containing `Text` in favor of `text-file` and `binary-file` -- The former is what `file` used to be while the latter is for files containing `Byte`s.


mmhat:
It happened again: The Hydra failure appears unrelated to this PR :disappointed:

Gabriella439:
Sorry about the Hydra failures 😓 

I'll try to find a way to fix them (or move the build off of Hydra)

Issue #2567, opened 2024-02-18
Build failure on GHC 9.8.1: 'Illegal invisible type variable binder'


Vekhir:
Hi,
when building `dhall` 1.42.1 on Arch Linux with GHC 9.8.1 and `template-haskell` 2.21.0.0, the following error occurs while building the tests:

```
tests/Dhall/Test/TH.hs:111:1: error: [GHC-58589]
    Illegal invisible type variable binder: @b0
    Suggested fix: Perhaps you intended to use TypeAbstractions
    |
111 | Dhall.TH.makeHaskellTypesWith (Dhall.TH.defaultGenerateOptions
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```
The same error occurs 4 times (see Appendix). The error itself suggests to use TypeAbstractions to fix the issue.\
The offending line is [dhall/tests/Dhall/Test/TH.hs#L111](https://github.com/dhall-lang/dhall-haskell/blob/main/dhall/tests/Dhall/Test/TH.hs#L111), and the issue doesn't seem fixed on `main`.

`template-haskell` from 2.21 onwards supports Invisible binders in type declarations, see also [the migration guide](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.8#invisible-binders-in-type-level-declarations) and [ GHC proposal #425](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0425-decl-invis-binders.rst).\
`dhall` 1.42.1 advertises support for `template-haskell` 2.21 and GHC 9.8 - not sure what's going on there.

-- Vekhir

## Appendix
```
[ 5 of 21] Compiling Dhall.Test.TH    ( tests/Dhall/Test/TH.hs, dist/build/tasty/tasty-tmp/Dhall/Test/TH.dyn_o )

tests/Dhall/Test/TH.hs:111:1: error: [GHC-58589]
    Illegal invisible type variable binder: @a0
    Suggested fix: Perhaps you intended to use TypeAbstractions
    |
111 | Dhall.TH.makeHaskellTypesWith (Dhall.TH.defaultGenerateOptions
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...

tests/Dhall/Test/TH.hs:111:1: error: [GHC-58589]
[...]

PR #2542, opened 2023-10-16
compatibility with template-haskell-2.21.0 and ghc 9.8

locallycompact:

PR #2622, opened 2024-12-16
Fix TyVar BndrVis

TeofilC:
<https://github.com/dhall-lang/dhall-haskell/pull/2542> allowed `dhall` to compile with template-haskell-2.21 and GHC-9.8 by adapting to the addition of a binder visibility field on type variables.

Previously all binders were taken to be required, but in GHC-9.8 the possibility of invisible binders was introduced.

The above patch mistakenly set all binders generated by Dhall to be invisible, rather than default value of required. This changes the semantics of the code and broke some examples in the test suite.

This patch fixes this by correctly setting binders to be BndrReq.

Resolves https://github.com/dhall-lang/dhall-haskell/issues/2567

### DIFF
--- a/dhall/src/Dhall/TH.hs
+++ b/dhall/src/Dhall/TH.hs
@@ -265,7 +265,7 @@ toDeclaration generateOptions@GenerateOptions{..} haskellTypes typ =
         interpretOptions = generateToInterpretOptions generateOptions typ
 
 #if MIN_VERSION_template_haskell(2,21,0)
-        toTypeVar (V n i) = Syntax.PlainTV (Syntax.mkName (Text.unpack n ++ show i)) Syntax.BndrInvis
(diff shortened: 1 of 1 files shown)


TeofilC:
You should also consider whether it's worth deprecating versions of dhall between this bug being introduced and fixed as users using the TH functions with GHC 9.8+ will have generated broken code

ysangkok:
This is great, thank you! Tagging @Gabriella439, not sure if you're subscribed.

Gabriella439:
I am subscribed!  I receive email notifications for PRs opened against this repository

ysangkok:
Submitted

* https://github.com/commercialhaskell/lts-haskell/pull/393

to get dhall-1.42.2 added to the current Stackage LTS for GHC 9.8. That includes this patch

PR #2625, opened 2025-01-04
updated documentation strings to include empty tag in a union

amkhlv:
The existing documentation on `union` decoder misses the case when one of constructors is an empty tag. This patch adds an example of that.


dhall/src/Dhall/Marshal/Encode.hs
```diff
@@ -963,11 +965,13 @@ injectStatus = adapt >$< unionEncoder
...
+  >|< encodeConstructorWith "Unreachable" inject
   )
   where
-    adapt (Queued  n) = Left n
-    adapt (Result  t) = Right (Left t)
-    adapt (Errored e) = Right (Right e)
+    adapt (Queued  n) = Just (Left n)
+    adapt (Result  t) = Just (Right (Left t))
+    adapt (Errored e) = Just (Right (Right e))
+    adapt Unreachable = Nothing
```
Gabriella439, line 974:
```suggestion
    adapt (Queued  n) = Left n
    adapt (Result  t) = Right (Left t)
    adapt (Errored e) = Right (Right (Left e))
    adapt Unreachable = Right (Right (Right ()))
```

dhall/src/Dhall/Marshal/Encode.hs
```diff
@@ -978,11 +982,13 @@ injectStatus = adapt >$< unionEncoder
...
+  >|< encodeConstructor "Unreachable"
   )
   where
-    adapt (Queued  n) = Left n
-    adapt (Result  t) = Right (Left t)
-    adapt (Errored e) = Right (Right e)
+    adapt (Queued  n) = Just (Left n)
+    adapt (Result  t) = Just (Right (Left t))
+    adapt (Errored e) = Just (Right (Right e))
+    adapt Unreachable = Nothing
```
Gabriella439, line 991:
```suggestion
    adapt (Queued  n) = Left n
    adapt (Result  t) = Right (Left t)
    adapt (Errored e) = Right (Right (Left e))
    adapt Unreachable = Right (Right (Right ()))
```

dhall/src/Dhall/Marshal/Encode.hs
```diff
@@ -943,6 +943,7 @@ encodeFieldWith name encodeType = RecordEncoder $ Dhall.Map.singleton name encod
 data Status = Queued Natural
             | Result Text
             | Errored Text
+            | Unreachable ()
```
Gabriella439, line 946:
```suggestion
            | Unreachable
```

Gabriella439 (commented):
I think this is what you want

dhall/src/Dhall/Marshal/Encode.hs
amkhlv: I think I am getting a bit lost. The [documentation on Hackage](https://hackage.haskell.org/package/dhall-1.42.1/docs/Dhall-Marshal-Decode.html) --- I think it is different from the docstring in the source file which I have been editing. Is this a newer version? What does the `adapt` function do? 
That's why I wanted to compile it, to see the up-to-date documentation. But I could not compile it...

Gabriella439: Yeah, it's been a while since the last release to Hackage, which is the reason for the difference.   However, I'm working on cutting a new release soon

dhall/src/Dhall/Marshal/Encode.hs
amkhlv: This is how I had it: 

`data DateFinished = Current () | Done DATE.Day | Dropped () deriving (Generic, Show, Eq)`

and then:

```
dateFinishedDecoder :: Decoder DateFinished
dateFinishedDecoder =
  union
    ( (Current <$> constructor "Current" unit)
        <> (Done <$> constructor "Done" day)
        <> (Dropped <$> constructor "Dropped" unit)
    )
```
But if I do as you suggested, _i.e._  `data DateFinished = Current  | Done DATE.Day | Dropped deriving (Generic, Show, Eq)`  --- then it does not compile

Gabriella439: If you change that to:

```haskell
dateFinishedDecoder :: Decoder DateFinished
dateFinishedDecoder =
  union
    ( (Current <$ constructor "Current" unit)  -- Carefully note the `<$`
        <> (Done <$> constructor "Done" day)
        <> (Dropped <$> constructor "Dropped" unit)
    )
```

… then it should compile

amkhlv: Oh, I got it ! Cute (although smells of witchcraft...). Then, maybe, instead of my suggested patch, would you rather write something pedagogical, in that docstring?

amkhlv: I just requested the pull for the correcting patch, as you suggested.

dhall/src/Dhall/Marshal/Encode.hs
amkhlv: Stackage would also be great. It would enable Hoogle...

PR #2633, opened 2025-01-20
dhall: provide `*WithIndex` instances for `Map`

ozkutuk:
This PR adds the following instances for `Dhall.Map`:
- `FunctorWithIndex`
- `FoldableWithIndex`
- `TraversableWithIndex`

`indexed-traversable` package is already a dependency of `dhall`, so this doesn't effect the dependency footprint. These instances would allow the usage of indexed optics provided by the optics libraries.


mmhat:
Nice, thank you!

PR #2641, opened 2025-02-11
DirectoryTree: Add `binary-file` to support creating files with `Bytes` content

mmhat:


mmhat:
Corresponding change in Prelude: https://github.com/dhall-lang/dhall-lang/pull/1398

Gabriella439 (approved):

PR #2642, opened 2025-02-13
Use `microlens` package family instead of `lens-family-core`

### LINKED ISSUE #998: Eventual fate of the Dhall.Optics module
I think it's a bit silly for `dhall` to have its own mini-`lens` - it's a violation of the separation of concerns. At the same time, the combinators are useful, and avoiding a dependency on `lens` does have value.

Possible directions:

* Do nothing; the situation isn't terrible. A few combinators being copy-pasted isn't the worst thing.

* Bite the bullet and incur a `lens` dependency. `dhall` is already a heavy package and I think the `lens` dependency only amounts to a few additional packages (there's already a dependency on `profunctors`, though not on `semigroupoids`), though `lens` itself is a meaty compile. The upside is getting to use all the fun things in `lens`.

* Try to get the combinators upstreamed into `lens-family-core`, which is already a dependency. I don't know how receptive that package is to implementing combinators from `lens`.

* Move to `microlens` instead of `lens-family-core`. This package definitely does accept random combinators from `lens`, and it's similarly lightweight. I don't know whether `lens-family-core` might have some countervailing advantage, though.

mmhat:
The switch is motivated by

- the desire to get rid of the `profunctors` dependency.
- I think we should could slim down `Dhall.Optics` (and deprecate it eventually).

A related comment of mine is here: https://github.com/dhall-lang/dhall-haskell/issues/1102#issuecomment-901061431

Fixes #998


mmhat:
Depends on https://github.com/stevenfontanella/microlens/pull/185

Gabriella439 (approved):

TristanCacqueray:
Well done @mmhat , that's great, one less lens library to think about!

PR #2646, opened 2025-02-28
`Dhall.Import` refactorings

mmhat:
This PR includes the following changes:

- Use more lenses in `Dhall.Import` and friends. My hope is that we can ultimately move to a MTL approach in `Dhall.Import`, i.e. use `(MonadIO m, MonadState Status m) => m a` instead of `StateT Status IO a`.

- I moved `EvaluateSettings` and `InputSettings` to a new (internal) module `Dhall.Settings`. This enables us to have a field `_evaluateSettings :: EvaluateSettings` in `Dhall.Import.Types.Status`, since we need all the fields of `EvaluateSettings` during import resolution anyway.

- A new `HasInputSettings` class similiar to `HasEvaluateSettings`. The functions for the different phases in the `Dhall` module now take a value of `HasInputSettings s => s` instead of `InputSettings` accordingly.


winitzki:
These changes are now taken up here: https://github.com/dhall-lang/dhall-haskell/pull/2754

PR #2655, opened 2025-05-13
Add Date and Time component extraction builtins

kukimik:
Implements https://github.com/dhall-lang/dhall-lang/pull/1408.

I have doubts about the Nix translations of the builtins. I follow the example given by `Date/show` and `Time/show`, but it just doesn't feel right. (I also don't know Nix.)


winitzki:
@kukimik  ~There is a build failure with `stack.ghc-9.2.yaml` where it seems that `unix-time-0.4.10` needs to be added to `extra-deps` in that yaml file.  Could you please try that?~

I updated the branch with latest fixes in master so that CI works reliably now, please update your fork as well if needed.

winitzki:
@kukimik Now with these new primitives, we can implement comparison for `Date` (but not for `Time` as the fractional seconds are not extracted). Should we add `Date/equal` and `Date/lessThan` and other such functions to the Prelude?

kukimik:
> Should we add `Date/equal` and `Date/lessThan` and other such functions to the Prelude?

In my opinion: yes, it's the natural next step.

winitzki (approved):
LGTM

PR #2659, opened 2025-07-09
Relax upper bounds

tbidne:
Relaxes upper bounds for `containters`, `data-default`, `optparse-applicative`, and `QuickCheck`. Requires no code changes, so can be done with hackage revisions.

Thanks!

### DIFF
--- a/dhall-bash/dhall-bash.cabal
+++ b/dhall-bash/dhall-bash.cabal
@@ -29,7 +29,7 @@ Library
     Build-Depends:
         base                      >= 4.11.0.0 && < 5   ,
         bytestring                               < 0.13,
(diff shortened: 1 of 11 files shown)


ysangkok:
@mmhat Would you be able to merge this? Then dhall could get back into Stackage.

ulysses4ever:
This is also holding back nixpkgs update to the latest Stackage https://github.com/NixOS/nixpkgs/pull/429810

ysangkok:
@ulysses4ever There is a larger issue holding back nixpkgs. dhall-nix needs hnix, but hnix-0.17.0 currently has the bound template-haskell <2.22` which means it excludes GHC 9.10+.

PR #2666, opened 2025-09-19
dhall-json: implement --type mode for encoding type as json

amosr-msft:
Hello,
I'm working on a project where we are generating datatypes for a given Dhall schema. It would be helpful for us if dhall-json could generate a JSON representation of types, as well as values. Would you be interested in adding something like this?

This PR has two changes: I've modified dhallToJSON to be able to emit value types, and added a `--type` option to `dhall-to-json`.

I think this representation should be unambiguous and round-trippable, but I'm happy to use any representation. I have only briefly read about JSON schemas, to be honest, but I thought that a Dhall-specific type encoding would be better to faithfully preserve the original Dhall type.

The extra `--type` option is just a shorthand which might be useful, but we can easily make do without the option if it's not likely to be useful to anyone else.

Here's an example showing how the output looks: for an input schema:
```
{
    field: Text,
    nested: { nested_field: Natural },
    list: List Bool,
    optional: Optional Text,
    union: < A: Natural | B: Text | C >
}
```

This gives the following JSON type:
```
$ dhall-to-json --file dhall/schema_test.dhall
{
  "fields": {
    "field": "Text",
    "list": {
      "element": "Bool",
      "type": "List"
    },
    "nested": {
      "fields": {
        "nested_field": "Natural"
      },
      "type": "Record"
    },
    "optional": {
      "element": "Text",
      "type": "Optional"
    },
    "union": {
      "choices": {
        "A": "Natural",
        "B": "Text",
        "C": {}
      },
      "type": "Union"
    }
  },
  "type": "Record"
}
```

### DIFF
--- a/dhall-json/tasty/Main.hs
+++ b/dhall-json/tasty/Main.hs
@@ -34,6 +34,7 @@ testTree =
         , testDhallToJSON "./tasty/data/emptyObjectStrongType"
         , testDhallToJSON "./tasty/data/toArbitraryJSON_12_0_0"
         , testDhallToJSON "./tasty/data/toArbitraryJSON_13_0_0"
+        , testDhallToJSON "./tasty/data/types"
         , testJSONToDhall "./tasty/data/emptyAlternative"
         , testJSONToDhall "./tasty/data/emptyObject"
         , testJSONToDhall "./tasty/data/emptyList"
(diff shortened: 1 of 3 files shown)


amosr-msft:
Hi @Gabriella439 , I hope I'm not pestering you, I just wanted to ping you in case you'd missed this - do you think this would be a reasonable feature?

PR #2711, opened 2026-03-02
Bugfix: Text/Replace becomes a no-op when its input contains appends

nikita-volkov:
I've managed to manually reproduce the bug and added a regression test. Then I fed it to an LLM to fix it.

In [this run](https://github.com/dhall-lang/dhall-haskell/actions/runs/22578430487/job/65403665045?pr=2711) you'll see the reproduction causing a failure.

Following is the report about the fix produced by the LLM.

## Root Cause

The bug was in `Normalize.hs`, in two places inside the `Text/replace` normalization logic.

**The scenario:** `Text/replace "." "!" (x ++ "!")` is normalized with `x` as a free variable. `x ++ "!"` becomes `TextLit (Chunks [("", Var "x")] "!")` — a text literal with one interpolated chunk.

**Bug 1 (lines 332–345):** The case that handles `(needle: plain text, replacement: plain text, haystack: any TextLit)` applied `Text.replace` to all static string parts of the chunks but left the interpolated expressions (`y` in each `(x, y)` pair) untouched and returned the result as if the replacement was complete. Since neither `""` nor `"!"` contains `"."`, the result was identical to the input — the `Text/replace` was silently dropped. When `x = "..."` was substituted later, there was no longer a `Text/replace` to handle the dots.

**Bug 2 (line 378):** In the chunked-haystack case (lines 363–385), when the needle is not found in `firstText`, the code emitted `firstInterpolation` verbatim without wrapping it in `Text/replace`. Same problem: the replacement was never applied to the interpolated expression's eventual value.

## Fix

Two targeted edits to `Normalize.hs`:

1. **Case 332–345**: Narrowed the pattern from `TextLit (Chunks xys z)` to `TextLit (Chunks [] z)` — it now only fires for pure-text haystacks (no interpolations). Haystacks with interpolations fall through to the chunked case.

2. **Line 378 (`if Text.null suffix then`)**: Changed `firstInterpolation` to `App (App (App TextReplace needle) replacement) firstInterpolation` — wrapping it in a `Text/replace` application so that when the interpolation becomes a concrete text, the replacement is applied to it.


nikita-volkov:
I've just discovered one more related bug. I'll post updates soon

winitzki:
I wonder - is this bug also present in the standard specification for `Text/replace`, or is it just an implementation bug?

dhall/src/Dhall/Normalize.hs
```diff
@@ -333,16 +333,10 @@ normalizeWithM ctx e0 = loop (Syntax.denote e0)
...
-                                    (x, y) <- xys
-
-                                    let x' = Text.replace needleText replacementText x
-                                    return (x', y)
-
+                        (TextLit (Chunks [] z)) -> do
                             let z' = Text.replace needleText replacementText z
 
-                            return (TextLit (Chunks xys' z'))
+                            return (TextLit (Chunks [] z'))
```
Gabriella439: This change is good, although it might be possible to also delete the entirety of this branch of the case expression, too, given that I believe it is subsumed by the next branch now.

nikita-volkov: Done

winitzki (approved):
Looks good; later we need to revise the code in Normalize.hs again, to make sure no duplication exists between Eval.normalize and normalizeWithM

PR #2716, opened 2026-03-12
Add CSE

nikita-volkov:
Addresses https://github.com/dhall-lang/dhall-haskell/issues/2715

(LLM-generated and edited by me).


dhall/src/Dhall/Core/CSE.hs
```diff
@@ -0,0 +1,222 @@
...
+-}
+cse :: forall a. Ord a => Expr Void a -> Expr Void a
+cse expr = go freshNames candidates expr
+  where
+    -- Count occurrences of each subexpression using cosmosOf to traverse all levels
+    counts :: Map (Expr Void a) Int
+    counts = Map.fromListWith (+)
+        [ (sub, 1)
+        | sub <- Lens.foldMapOf (Lens.cosmosOf subExpressions) (:[]) expr
+        ]
```
Gabriella439: I think this will not correctly handle the following case:

```
let x  = 1

let y = nonTrivialExpressionOf x

let x = 2

in  nonTrivialExpressionOf x
```

If I understand correctly, the logic would count both occurrences of `nonTrivialExpressionOf x` as the same sub-expression even though `x` would differ in both cases.

My intuition here is that this would require explicit recursion over the syntax tree to correctly handle this.

PR #2734, opened 2026-04-29
Fix a space leak by forcing the decoded expression to normal form

### LINKED ISSUE #1420: Space leak leading to OOM
Having a file named `bug.dhall` with the following contents:

```dhall
https://raw.githubusercontent.com/pgenie-io/rust.gen/3cd5f94f82de7491b77b39a0a9baaad02ee49413/gen/Gen.dhall
```

and running `dhall freeze bug.dhall` leads to the machine acquiring all available memory. The issue is not only with the freeze command, but it's one way to reproduce it.

nikita-volkov:
# The bug

This fixes https://github.com/dhall-lang/dhall-lang/issues/1420. Details follow.

# Root Cause

The leak is in decodeExpression (Dhall.Binary.decodeExpression), specifically in how it returns the decoded expression.

decodeExpression uses decodeExpressionInternal, which recursively builds the Expr tree inside cborg's CPS Decoder monad. The Decoder monad sequences operations strictly, but the result values are constructed lazily. This means decodeExpressionInternal does not return a fully-evaluated expression tree; it returns a massive tree of thunks — unevaluated closures representing the decoded AST.

When loadWith later traverses this tree (via bindingExprs on Let bindings, or unsafeSubExpressions on other constructors), it forces these thunks one by one. Because the tree is lazy, the following happens:

The unevaluated thunk tree and the forced result coexist in memory during traversal. Each thunk retains references to the decoder's continuation closures and intermediate CBOR decoding state. For a large expression this laziness causes memory usage to balloon far beyond the size of the expression itself. The profiling showed decodeExpression as the hotspot because that's where the lazy thunk tree originates. The stack trace decodeExpression ← bindingExprs ← loadWith confirms that loadWith is the code path that forces the thunks, exposing the leak.

# The Fix

The fix is to force the decoded expression to normal form immediately, before returning it from decodeExpression. This evaluates all the thunks up front, so loadWith traverses a compact, fully-evaluated data structure instead of a lazy thunk tree.

### DIFF
--- a/dhall/src/Dhall/Binary.hs
+++ b/dhall/src/Dhall/Binary.hs
@@ -25,6 +25,7 @@ import Codec.CBOR.Decoding  (Decoder, TokenType (..))
 import Codec.CBOR.Encoding  (Encoding)
 import Codec.Serialise      (Serialise (decode, encode))
 import Control.Applicative  (empty, (<|>))
+import Control.DeepSeq      (NFData, force)
 import Control.Exception    (Exception)
 import Data.ByteString.Lazy (ByteString)
 import Dhall.Syntax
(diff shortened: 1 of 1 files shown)


nikita-volkov:
Looks like I made a mistake. It doesn't fix anything.

Gabriella439:
no worries! ^^

PR #2743, opened 2026-05-19
Fix normalization with Kimi

nikita-volkov:
## Summary

Implements **Proposal 1 (Core)**: Dhall normalization now preserves outer `let` bindings around lambda abstractions when the let variable is not shadowed by the lambda parameter and is not free in the lambda body. This prevents loss of sharing for top-level definitions.

#### `dhall/src/Dhall/Eval.hs`
- Extended `Environment` with two new constructors:
  - `LetBind !(Environment a) !Text (Val a)` — tracks a let binding in the evaluation environment
  - `Extracted !(Environment a) !Text` — marks a binding as already extracted during quoting
- Modified `eval` on `Let` to push `LetBind` into the environment instead of inlining the bound value
- Added `extractLetBinds` to collect preserved let bindings from the closure environment during `quote` of `VLam`, and wrap the quoted lambda with them

#### `dhall/src/Dhall/Normalize.hs`
- Updated `isNormalized` to recognize preserved-let shapes (`let x = a in λ(y : T) → b`) as normal
- Updated `normalizeWithM` to mirror the new `eval` behavior for let-preservation

#### `dhall/src/Dhall/DirectoryTree.hs`
- Added `collectOuterLets` to strip leading `Let` bindings
- Updated `toDirectoryTree` pattern match to recognize `Let...Lam...Lam...` shape (in addition to the old `Lam...Lam...`)
- Rewrote `decodeDirectoryTree` to handle the new normalized shape:
  - Works with the original (non-α-normalized) expression to avoid de Bruijn index corruption
  - Substitutes let values into the lambda body using `Dhall.Normalize.subst`
  - Normalizes the result, then manually renames the `make` parameter to `_` so the `Make` pattern in the decoder matches

#### `dhall/src/Dhall/Import.hs`
- Added `loadWithPreserving` variant that keeps `sha256:` imports as `Embed` nodes instead of resolving them, preserving the let-bound import structure

### Test Results
- All 1790 `tasty` tests pass
- All 19 QuickCheck property tests pass (including `isNormalizedIsConsistentWithNormalize`)
- All 7 `to-directory-tree` tests pass (including the 3 fixpointed tests that were failing)

### Known Follow-up
The `x == param` guard in `extractLetBinds` interacts carefully with `alphaNormalize` (which renames all bound variables to `_`). The `decodeDirectoryTree` fix sidesteps this by operating on the original expression before α-normalization.

### DIFF
--- a/dhall/src/Dhall/DirectoryTree.hs
+++ b/dhall/src/Dhall/DirectoryTree.hs
@@ -34,7 +34,8 @@ import Dhall.DirectoryTree.Types
 import Dhall.Marshal.Decode      (Decoder (..), Expector)
 import Dhall.Src                 (Src)
 import Dhall.Syntax
-    ( Chunks (..)
+    ( Binding (Binding)
+    , Chunks (..)
     , Const (..)
     , Expr (..)
     , RecordField (..)
@@ -51,6 +52,7 @@ import qualified Data.Text.IO                as Text.IO
 import qualified Dhall.Core                  as Core
 import qualified Dhall.Map                   as Map
 import qualified Dhall.Marshal.Decode        as Decode
+import qualified Dhall.Normalize
 import qualified Dhall.Pretty
 import qualified Dhall.TypeCheck             as TypeCheck
 import qualified Dhall.Util                  as Util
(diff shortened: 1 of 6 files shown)


nikita-volkov:
Meant to PR into my fork. This is only an experiment.

PR #2745, opened 2026-05-27
Make the warning-reporting controllable

nikita-volkov:
In my app that executes a Dhall code while also printing to `stderr` I've noticed garbled output like the following:

```
Warni
ng[1[;03m3:m WCaorunlidn gno[t0 mg:e tC oourl dc rneoatt eg etth eo rd ecfraeualtte  ctahceh ed edfiaruelctt ocrayc:h
e
 ↳d i/rUescetrosr/yr:u
n
n↳e r//U.scearcsh/er/udnhnaelrl/
.
cYaocuh ec/adnh aelnla
b
lYeo uc accahni negn abbyl ec rceaacthiinngg  ibty  icfr enaeteidnegd  iatn di fs enteteidnegd  raenadd ,s
ewtrtiitneg  arneda ds,e
awrrciht ep earnmdi ssseiaorncsh  opne rimti sosri opnrso voind iintg  oarn optrhoevri dciancgh ea nboatshee
rd icraecchteo rbya sbey
 dsiertetcitnogr yt hbey  $sXeDtGt_iCnAgC HtEh_eH O$MXED Ge_nCvAiCrHoEn_mHeOnMtE  veanrviiarbolnem.e
n
t
 variable.
```

My analysis led me to discovery that Dhall prints warnings directly to `stderr`, which results in competing writes to the same output stream leading to such garbling.

This PR replaces writes to `stderr` with a configurable action, which by default behaves the old way.

### DIFF
--- a/dhall/src/Dhall/Util.hs
+++ b/dhall/src/Dhall/Util.hs
@@ -118,7 +118,7 @@ _WARNING :: IsString string => string
 _WARNING = "\ESC[1;33mWarning\ESC[0m"
 
 -- | Output a warning message on stderr.
-printWarning :: (MonadIO m) => String -> m ()
+printWarning :: (MonadIO m) => Text -> m ()
 printWarning message = do
     let warning =
                 "\n"
@@ -126,7 +126,7 @@ printWarning message = do
             <> ": "
             <> message
 
-    liftIO $ IO.hPutStrLn IO.stderr warning
+    liftIO $ Data.Text.IO.hPutStrLn IO.stderr warning
 
 get
     :: (String -> Text -> Either ParseError a)
--- a/dhall/tests/Dhall/Test/Import.hs
+++ b/dhall/tests/Dhall/Test/Import.hs
@@ -156,10 +156,13 @@ successTest prefix = do
         let status = Import.emptyStatus directoryString
 #endif
 
+        let status' =
+                status { Import._reportWarning = \_ -> return () }
+
         let load =
                 State.evalStateT
                     (Test.Util.loadWith actualExpr)
-                    status
+                    status'
 
         let usesCache = [ "hashFromCache"
                         , "unit/asLocation/Hash"
--- a/dhall/src/Dhall.hs
+++ b/dhall/src/Dhall.hs
@@ -34,6 +34,7 @@ module Dhall
     , substitutions
     , normalizer
     , newManager
+    , reportWarning
     , defaultInputSettings
     , InputSettings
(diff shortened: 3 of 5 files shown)


dhall/tests/Dhall/Test/Import.hs
nikita-volkov, line 1: I've updated the old test-suite that was producing noisy warnings in the output. This also showcases how this new feature can be used.

dhall/src/Dhall/Util.hs
```diff
@@ -118,15 +118,15 @@ _WARNING :: IsString string => string
 _WARNING = "\ESC[1;33mWarning\ESC[0m"
 
 -- | Output a warning message on stderr.
-printWarning :: (MonadIO m) => String -> m ()
+printWarning :: (MonadIO m) => Text -> m ()
```
winitzki, line 121: What is the significance of changing `String` to `Text` here?

winitzki (commented):
Generally I believe it's a good change. I have a question: can we make output to stderr somehow automatically chunked so that concurrent printing won't produce jumbled results? I'm not so familiar with the logging tools in the Haskell ecosystem but, say, on JVM this sort of thing just never happens - all printing to stderr or stdout is synchronized by default.

dhall/src/Dhall/Util.hs
nikita-volkov: It adapts to the signatures of the only two places where this function is used now:

- https://github.com/dhall-lang/dhall-haskell/pull/2745/changes#diff-7a43286e905afe6d1e9df0f9ff0f4a57cc74fa220b7dd45462d7056f52c05df5R137
- https://github.com/dhall-lang/dhall-haskell/pull/2745/changes#diff-174197e43c505a99bd44ddec58375d7f4c082123c8e7241c39d7ad49809080faR134

`String` is gradually getting replaced by `Text` in the Haskell community and most of the definitions in this codebase are already `Text`-oriented.

nikita-volkov:
> can we make output to stderr somehow automatically chunked so that concurrent printing won't produce jumbled results? I'm not so familiar with the logging tools in the Haskell ecosystem but, say, on JVM this sort of thing just never happens - all printing to stderr or stdout is synchronized by default.

We'd need exclusive access to the `stderr` handle to achieve that. Usually this is solved via abstractions involving mutex or channel. That requires control over the code outside of this library, which is why the solution is to just delegate that to the user, which is what this PR does. Besides it also let's the user process this output, redirect it or completely silence it.

PR #2746, opened 2026-05-30
fix(dhall-lsp-server): don't loudly fail when VFS URI is not found

SrGaabriel:
While trying to port `dhall-lsp-server` to Zed, I noticed that whenever I first made an edit to a file that was not tracked (any non-`.dhall` file basically), I would get the following error followed by a crash:

```
dhall-lsp-server: user error (Could not find Uri {getUri = "path-to-non-dhall-file-here"} in VFS.)
```

I believe the reason this can't be reproduced in VSCode (I didn't find any related issues/PRs) is likely because Zed, in contrast to VSCode, sends `didSave` for files it did not send `didOpen` for (non tracked files).

I am creating this PR as a draft because I'm not sure this is the most principled fix (would appreciate feedback) and to see how CI goes.

### DIFF
--- a/dhall-lsp-server/src/Dhall/LSP/Handlers.hs
+++ b/dhall-lsp-server/src/Dhall/LSP/Handlers.hs
@@ -118,10 +118,10 @@ tryReadUri uri_ = do
 #endif
     Nothing -> Nothing
 
--- | Like `readUriMay`, but fails if the URI was not found.
+-- | Like `tryReadUri`, but fails if the URI was not found.
 readUri :: Uri -> HandlerM Text
 readUri uri_ = do
-  mText <- readUriMay uri_
+  mText <- tryReadUri uri_
   case mText of
     Just text -> return text
     Nothing ->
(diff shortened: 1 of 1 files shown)


winitzki (approved):
Looks good.

winitzki:
@SrGaabriel  One comment: Currently the new code just silently omits the unknown URIs (that previously would have generated a crash). What if we printed a warning instead? Would that warning become too verbose when using Zed in your project?

SrGaabriel:
> @SrGaabriel One comment: Currently the new code just silently omits the unknown URIs (that previously would have generated a crash). What if we printed a warning instead? Would that warning become too verbose when using Zed in your project?

I think warnings/messages would be too verbose since they are very common, but stderr logs should be fine. What do you think?

winitzki:
> > @SrGaabriel One comment: Currently the new code just silently omits the unknown URIs (that previously would have generated a crash). What if we printed a warning instead? Would that warning become too verbose when using Zed in your project?
> 
> I think warnings/messages would be too verbose since they are very common, but stderr logs should be fine. What do you think?

I'm just thinking that it's better not to fail silently in case the failure is important to the user. So, yes, stderr logs would be a solution if you don't mind implementing it.

SrGaabriel:
Sorry about the delay, I see you have already implemented it! I've been a bit busy recently and forgot, but thanks for merging anyway!

winitzki:
@SrGaabriel  Let me know if you have a working dhall extension for Zed! I'm curious to try.

SrGaabriel:
> @SrGaabriel  Let me know if you have a working dhall extension for Zed! I'm curious to try.

I do! It's available on the Zed extension store if you want to try it out :)

PR #2753, opened 2026-06-26
issue 1425 - implement a new special form readConstructor

winitzki:
The new special form `readConstructor` has 1 argument, which must be an enum-like union type.

It is a parse error if `readConstructor` is unsaturated. It's a type error if `readConstructor` is used on a non-enum-like union type or on a type that is not a union type.

```dhall
let a = readConstructor -- parse error
let b = readConstructor Bool -- type error
let c = readConstructor < A | B : Bool > -- type error
let d : Text → Optional <A | B > = readConstructor < A | B > -- OK
let e = d "A"  -- evaluates to Some < A | B >.A
let f = d "C"  -- evaluates to None < A | B >
```


nikita-volkov (approved):
Perfect! This will likely resolve most of my issues!

PR #2754, opened 2026-06-27
lang-1428 support imports `as Source`

winitzki:
For design considerations and usage examples, see https://github.com/dhall-lang/dhall-lang/issues/1428

- A new import syntax `as Source`
- An import `as Source` stops expanding all transitive imports 
- Hash-protected transitive imports are kept unchanged (not inlined) in the cached artifact
- Non-hash-protected transitive imports are inlined but still not expanded to normal forms

- [x] Performance testing to validate that `as Source` speeds up evaluation and reduces memory consumption
- [x] A full review of the import mechanism and better documentation


kindlnatuvion:
Hi Sergei,


this PR looks really exciting and I hope this will speed up some things.
I have read your comments on [#1428](https://github.com/dhall-lang/dhall-lang/issues/1428) but I have not fully understood how the setup should look like to get the most performance out of it.

Currently we often have files looking like this:
```
let Prelude = ./lib/Prelude/package.dhall
let CustomPrelude = ./lib/CustomPrelude/package.dhall

in someSmallExpressionUsingPreludeAndCustomPrelude
```
and we pay a big price just for the evaluation of the preludes.
A file like this can take up to 4 minutes on a cold cache.

So first I just changed to `as Source` imports, so that they would look like this:
```
let Prelude = ./lib/Prelude/package.dhall as Source
let CustomPrelude = ./lib/CustomPrelude/package.dhall as Source
```
However, reading in the file now had the same performance as before.
It seems this was to be expected, because we still need to make changes to the preludes transitively, if I understood that correctly.

But how should the setup look like?
Currently the imports inside the preludes are frozen with the cache option, looking like this:
```
let List/map =
        missing
          sha256:dd845ffb4568d40327f2a817eb42d1c6138b929ca758d50bc33112ef3c885680
      ? ./map.dhall
```

I wrote a small tool to change all imports in a file to `as Source` imports, so they would instead look like this:
```
let List/map = ./map.dhall as Source
```
But now reading in the file even had a little worse performance than before, taking around 6 minutes.

Well, the tool removed the hashes, so as a third step I ran freeze with the `--secure` option, which changed the imports to look like this:
```
let List/map =
      ../List/map.dhall
        sha256:5b43b1207f0c5f69e80a94bf78d52a2b2189b2658a70652cb805001ced08b5ae as Source
```
But now, reading in the file on a cold cache will error with integrity checks failing.
The issue seems to be that previously adding hashes to imports did not change the hash of the file itself.
However with `as Source` they do.
Let's say we have following import chain:
`main.dhall --imports--> Prelude/package.dhall --imports--> List/map.dhall`
and we first freeze main.dhall by adding the hash for Prelude/package.dhall.
The hash will be incorrect as soon as we edit Prelude/package.dhall and add the hash for List/map.dhall.


So I am a bit confused about how the setup should look like. I am happy to try out some things and provide more info if necessary.


Best regards,
Janosch

winitzki:
@kindlnatuvion   Thank you for your feedback. Here are my comments:

1. First I'd like to be able to reproduce the slowness. I created a small file just with the standard prelude:

```dhall
let Prelude = ./dhall/dhall-lang/Prelude/package.dhall

let sum1 : List Natural → Natural
    = λ(xs : List Natural) → Prelude.List.fold Natural xs Natural Prelude.Operator.`+` 0
let filter1 = Prelude.List.filter Natural Prelude.Natural.even
let map1 = Prelude.List.map Natural Natural (λ(x : Natural) → x + 1)
let generate1 = Prelude.List.generate 1000 Natural (Prelude.Function.identity Natural)

in  sum1 (filter1 (map1 generate1))
```

The run time is about 0.3 seconds, even after removing the entire cache (`$HOME/.cache/dhall-haskell` and `$HOME/.cache/dhall`), even though the Prelude is used seven times in this expression. Most of this run time is reading the Prelude from the Web or from the cache.

Fully evaluating the Prelude alone takes about 0.3 seconds with or without cache (only about 5% faster with cache). This is with the master branch, not with this PR.

So, this computation is dominated by the evaluation of the Prelude.

With this PR, adding `as Source` to the Prelude import makes the evaluation time 0.6 seconds - which is 2x slower. (Whether with or without pre-existing cache.) This is interesting; I didn't expect this to become slower.

Doing `dhall freeze --all` on this file speeds up considerably; but still `as Source` is about 2x slower than the ordinary import after freezing.

Could you suggest a test that takes several minutes with Prelude? I don't know what your "custom prelude" contains, maybe there are very large types or something like that.

There exist old issues that talk about slowness with Dhall-Kubernetes, due to very large types and repeated typechecking. Maybe that's similar to what you have?

2. If you fully evaluate a dhall file, you compute its normal form. The result requires evaluating all imports as well. When imports are `as Source` it means they won't be expanded to normal form at the time of importing or caching; but the final expression still needs to be expanded to normal form. If the final result is small then I would expect a speedup when imports are done `as Source`.

3. Indeed I see a problem that freezing of an import will change the hash of the parent file. Previously this was not the case, because imports were always fully inlined and expanded. Now, a hash-protected import `as Source` contains the hash of the imported code, rather than the full inlined and normalized code. I will think about it some more.

4. My idea was that the user won't need to change the Prelude (or another library) to have `as Source` in it; just import the library itself `as Source` from user scripts. This was designed to avoid expansion during import time, and it did work in some initial tests. When the library is imported `as Source` then all transitive imports within in the library will be also imported without full normalization. I realize now that there is a question: does this break existing code such as the Prelude with existing hash-protected imports. If it does, the approach `as Source` needs to be redesigned. We must be able to load all existing code `as Source`, with or without cache. 

So far I didn't see any existing code getting broken. Here is a shorter example:

```dhall
-- This file is test1.dhall - evaluation 0.335s without cache, 0.260s with cache
let Prelude = ./dhall/dhall-lang/Prelude/package.dhall
in Prelude.List.fold
```

```dhall
-- This file is test2.dhall - evaluation 0.595s without cache, 0.584s with cache
let Prelude = ./dhall/dhall-lang/Prelude/package.dhall
    as Source
in Prelude.List.fold
```


```dhall
-- This file is test3.dhall - evaluation 0.327s without cache, 0.034s with cache
let Prelude = ./dhall/dhall-lang/Prelude/package.dhall
    sha256:931cbfae9d746c4611b07633ab1e547637ab4ba138b16bf65ef1b9ad66a60b7f
in Prelude.List.fold
```


```dhall
-- This file is test4.dhall - evaluation 0.599s without cache, 0.059s with cache
let Prelude = ./dhall/dhall-lang/Prelude/package.dhall
    sha256:9c8b438f75515d0d10736eb831dd25c87709fba272fec02485263be8c73469af
    as Source
in Prelude.List.fold
```

kindlnatuvion:
Hello Sergei,

Thanks for the examples.

I reran a small test with a profiling build and it seems that the `as Source` imports shifts the loading times to parsing.
There might be a space leak somewhere, a lot of time is spent in blockComments for example.
We kind of had the assumption, that normalization would be a bottleneck and the `as Source` imports would help with that, but it seems that this is more for special cases where you want to address explosion of normalization, but not for the general case.

I can mail you the .prof files if you want.

winitzki:
@kindlnatuvion  Yes, I'd like to have the .prof files if it's not too hard. Thank you!

winitzki:
@kindlnatuvion  I've been working on optimizing the `as Source` imports. 

- In this branch now I have changed the sha256 computation so that (I believe) freezing of an import no longer changes the hash of the parent file. 

- Also I have introduced new benchmarks to track performance.
- There are some optimizations and `as Source` is generally faster than it was initially.

Could you please try the current branch with your import-heavy test and let me know how the performance looks now?

winitzki:
@kindlnatuvion  I opened another issue here https://github.com/dhall-lang/dhall-lang/issues/1437 as I realized that perhaps `?` should not mix different sha256. Maybe this will be of interest to you as well, please let me know if you have any comments.

kindlnatuvion:
Hello Sergei,

thank you for looking into the as-source imports and improving performance.
I have rebuild a profiling build and ran it again and I will send you the prof files via e-mail.

It definitely is faster now, great work!
However, just regularly importing the big custom prelude is still the fastest option.
As far as I understand, heavy imports were not the target of this feature anyway and it was more of an exploration if it would change something.

To your cache related points, I have to think more about it.

If there is anything I should change in the setup or I can help you test, let me know!

Best regards,
Janosch

winitzki:
@kindlnatuvion  Your profiler info indicates that your run is heavy on substitution work. I added special benchmarks and optimizations just for substitutions. It would be great if you could try the latest code with your workload and let me know if something has changed to the better.

winitzki:
@kindlnatuvion Could you please run your benchmark again with the latest code changes. I'm trying to reproduce your workload via benchmarks, but it's not so easy.

kindlnatuvion:
Thank you Sergei for pinging me, I will rerun it.

winitzki:
@kindlnatuvion  I pushed some more changes for speeding up "as source", and if you have time, I'd appreciate if you run the latest code here.

winitzki:
I have an impression that performance of "as source" has become worse after all those changes; while the performance of "as code" imports has improved significantly. The code has become a bunch of AI slop (generated in hopes of improving performance). At this point I will need to revisit the design of "as source" and implement cleanly; or abandon this feature if the performance of "as code" has become acceptable.

winitzki:
This PR is not to be merged. Some changes need to be extracted from it and possibly merged separately.

PR #2755, opened 2026-06-28
remove imports that cause warnings

winitzki:
- Remove some unneeded imports that cause warnings
- Set `{-# OPTIONS_GHC -Wno-unused-imports #-}` for some files where warnings are due to change in `liftA2`

### DIFF
--- a/dhall/src/Dhall/Import/Headers.hs
+++ b/dhall/src/Dhall/Import/Headers.hs
@@ -3,6 +3,8 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE ViewPatterns        #-}
 
+{-# OPTIONS_GHC -Wno-unused-imports #-}
+
 module Dhall.Import.Headers
     ( normalizeHeaders
     , originHeadersTypeExpr
--- a/dhall/src/Dhall/Import/Types.hs
+++ b/dhall/src/Dhall/Import/Types.hs
@@ -31,7 +31,6 @@ import Prettyprinter                    (Pretty (..))
 import qualified Dhall.Import.Manager
 #endif
 
-import qualified Data.Text.IO
 import qualified Dhall.Context
 import qualified Dhall.Map          as Map
 import qualified Dhall.Substitution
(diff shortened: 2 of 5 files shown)


nikita-volkov (changes requested):
No need to add it to source. This is better kept opt-in.

When you want to lint you can run something like:

```bash
cabal build --ghc-options="-Werror -Wall -Wincomplete-uni-patterns -Wincomplete-record-updates -Wredundant-constraints -Wunused-packages"
```

winitzki:
> No need to add it to source. This is better kept opt-in.
> 
> When you want to lint you can run something like:
> 
> ```shell
> cabal build --ghc-options="-Werror -Wall -Wincomplete-uni-patterns -Wincomplete-record-updates -Wredundant-constraints -Wunused-packages"
> ```

I'd like to remove the warnings that are not useful. Because of those warnings that we are forced to ignore, we will tend to ignore other warnings.

Currently one can't remove the imports because we build Dhall with very different GHC versions, some of these versions have those imports in the Prelude and others don't.

There are two possible ways out:

- reduce the number of GHC versions with which we build Dhall in CI
- silence warnings in the GHC options that CI is using

What would you propose to do?

nikita-volkov (approved):
Sorry. I misread the changes in the commit. I didn't notice the `no` in `-Wno`. It's a perfectly viable technique. I approve.

nikita-volkov:
> Currently one can't remove the imports because we build Dhall with very different GHC versions, some of these versions have those imports in the Prelude and others don't.
> 
> There are two possible ways out:
> 
> * reduce the number of GHC versions with which we build Dhall in CI
> * silence warnings in the GHC options that CI is using
> 
> What would you propose to do?

There's several issues at play here:

- [No linting policies or CI checks](https://github.com/dhall-lang/dhall-haskell/issues/2761)
- [Too wide dependency range support](https://github.com/dhall-lang/dhall-haskell/issues/2762)
- Need to suppress warnings per module due to conflicting imports between different versions of GHC. You've perfectly addressed that here.

PR #2758, opened 2026-06-29
Implement Text/equal built-in

nikita-volkov:
Yes. I know. The [design choices](https://docs.dhall-lang.org/discussions/Design-choices.html#text-manipulation). This PR is for when we reconsider them.


kukimik:
Note that this allows one (via `*/show`) to implement `Date/equal`, `Time/equal`, `TimeZone/equal` (which probably are not very controversial), but also `Double/equal`, which is in conflict with another [design choice](https://docs.dhall-lang.org/discussions/Design-choices.html#arithmetic).

PR #2766, opened 2026-07-07
Fix Nix build: pin microlens 0.4.14.0 to satisfy dhall.cabal bound

julmue:
The pinned nixpkgs (nix/pinnedNixpkgs.nix, Dec 2021) ships microlens 0.4.12.0, but dhall.cabal has required microlens >= 0.4.14.0 && < 0.6 since #2642, so the Nix build fails to configure dhall:

    Setup: Encountered missing or private dependencies:
      microlens >=0.4.14.0 && <0.6

Add nix/packages/microlens.nix (a cabal2nix derivation for microlens 0.4.14.0). nix/shared.nix already imports that directory via packagesFromDirectory, so this overrides the pin's 0.4.12.0 in the ghc8107 package set with no change to shared.nix. This mirrors what every stack.ghc-*.yaml already does (microlens-0.4.14.0 as an extra-dep) and keeps the current GHC 8.10.7.


nix/packages/microlens.nix
```diff
@@ -0,0 +1,10 @@
+{ mkDerivation, base, lib }:
```
winitzki, line 1: How will this file override the nix behavior? Should not this file be imported or referenced in some other nix file?

julmue: It gets picked up automatically by [fold](https://github.com/dhall-lang/dhall-haskell/blob/f2450ca627022cce71b3e57ed7d348b80fbc36d5/nix/shared.nix#L236).

winitzki (approved):
Good,thank you!

julian-mueller-na2 (approved):

PR #2774, opened 2026-07-10
Fix freeze not applying fields from settings

### LINKED ISSUE #2768: Freeze.freezeWithSettings does not copy over substitution and other EvaluateSettings
We use substitutions for type aliases and we were creating a custom freeze command, that passes those substitutions to `freezeWithSettings`. Even though `freezeWithSettings` takes `EvaluateSettings` as a parameter, it only creates a status with EvaluateSettings.newManager. It does not copy over the remaing fields, like e.g. `resolveAndStatusWithSettings` does: https://github.com/dhall-lang/dhall-haskell/blob/6deed7c74fd7dec75f0f3c9607f112e6937fb6a0/dhall/src/Dhall.hs#L277-L283

Is this on purpose?

If the fields should be copied, I can prepare a PR that either:
* Creates a transform similar to `resolveAndStatusWithSettings` and applies it
* Creates a helper `emptyStatusWithSettings` and uses that in relevant places

The helper could look like this:
```haskell
emptyStatusWithSettings :: EvaluateSettings -> FilePath -> Status
emptyStatusWithSettings EvaluateSettings{..} rootDirectory =
        transform (Dhall.Import.emptyStatusWithManager _newManager rootDirectory)
    where
        transform =
               Lens.set Dhall.Import.substitutions _substitutions
            .  Lens.set Dhall.Import.normalizer _normalizer
            .  Lens.set Dhall.Import.startingContext _startingContext
            .  Lens.set Dhall.Import.reportWarning _reportWarning
```

kindlnatuvion:
Fixes #2768 

Side note: `emptyStatusWithSettings` cannot be moved to Dhall.Import because it would introduce a cyclic dependency. I am open for suggestions if this should be moved into a different module.

### DIFF
--- a/dhall/src/Dhall.hs
+++ b/dhall/src/Dhall.hs
@@ -52,6 +52,7 @@ module Dhall
     , parseWithSettings
     , resolveWithSettings
     , resolveAndStatusWithSettings
+    , emptyStatusWithSettings
     , typecheckWithSettings
     , checkWithSettings
     , expectWithSettings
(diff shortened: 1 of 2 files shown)


winitzki (approved):
LGTM!

kindlnatuvion:
That was quick! Thank you

Issue #2780, opened 2026-07-14
Using `missing <hash> ? ./path.dhall` does not fill semantic cache


kindlnatuvion:
The `missing <hash> ? ./path.dhall` pattern is a common import pattern for example used in the Prelude.
I realized, that when the import is evaluated, the cache is checked for the hash, but if it is not found and going to the fallback path, the semantic cache will not be filled.
Is this an oversight or on purpose?

kindlnatuvion:
I now have a fix for this ready. I will do some testing and I can turn it into a PR.

winitzki:
@kindlnatuvion  I am thinking about this and I'm trying to understand the implications.

Suppose we have an import that looks like this:

```dhall
let x = missing sha256:abc4...  ? ./file2 sha256:def8...  ? ./file3
```
Suppose the `abc4...` is not in the cache. Then the `missing` will fail and we proceed to `./file2`. Suppose `file2` also fails to read. Then we proceed to `./file3`.  Now you propose to cache the contents of `file3`. But with what hash?

It is not excluded that the hashes are different for different files.

Second question: was the cache being filled when importing Prelude from http, but not being filled when importing Prelude from a local file? I thought if the cache wasn't being filled for `missing sha256:abc4... ? ./blah` imports then this trick doesn't seem to give any advantage. I actually thought that the import pattern `missing sha256:abc4... ? ./blah` will always fill the cache after first use when importing Prelude from http. But I haven't tested that myself in detail.

kindlnatuvion:
To the first question:
I think the simplest version would be to recurse into the left hand side when looking for the hash.
I made some changes and pushed them to the PR branch.
This should also preserve the (a ? b) ? c == a ? (b ? c) property, if I have thought about it correctly.

To the second question:
I am not sure what you mean exactly.
It made no difference if the file was imported from the web or locally. In both cases it did not end up in the cache.

Testing this file with the current version:
```dhall
let Prelude = https://prelude.dhall-lang.org/package.dhall
in Prelude.List.map Natural Text (\(x : Natural) -> Natural/show x) [1, 2, 3]
```
or frozen:
```
let Prelude =
        missing
          sha256:931cbfae9d746c4611b07633ab1e547637ab4ba138b16bf65ef1b9ad66a60b7f
      ? https://prelude.dhall-lang.org/package.dhall

in  Prelude.List.map Natural Text (\(x : Natural) -> Natural/show x) [ 1, 2, 3 ]
```
[...]

PR #2781, opened 2026-07-14
Fix `missing <hash> ? ./path.dhall` not filling semantic cache

kindlnatuvion:
Addresses #2780

I realized that when running a simple file like the following on a cold cache:
```
let Prelude = ../../dhall/dhall-lang/Prelude/package.dhall
in Prelude.List.map Natural Text (\(x : Natural) -> Natural/show x) [1, 2, 3]
```
that the semantic cache ( ~/.cache/dhall ) was empty.
This meant that re-running the file was barely being sped up.

The old behavior can be simulated by cleaning ~/.cache/dhall after warming the cache.

### DIFF
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -1246,7 +1246,25 @@ loadWith expr₀ = case expr₀ of
           | any isNotResolutionError es₀ =
               throwM exception₀
           | otherwise = do
-              loadWith b `catch` handler₁
+              result <- loadWith b `catch` handler₁
+
+              -- If the left side was a frozen missing import
+              -- and the right side succeeded
+              -- populate the semantic cache.
+              case Core.shallowDenote a of
+                Embed (Import (ImportHashed (Just h) _) _) -> do
+                    Status { _reportWarning } <- State.get
+
+                    let bytes = encodeExpression (Core.alphaNormalize (Core.denote result))
+
+                    let actualHash = Dhall.Crypto.sha256Hash bytes
+
+                    if actualHash == h
+                        then zoom cacheWarning (writeToSemanticCache _reportWarning h bytes)
+                        else return ()
+                _ -> return ()
+
+              return result
         where
           handler₁ exception₁@(SourcedException (Src _ end text₁) (MissingImports es₁))
               | any isNotResolutionError es₁ =
--- a/dhall/src/Dhall/Import.hs
+++ b/dhall/src/Dhall/Import.hs
@@ -1249,21 +1249,26 @@ loadWith expr₀ = case expr₀ of
               -- If the left side was a frozen missing import
               -- and the right side succeeded
               -- populate the semantic cache.
-              case Core.shallowDenote a of
-                Embed (Import (ImportHashed (Just h) Missing) _) -> do
+              case findMissingImportHash a of
+                Just hash -> do
                     Status { _reportWarning } <- State.get
 
                     let bytes = encodeExpression (Core.alphaNormalize (Core.denote result))
 
                     let actualHash = Dhall.Crypto.sha256Hash bytes
 
-                    if actualHash == h
-                        then zoom cacheWarning (writeToSemanticCache _reportWarning h bytes)
+                    if actualHash == hash
+                        then zoom cacheWarning (writeToSemanticCache _reportWarning hash bytes)
                         else return ()
-                _ -> return ()
+                Nothing -> return ()
 
               return result
         where
+          findMissingImportHash expr = case Core.shallowDenote expr of
+            Embed (Import (ImportHashed (Just hash) Missing) _) -> Just hash
+            ImportAlt left _ -> findMissingImportHash left
+            _ -> Nothing
+
           handler₁ exception₁@(SourcedException (Src _ end text₁) (MissingImports es₁))
               | any isNotResolutionError es₁ =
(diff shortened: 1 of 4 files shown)


dhall/src/Dhall/Import.hs
```diff
@@ -1246,7 +1244,25 @@ loadWith expr₀ = case expr₀ of
...
               throwM exception₀
           | otherwise = do
-              loadWith b `catch` handler₁
+              result <- loadWith b `catch` handler₁
+
+              -- If the left side was a frozen missing import
+              -- and the right side succeeded
+              -- populate the semantic cache.
+              case Core.shallowDenote a of
+                Embed (Import (ImportHashed (Just h) Missing) _) -> do
```
kindlnatuvion: I was not entirely sure if this should match on any hashed import or only on `missing`

dhall/src/Dhall/Import.hs
```diff
@@ -1213,8 +1213,6 @@ loadWith expr₀ = case expr₀ of
         then return ()
         else throwMissingImport (Imported _stack (ReferentiallyOpaque import₀))
 
-    let _stack' = NonEmpty.cons child _stack
```
kindlnatuvion, line 1219: There was a missing variable warning and I removed it

winitzki:
I have some questions here: https://github.com/dhall-lang/dhall-haskell/issues/2780#issuecomment-4981688047

dhall/src/Dhall/Import.hs
```diff
@@ -1246,8 +1244,31 @@ loadWith expr₀ = case expr₀ of
...
+
+                    if actualHash == hash
+                        then zoom cacheWarning (writeToSemanticCache _reportWarning hash bytes)
+                        else return ()
+                Nothing -> return ()
+
+              return result
         where
+          findMissingImportHash expr = case Core.shallowDenote expr of
+            Embed (Import (ImportHashed (Just hash) Missing) _) -> Just hash
```
winitzki: I believe it's better to match on any import, not necessarily on a `missing`.

winitzki:
@kindlnatuvion  I am having doubts about the associativity of `?` with this change.

Consider two examples:

(1) `(missing sha256:x ? p) ? q` and `missing sha256:x ? (p ? q)` where `p` resolves successfully.

(2) `(missing sha256:x ? p) ? q` and `missing sha256:x ? (p ? q)` where `p` fails but `q` resolves successfully.

I believe that, with this PR change, in example (1) the cache product for `p` will be written for `(missing sha256:x ? p) ? q` but will not be written for `missing sha256:x ? (p ? q)`. 

In example (2), I believe that the cache product for `q` will be written for `(missing sha256:x ? p) ? q` but will not be written for `missing sha256:x ? (p ? q)`. Is this correct?

I am not yet sure how to test this behavior. There are no standard tests that assert that cache products should be written to the cache. (There are tests asserting that cache should be read though.)

kindlnatuvion:
Hello Sergei,

Thanks for your interest and input on this issue.

I usually had tested this manually by running it and checking what was in cache.
First, I was thinking if we want to test this automatically it might make sense to have an option to change the cache folder.
I had a look into the tests and there is already a setup to simulate that by setting the env variable `XDG_CACHE_HOME` to a temporary folder.

> There are no standard tests that assert that cache products should be written to the cache.
I think I could just:
* Create a temporary cache
* Evaluate an expression with import
* Check if the cache file is there
* Clear tmp cache
* Continue with other expressions

I will look into a test that checks (a ? b ? c) with missing imports.

Then I can also better answer the question
* If this should only happen for missing-imports or all
* If we should recurse in finding the hash

dhall/src/Dhall/Import.hs
kindlnatuvion: I have updated it.

kindlnatuvion:
I now match on all imports with hash and go in both directions. I have added a test.

dhall/tests/Dhall/Test/CacheFill.hs
```diff
@@ -0,0 +1,66 @@
...
+
+    expr <- Dhall.inputExpr simpleValue
+
+    let expectedHash = Import.hashExpression (Core.denote expr)
+
+    let semanticCacheDir = cacheDir </> "dhall"
+
+    let cacheFile = semanticCacheDir </> "1220" <> show expectedHash
+
+    tempFile <- Temp.writeTempFile "." "tmp.dhall" (Text.unpack simpleValue)
```
winitzki, line 33: This fails tests under Windows because:

```
 Error: Invalid input
    
    (input):1:98:
      |
    1 | (missing sha256:27abdeddfe8503496adeb623466caa47da5f63abd2bc6fa19f6cfcb73ecfed70 ? missing) ? ./.\tmp140F.dhall
      |                                                                                                  ^
    unexpected '\'
```
https://github.com/dhall-lang/dhall-haskell/actions/runs/29508580990/job/87655773634?pr=2781

I believe your path is `././tmp140F.dhall` on Unix-like systems (which is valid) but `./.\tmp140F.dhall`  on Windows, which is invalid as Dhall syntax. 

Either you would need to create the Dhall path to `tempFile` separately from creating the `tempFile` itself, or you could convert backslash to slash.

winitzki (approved):
This sounds like  a good improvement. I will follow up with more tests and perhaps a clarification in the Dhall standard.

PR #2789, opened 2026-07-27
Fix: Add threaded option for cabal test

kindlnatuvion:
Running `cabal test` had the tests requiring a local server fail, because the test server required the threaded option. This PR adds it to the cabal file.


felixspringernatuvion (approved):
I don't see any reason why would want the test to be single threaded.

winitzki (approved):

PR #2791, opened 2026-08-02
Map-based substitution

kindlnatuvion:
We have a big prelude and evaluating it for the first time took around 10 minutes.
We profiled and found substitution was the bottleneck.
For us the substitutions work as some form of type aliases where we substitute vars with their corresponding types, so that we do not have to write out e.g. big records.
As we have many types in the map, this became quite expensive, because expressions had to be traversed for each entry.
Using a map-based approach the expression has to be traversed only once and evaluation time was reduced to around 2 minutes, so this was a noticeable speed up.
As a side note: we handle interdependecies by applying the map to itself.

The tests are a little bit ad-hoc and thrown together.
You can decide which tests are essential and I can throw some out if they are unnecessary.
The substitute code should be fairly readable, but I am also open for suggestions.

### DIFF
--- a/dhall/src/Dhall/Substitution.hs
+++ b/dhall/src/Dhall/Substitution.hs
@@ -4,12 +4,14 @@
 
 module Dhall.Substitution where
 
-import Data.List       (foldl')
 import Data.Text       (Text)
-import Dhall.Normalize (subst)
-import Dhall.Syntax    (Expr, Var (..))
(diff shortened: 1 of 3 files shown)


winitzki (approved):
I ran eval benchmarks, performance seems to be unchanged for normal cases.

TristanCacqueray:
That's great, thanks!

PR #2793, opened 2026-08-06
Remove unconditional short-circuit from Natural/fold evaluation

winitzki:
Previously `Natural/fold` was optimized so that the loop stops early when the current value no longer changes. 

https://github.com/dhall-lang/dhall-haskell/pull/2585
https://github.com/dhall-lang/dhall-haskell/pull/2596

However,  if `Natural/fold` is used to build up a big list of values or a big lambda function, then we aren't going to benefit from this optimization and in fact it makes things slower, as a big structure is going to be compared with another big structure all the time and the shortcut is pretty much never going to be realized. 

This PR changes the logic so that the shortcut is only used for simple types (`Natural`, `Integer`, `Text`, and records / unions / `Optional` containing those), but not for `List` or for function types.

To be tested (probably we want some of the short-circuit eventually). Results so far:

Baseline (main branch):

```
  normalize
    ChurchEval
      typecheck:  OK
        22.9 μs ± 1.7 μs
      evaluation: OK
        3.070 s ±  68 ms
    FunCompose
      typecheck:  OK
        9.75 μs ± 871 ns
      evaluation: OK
        2.133 s ±  38 ms
    ListBench
      typecheck:  OK
        2.552 s ±  43 ms
      evaluation: OK
        1.281 s ±  31 ms
    ListBenchAlt
      typecheck:  OK
        3.752 s ± 153 ms
      evaluation: OK
        1.866 s ±  35 ms
```

This branch:

```
  normalize
    ChurchEval
      typecheck:  OK
        23.0 μs ± 2.2 μs
      evaluation: OK
        3.115 s ±  61 ms
    FunCompose
      typecheck:  OK
        10.0 μs ± 419 ns
      evaluation: OK
        513  μs ±  43 μs
    ListBench
      typecheck:  OK
        1.916 s ±  50 ms
      evaluation: OK
        948  ms ±  33 ms
    ListBenchAlt
      typecheck:  OK
        2.52 ms ± 107 μs
      evaluation: OK
        2.40 ms ± 214 μs
   NaturalFoldShortcut
      typecheck:  OK
        480  ns ±  29 ns
      evaluation: OK
        8.03 μs ± 436 ns

```

### DIFF
--- a/.github/workflows/renovatebot.yml
+++ b/.github/workflows/renovatebot.yml
@@ -1,6 +1,8 @@
 # See: https://github.com/renovatebot/github-action?tab=readme-ov-file#example
 # See: https://github.com/jenkinsci/helm-charts/blob/efc6225850a9c4e5a7f83b15223c21ff06c5b093/.github/workflows/renovate.yml
 
+# For the latest version, see: https://github.com/renovatebot/github-action/releases
+
 name: renovatebot
 
 on:
@@ -20,7 +22,7 @@ jobs:
         uses: actions/checkout@v7
 
       - name: Self-hosted Renovate
-        uses: renovatebot/github-action@v46.1.21
+        uses: renovatebot/github-action@v46.2.1
         with:
           configurationFile: .github/renovate.json5
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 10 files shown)


winitzki:
Github has provisioning issues. Will need to retry tomorrow.

PR #2822, opened 2026-08-31
Fix compatibility with mmark 0.1

felixonmars:
PR #2802 allow mmark 0.1 but probably because of cached build deps, the CI doesn't seem to catch the actual build failure:

```
 [ 2 of 10] Compiling Dhall.Docs.Markdown ( src/Dhall/Docs/Markdown.hs, dist/build/Dhall/Docs/Markdown.dyn_o )

 src/Dhall/Docs/Markdown.hs:33:5: error: [GHC-83865]
     • Couldn't match type: MMark -> Html ()
                      with: HtmlT Data.Functor.Identity.Identity ()
       Expected: MMark.RenderExtension -> Html ()
         Actual: MMark.RenderExtension -> MMark -> Html ()
     • Probable cause: ‘MMark.render’ is applied to too few arguments
       In the first argument of ‘(<$>)’, namely ‘MMark.render’
       In the expression: MMark.render <$> parseMarkdown relFile contents
       In an equation for ‘markdownToHtml’:
           markdownToHtml relFile contents
             = MMark.render <$> parseMarkdown relFile contents
    |
 33 |     MMark.render <$> parseMarkdown relFile contents
    |     ^^^^^^^^^^^^

 src/Dhall/Docs/Markdown.hs:33:22: error: [GHC-83865]
     • Couldn't match type ‘MMark’ with ‘MMark.RenderExtension’
       Expected: Either MarkdownParseError MMark.RenderExtension
         Actual: Either MarkdownParseError MMark
     • In the second argument of ‘(<$>)’, namely
         ‘parseMarkdown relFile contents’
       In the expression: MMark.render <$> parseMarkdown relFile contents
       In an equation for ‘markdownToHtml’:
           markdownToHtml relFile contents
             = MMark.render <$> parseMarkdown relFile contents
    |
 33 |     MMark.render <$> parseMarkdown relFile contents
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Let's fix it.


winitzki:
@felixonmars Does this change now break the GHC 8.10 build? Can that be fixed?

As far as I know the only reason for keeping GHC 8.10 in the build is the old ghcjs build used for `dhall-try` (the javascript-based backend used in the web site to allow users to evaluate Dhall expressions in the browser). There is a PR intending to allow `dhall-try` to build with GHC 9.6 https://github.com/dhall-lang/dhall-haskell/pull/2810 but I haven't verified that it works (just that it builds). If it works and if we can move the Dhall website to the newly built `dhall-try` then we could drop GHC 8.10.

felixonmars:
I can also fix this using CPP to preserve compatibility with old `mmark-0.0.x`. Will update later.

PR #2827, opened 2026-09-16
Fix GHC 9.6 CPP guard mismatch for Network.TLS.Supported

nikita-volkov:
The checks within the file were targeting GHC 9.6 while the necessary imports were targeting 9.8.

https://github.com/dhall-lang/dhall-haskell/blob/c85811f76fa2cb4833b1ee91ea27c58b1000a929/dhall/tests/Dhall/Test/Import.hs#L231


winitzki (approved):
If all builds pass then it's good.

PR #2829, opened 2026-09-16
Add advisory cabal CI matrix alongside Stack

nikita-volkov:
Step 2 of the CI/CD migration (docs/ci-cd-migration-plan.md, not itself
tracked in the repo — internal working doc, happy to share on request).

Adds `.github/workflows/cabal.yml`: a build+test matrix across GHC
8.10 through `latest`, on 5 platforms (linux-x64, linux-arm64,
macos-arm64, macos-x64, windows-x64), running alongside the existing
Stack-based `main.yml`. It is **not a merge gate** — no job here is
named in `.mergify.yml`, so it cannot block anything.

`cabal.project` already encodes the GHC-conditional package set
(`dhall-nix`/`dhall-nixpkgs` excluded below GHC 9.8), so unlike
`main.yml`'s three copy-pasted bash `if` blocks, this workflow needs no
per-package skip logic — `cabal build all` / `cabal test all` just work
across the matrix.

**Known-red leg, filed openly rather than hidden:** the GHC 8.10 job is
expected to fail on `dhall-docs`. Verified locally (with and without
the untracked `cabal.project.local` overrides): `dhall-docs` depends on
`mmark >= 0.0.7.0 && < 0.2`, every release in that range requires
`html-entity-map >= 0.1 && < 0.2`, and the only Hackage version
satisfying that bound, `html-entity-map-0.1.0.0`, requires
`base >= 4.15` — GHC 8.10 ships `base-4.14.3.0`. `html-entity-map` has
exactly two releases ever, so no cabal constraint threads this needle;
it's a real dead end on current Hackage, not a transient index issue.
Stack's `lts-18.28` sidesteps it via a 2021 index snapshot predating
the bump, which cabal's plain Hackage solve has no equivalent of.
GHC 9.8.4 and 9.12.4 both solve and were verified locally; 9.2/9.4/9.6
are untested but should be fine since the failure threshold is
`base >= 4.15`, which all three exceed.

Left as a maintainer call for a follow-up: either exclude `dhall-docs`
from the GHC 8.10 leg the same way `dhall-nix`/`dhall-nixpkgs` are
excluded above GHC 9.8, or drop GHC 8.10 from the cabal matrix
outright. Not resolved in this PR since it's a real scope/support
decision, not a mechanical fix.

## Test plan

- [x] `cabal build all --dry-run --enable-tests` verified locally to
      resolve on GHC 9.8.4 and 9.12.4
- [x] Same command verified to fail identically on GHC 8.10.7, with and
      without `cabal.project.local`'s `allow-newer` overrides in place
- [x] YAML validated (`ruby -ryaml`)
- [ ] Actual GitHub Actions run on this PR (advisory, so red legs here
      are informative, not blocking)


nikita-volkov:
Accident.

PR #2833, opened 2026-09-18
As-Source optimizations

winitzki:
Selected optimizations on top of the bare-bones implementation of `as Source` imports.


winitzki:
@kindlnatuvion  I have added some more optimizations; it would be interesting if you try your workload with this PR's code when you have time.

I merged some changes into `main` that effectively do `as source` imports by default, avoiding to expand each import into the full normal form, unless the full NF is really necessary (it is only necessary for hash-protected imports). Some other fixes to do with laziness made the `main` branch faster on big imports. So, it's also interesting if you just try the current `main` branch code, without `as Source`.

I am getting an impression that the current `main` branch is actually faster than `as Source` at this point, and the feature of imports `as Source` is no longer necessary or helpful.

